PR #1081, opened 2023-01-21
add new `from` and `fromJavaObject` methods

MagicFollower:
… to JSONObject, corresponding to the `to` and `toJavaObject` apis.

After analyzing the conversion relationship between JSONObject types, we find that there is a JSONObject toJavaObject API (to/toJavaObject), but there is no reverse. Currently, PR is designed to submit the reverse correlation APIs to implement the closed-loop link between JSONObject and JavaObject.

### Summary of your change
Add new from and fromJavaObject methods for converting JavaObject to JSONObject, corresponding to the To and ToJavaObject apis.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
@@ -1473,7 +1473,6 @@ public class JSONArray
         return JSON.parseArray(input, type);
     }
 
-
     /**
      * See {@link JSON#toJSON} for details
      */
--- a/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
@@ -1486,18 +1486,4 @@ public class JSONArray
     public static JSONArray from(Object obj, JSONWriter.Feature... writeFeatures) {
         return (JSONArray) JSON.toJSON(obj, writeFeatures);
     }
-
-    /**
-     * See {@link JSON#toJSON} for details
-     */
-    public static JSONArray fromJavaObject(Object obj) {
-        return from(obj);
-    }
-
-    /**
-     * See {@link JSON#toJSON} for details
-     */
-    public static JSONArray fromJavaObject(Object obj, JSONWriter.Feature... writeFeatures) {
-        return from(obj, writeFeatures);
-    }
 }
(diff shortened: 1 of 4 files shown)


MagicFollower:
Add new from and fromJavaObject methods for converting JavaObject to JSONObject, corresponding to the To and ToJavaObject apis.

wenshao:
all method is private?

MagicFollower:
> all method is private?

It was an oversight on my part, I'll fix it and resubmit.

MagicFollower:
> there is already a method : JSON#toJSON
> 
> ```java
> interface JSON {
>       static Object toJSON(Object object, JSONWriter.Feature... features) {}
> }
> ```

In fact, My first motivation is to provide a more intuitive API for JSONObject, similar to the to() series. Because when someone see to(), immediately think api is from().
And since JSON has an existing API, it is a better solution to build a new api using an existing one.
1. If you also think this is necessary, I'll find time to provide the from() family API for both JSONObject and JSONArray.
2. Else we can directly use JSON#toJSON, but this API seems so far from JSONObject and JSONArray （🚩）

wenshao:
from and fromJavaObject are duplicate,  can you remove from?

MagicFollower:
> from and fromJavaObject are duplicate, can you remove from?
Rather than removing one, I recommend keeping both，beacuse API from() corresponds to to(), fromJavaObject() corresponds to toJavaObject(), like that:

`        User user = new User(UUID.randomUUID().toString().replaceAll("-", ""), "AAA");

        /* New 👇 */
        JSONObject from = JSONObject.from(user);
        JSONObject from1 = JSONObject.fromJavaObject(user);


        /* Exists 👇 */
        User to = from.to(User.class);
        User to1 = from.toJavaObject(User.class);`

wenshao:
toJavaObject & fromJavaObject are more unambiguous, add the 'to' method, I have always suspected that the decision I made was correct.

MagicFollower:
> toJavaObject & fromJavaObject are more unambiguous, add the 'to' method, I have always suspected that the decision I made was correct.


From my point of view, the `core` task of this library is the conversion of Java-Object and JSON-Object. You maybe worry about the ambiguity of this `to()` in the future and wanna give it more, don't suspect, it is correct to keep it and keep your original decision, that's it, short and powerful😊

kraity:
the `toJavaObject` method is marked `@Deprecated` and it will be deprecated in future releases. so, shouldn't add `fromJavaObject` method

PR #1330, opened 2023-04-07
Some refactoring to improve design and implementation

msinojia:
I went through the code structure and applied some refactoring techniques in an attempt to improve some of the potential design or implementation issues.

### Summary of your change
In specific, I have applied the following refactoring techniques (one in each commit):
1) Decompose conditional
2) Extract method
3) Introduce explaining variable
4) Replace conditional with polymorphism
5) Extract class
6) Move method

I would be glad to hear what you think about these refactorings, and if they make sense in this project or if I need to improve anything to make it worth it.

- [✓ ] Made sure tests are passing and test coverage is added if needed.
- [✓ ] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [ ✓] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
@@ -358,11 +358,7 @@ class JSONReaderJSONB
                     value = new Integer(int32Value);
                 } else if (valueType == BC_INT32) {
                     offset++;
-                    int int32Value =
-                            ((bytes[offset + 3] & 0xFF)) +
-                                    ((bytes[offset + 2] & 0xFF) << 8) +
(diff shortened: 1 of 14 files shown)


wenshao:
Thank you very much for submitting the PR, I hope you will continue to participate in the future.

But this PR has some problems and cannot be merged, because : 
In order to improve performance and make the call stack less, fastjson2 has done a lot of manual inlining

Some of the issues here are doing this on purpose, the smaller the call stack, the better the performance in many cases.

msinojia:
Thank you for the feedback. I now understand why certain things are done the way they are. It makes sense for this library to have performance as the top priority.

PR #1464, opened 2023-05-11
fix getString method doesn't work as expected for issue #1446

sadblack:

fix getString method doesn't work as expected for issue #1446

方法判断条件里增加了 TemporalAccessor，经测试 对于 java.time下的类，getString 方法与 fastjson的 getString 返回结果一致


wenshao:
JSONArray的getString方法也一并改了吧

sadblack:
您好，改好啦

PR #1522, opened 2023-06-01
update JSONPathParser with a nullable judgement added, for issue #1516

yanxutao89:
Sorry for the second pr on the same issue. The more defensive, the more robust. Even if the GourpFilter.Filters can not be null in current version, but we can not guarantee that stands in future version. So, I add a nullable judgement before iterating over the filters.

### Summary of your change

image

add a nullable judgement before iterating over the filters.


wenshao:
run 'mvn validate' and fix checkstyle error

PR #1839, opened 2023-09-07
feat enhance the reading of number when illegal inputted

yanxutao89:

feat enhance the reading of number when illegal inputted

### Summary of your change

add a valid flag to enable fail-fast when illegal number inputted

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
@@ -4049,8 +4049,6 @@ class JSONReaderUTF8
                 }
                 return null;
             }
-        } else if (ch == ',' || ch == '\r' || ch == '\n') {
-            return null;
         }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
```diff
@@ -4052,6 +4062,7 @@ public final Integer readInt32() {
 
         boolean overflow = false;
         while (ch >= '0' && ch <= '9') {
+            valid = true;
```
wenshao, line 4063: 做这个改动之后，要在删除这个类中的4052~4053行代码
```
        } else if (ch == ',' || ch == '\r' || ch == '\n') {
            return null;
```

wenshao:
JSONReaderUTF8#readNumber0这个方法中还有没改的
```java
} else if (ch == ',' || ch == '\r' || ch == '\n') {
```

Issue #2029, opened 2023-11-19
[BUG] Detected 4 Flaky Tests in `fastjson1-compatible`


krishnanand5:
### 问题描述
As a follow up to https://github.com/alibaba/fastjson2/issues/2017, I am creating this issue to document the findings seen in 4 other tests in `fastjson1-compatible` module. 

Just like the previous issue, I have used the [NonDex](https://github.com/TestingResearchIllinois/NonDex) to detect the flakiness seen in the tests mentioned in this issue.

The four tests identified as part of this exercise are:
- `com.alibaba.fastjson.issue_2100.Issue2182#test_for_issue`.
- `com.alibaba.fastjson.issue_2400.Issue2447#test_for_issue`.
- `com.alibaba.fastjson.issue_1100.Issue1177_1#test_for_issue`.
- `com.alibaba.fastjson.issue_2400.Issue2428#test_for_issue`.

### 环境信息
*请填写以下信息：*

 - OS信息：  [MacOS Sonoma 8GB]
 - JDK信息： [JAVA 11.0.21-amzn]
 - 版本信息：[2.0.42]
 

### 重现步骤
*如何操作可以重现该问题：*

#### `com.alibaba.fastjson.issue_2100.Issue2182#test_for_issue`

- Run ` mvn -pl fastjson1-compatible edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=com.alibaba.fastjson.issue_2100.Issue2182#test_for_issue -DnondexMode=ONE`.
- Due to shuffling, the order of elements in the jsonObject will be shuffled and when this JSON is compared with a static JSON, the assertion can fail as the exact order of elements will not match.

```
[INFO] Running com.alibaba.fastjson.issue_2100.Issue2182
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.136 s <<< FAILURE! -- in com.alibaba.fastjson.issue_2100.Issue2182
[ERROR] com.alibaba.fastjson.issue_2100.Issue2182.test_for_issue -- Time elapsed: 0.126 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <{"admin":["admin.create","admin.update","admin.delete"],"user":["user.create","user.delete"]}> but was: <{"user":["user.create","user.delete"],"admin":["admin.create","admin.update","admin.delete"]}>
[...]

PR #2030, opened 2023-11-19
Addressing 4 flaky tests in `fastjson1-compatible`

krishnanand5:

This PR fixes the issues captured in https://github.com/alibaba/fastjson2/issues/2029

### Summary of your change

- As elaborated in the above mentioned issue, The assertions performed in the 4 tests are asserting on JSON data but the order of elements between the anticipated JSON and generated JSON varies. 
- The 4 identified tests are :

  - `com.alibaba.fastjson.issue_2100.Issue2182#test_for_issue`.
  - `com.alibaba.fastjson.issue_2400.Issue2447#test_for_issue`.
  - `com.alibaba.fastjson.issue_1100.Issue1177_1#test_for_issue`.
  - `com.alibaba.fastjson.issue_2400.Issue2428#test_for_issue`.
  
- This leads to the assertion failing non-deterministically as the order of elements does not remain consistent.
- In order to address this issue and carry out the validation on both the presence and order of elements in the JSON object, we can use the `JSONAssert` instead of plain string assertion.
- The biggest advantage of `JSONAssert` is the ability to configure the assertion to check not only the components but also their order.
- Usage of `JSONAssert` will require the `org.skyscreamer.jsonassert.JSONAssert` dependancy and the necessary changes have been made in pom.xml

### DIFF
--- a/fastjson1-compatible/pom.xml
+++ b/fastjson1-compatible/pom.xml
@@ -339,6 +339,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.1</version>
+            <scope>test</scope>
+        </dependency>
+
         <!-- spring libs -->
         <dependency>
             <groupId>org.springframework</groupId>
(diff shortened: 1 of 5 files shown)


krishnanand5:
Thank you @wenshao !

Issue #2033, opened 2023-11-20
[BUG] Identified 6 Flaky tests in `fastjson1-compatible`


krishnanand5:
### 问题描述
As a follow up to previous issues, https://github.com/alibaba/fastjson2/issues/2029 and https://github.com/alibaba/fastjson2/issues/2017, I have identified 6 fresh flaky tests in `fastjson1-compatible` module.

They are the following:
- `com.alibaba.fastjson.issue_1900.Issue1972#test_for_issue`.
- `com.alibaba.fastjson.issue_1300.Issue1363#test_for_issue`.
- `com.alibaba.fastjson.issue_1300.Issue1363#test_for_issue_1`.
- `com.alibaba.fastjson.geo.FeatureCollectionTest#test_geo`.
- `com.alibaba.fastjson.jsonp.JSONPParseTest1#test_f`.
- `com.alibaba.fastjson.jsonp.JSONPParseTest#test_f`.

### 环境信息
*请填写以下信息：*

 - OS信息：  [MacOS Sonoma 8GB]
 - JDK信息： [JAVA 11.0.21-amzn]
 - 版本信息：[2.0.42]
 

### 重现步骤
*如何操作可以重现该问题：*

#### `com.alibaba.fastjson.issue_1900.Issue1972#test_for_issue`
- Run `mvn -pl fastjson1-compatible edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=com.alibaba.fastjson.issue_1900.Issue1972#test_for_issue -DnondexMode=ONE`
- The assertion of **text** with a static JSON string causes issues as the order of elements in **text** is not deterministic.
- Hence, the assertion can fail if the order of elements do not match between **text** and the static expected entity.
```
[INFO] Running com.alibaba.fastjson.issue_1900.Issue1972
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.202 s <<< FAILURE! -- in com.alibaba.fastjson.issue_1900.Issue1972
[ERROR] com.alibaba.fastjson.issue_1900.Issue1972.test_for_issue -- Time elapsed: 0.188 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <{"a":{"b":{"c":"2018-04","d":[123]}}}> but was: <{"a":{"b":{"d":[123],"c":"2018-04"}}}>
[...]

PR #2034, opened 2023-11-20
Fixed 6 Flaky tests in `fastjson1-compatible` module

krishnanand5:

This PR fixes all flaky tests captured in https://github.com/alibaba/fastjson2/issues/2033.

### Summary of your change

- As elaborated in the above mentioned issue, 4 of the identified 6 tests are asserting on JSON objects and this could lead to failures if the ordering of elements in the JSON object is shuffled. 
- The other two tests `com.alibaba.fastjson.jsonp.JSONPParseTest1#test_f` and `com.alibaba.fastjson.jsonp.JSONPParseTest#test_f` can fail non-deterministically if the strings are compared directly without considering their order.
- For the four tests with JSON assertion issues, the solution is to use `JSONAssert` to perform the comparison as this compares not only the presence of elements but also their order.
- For `com.alibaba.fastjson.jsonp.JSONPParseTest1#test_f` and `com.alibaba.fastjson.jsonp.JSONPParseTest#test_f`, we can sort the string first using `SerializerFeature.MapSortField`.
- This ensures the strings are always in one particular order and then comparing with the expected string ensures the validation of both order of elements and their presence.


krishnanand5:
Thank you @wenshao !

Issue #2055, opened 2023-11-28
[BUG] Identified some flaky tests in `core` and `extension` modules


krishnanand5:
### 问题描述
With reference to issues https://github.com/alibaba/fastjson2/issues/2033, https://github.com/alibaba/fastjson2/issues/2029 and https://github.com/alibaba/fastjson2/issues/2017, I can confirm that all flaky tests that were detected in `fastjson1-compatible` module have been addressed. 

I have attempted to find tests showing flaky behaviour in other modules of fastjson2 over the last week.

I have identified the following tests to be flaky in the `core` module :

1. `com.alibaba.fastjson2.issues.Issue586#test`
2. `com.alibaba.fastjson2.issues.Issue1411#test5`
3. `com.alibaba.fastjson2.issues_1000.Issue1494#test`
4. `com.alibaba.fastjson2.issues_1000.Issue1401#test`
5. `com.alibaba.fastjson2.v1issues.issue_2400.Issue2430#testForIssue`
6. `com.alibaba.fastjson2.v1issues.issue_2400.Issue2430#testForIssue2`
7. `com.alibaba.fastjson2.issues_1700.Issue1745#test`
8. `com.alibaba.fastjson2.v1issues.issue_1100.Issue1177_2#test_for_issue `
9. `com.alibaba.fastjson2.writer.ObjectWriterSetTest#testJsonbSet`
10. `com.alibaba.fastjson2.issues_1000.Issue1269#test`
11. `com.alibaba.fastjson2.issues_1000.Issue1479#test`
12. `com.alibaba.fastjson2.v1issues.issue_4000.Issue4008#test_for_issue4008`
13. `com.alibaba.fastjson2.v1issues.issue_3800.Issue3824#test_for_issue3824`
14. `com.alibaba.fastjson2.features.MapSortFieldTest#test`
15. `com.alibaba.fastjson2.jackson_support.JsonIncludeTest#test1`
16. `com.alibaba.fastjson2.issues.Issue273#test`
17. `com.alibaba.fastjson2.issues.Issue507#test1`
18. `com.alibaba.fastjson2.v1issues.geo.FeatureCollectionTest#test_geo`
19. `com.alibaba.fastjson2.v1issues.issue_2100.Issue2182#test_for_issue` 

I have identified the following test to be flaky in the `extension` module :

1. `com.alibaba.fastjson2.geo.FeatureCollectionTest#test_geo`

### 环境信息
*请填写以下信息：*

[...]

PR #2056, opened 2023-11-28
Fixed identified Flaky tests in `Core` and `Extension` Modules

krishnanand5:

This PR aims to fix all flaky tests captured in https://github.com/alibaba/fastjson2/issues/2055.

### Summary of your change

- As elaborated in the mentioned issue, 19 tests in `core` and 1 test in `extension` have been detected as flaky using the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool.
- Out of the 20 detected flaky tests, 18 off them are showing flaky behaviour owing to the comparison of a static JSON with a JSON object. This comparison expects the JSON object to hold the elements in the exact order like the static JSON.
- In order to carry out the comparison by taking both order of elements and actual presence of elements into account, we can use the `jsonAssert` function.
- I have added the required dependency to both modules for leveraging jsonAssert in the affected tests.
- For `com.alibaba.fastjson2.issues_1000.Issue1494#test`, we can switch `map` to leverage a `LinkedHashMap` which helps to preserve the order of elements in it. The previous `HashMap` does not preserve the order of elements.
- On a similar note, for `com.alibaba.fastjson2.writer.ObjectWriterSetTest#testJsonbSet`, we can switch `set` from `HashSet` to `LinkedHashSet` as the former is not expected to retain the exact order of elements.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -253,6 +253,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-aop</artifactId>
             <version>${springframework5.version}</version>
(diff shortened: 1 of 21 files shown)


krishnanand5:
Thank you @wenshao !

PR #2123, opened 2023-12-25
fix TypeReference NullPointerException

yohugSu:
While I am in the process of using fastjson2 to maintain compatibility with fastjson1, I have encountered a NullPointerException. I believe it should have null-checking logic, similar to other overloaded methods, as follows:

case1:
`JSONObject j = new JSONObject();`
`String key = null;`
`j.getObject(key, new TypeReference<String>(){});`    -> NullPointerException
// The key here could be empty, it should perform null checks like other overloaded methods.

case2:
`TypeReference<String> typeReference = null;`
`JSON.parseObject("key1", typeReference);`    -> NullPointerException
// The typeReference here could be empty, it should perform null checks like the toJavaObject method.


wenshao:
Please sign CLA

PR #2267, opened 2024-02-25
[FEATURE] JSON 比对 #1946

yanchangyou:

[FEATURE] JSON 比对 #1946

### Summary of your change

add compare util


core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+            boolean json1Contain = jsonPath1.containsKey(path);
+            boolean json2Contain = jsonPath2.containsKey(path);
+
+            JSONObject pathResult = new JSONObject();
+            pathResult.put("path", path);
+
+            for (String ignorePathPrefix : ignorePathPrefixList) {
+                if (path.startsWith(ignorePathPrefix)) {
+                    continue outer;
+                }
```
wenshao, line 164: **[Critical]** `startsWith` prefix matching lacks a path-delimiter check — unrelated fields sharing a string prefix with an ignored container path are silently skipped.

Failure scenario: `json1={"obj":{"a":1},"objX":2}`, `json2={"objX":3}`. Path `"obj"` is REMOVE with a JSONObject value → added to `ignorePathPrefixList`. Then `"objX".startsWith("obj")` is `true` → the MODIFY diff for `"objX"` (2→3) is silently dropped. Any pair of field names where one is a string prefix of the other (`"data"`/`"data2"`, `"item"`/`"items"`) triggers this.

```suggestion
            for (String ignorePathPrefix : ignorePathPrefixList) {
                if (path.startsWith(ignorePathPrefix + ".") || path.startsWith(ignorePathPrefix + "[")) {
                    continue outer;
                }
```

`startsWith` 前缀匹配缺少路径分隔符检查——与已忽略容器路径共享字符串前缀的无关字段会被静默跳过。

失败场景：`json1={"obj":{"a":1},"objX":2}`，`json2={"objX":3}`。路径 `"obj"` 为 REMOVE 且值为 JSONObject → 加入 `ignorePathPrefixList`。随后 `"objX".startsWith("obj")` 为 `true` → `"objX"` 的 MODIFY 差异（2→3）被静默丢弃。任何字段名存在前缀关系的情况（如 `"data"`/`"data2"`）都会触发此问题。

</details>

core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * 处理特殊key，key中有点号，然后冲突
+     *
+     * @param key field key
+     * @return norm key
+     */
+    static String buildJsonPathKey(String key) {
+        if (key.contains("[")) {
+            key = key.replace("[", "").replace("]", "");
+        }
```
wenshao, line 312: **[Critical]** Stripping `[` and `]` from field names creates path collisions between distinct keys and makes `getByPath` look up the wrong field.

Failure scenario: `json1={"a[0]":1,"a0":2}`. `buildJsonPathKey("a[0]")` strips brackets → `"a0"`, identical to `buildJsonPathKey("a0")`. The second `put` overwrites the first in the flattened map, so field `"a[0]"` vanishes from the comparison. Additionally, `getByPath("a0")` resolves to the literal field `"a0"`, not `"a[0]"`, so even without a collision the value lookup targets the wrong field.

```suggestion
        if (key.contains("[") || key.contains("]")) {
            return "['".concat(key.replace("'", "\\'")).concat("']");
        }
```

从字段名中去除 `[` 和 `]` 会导致不同键的路径冲突，并使 `getByPath` 查找错误的字段。

失败场景：`json1={"a[0]":1,"a0":2}`。`buildJsonPathKey("a[0]")` 去除括号后变为 `"a0"`，与 `buildJsonPathKey("a0")` 相同。第二次 `put` 覆盖了第一次，导致字段 `"a[0]"` 从比较中消失。此外 `getByPath("a0")` 解析为字面字段 `"a0"` 而非 `"a[0]"`，即使没有冲突也会查找错误的值。

</details>

core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+        if (key.contains(".") || key.contains("-")) {
+            return "['".concat(key).concat("']");
+        }
```
wenshao, line 315: **[Critical]** Single quotes inside keys are not escaped when wrapping in `['...']` notation, producing malformed JSONPath expressions.

Failure scenario: key `"a'.b"` → `buildJsonPathKey` returns `['a'.b']` — the embedded `'` terminates the bracket-quoted segment early. `getByPath` either throws an exception or resolves the wrong element. Any JSON key containing a single quote combined with a dot or dash triggers this.

```suggestion
        if (key.contains(".") || key.contains("-")) {
            return "['".concat(key.replace("'", "\\'")).concat("']");
        }
```

将键包裹为 `['...']` 格式时未转义内部的单引号，导致生成格式错误的 JSONPath 表达式。

失败场景：键 `"a'.b"` → `buildJsonPathKey` 返回 `['a'.b']`——内嵌的 `'` 提前终止了括号引用段。`getByPath` 会抛出异常或解析到错误的元素。任何包含单引号加点号或短横线的 JSON 键都会触发此问题。

</details>

core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+                            String newKey = entry.getKey();
+                            if (!newKey.startsWith("[")) {
+                                newKey = ".".concat(newKey);
+                            }
+                            String newFieldName = isArray ? arrayFieldName.concat(newKey) : fieldName.concat(newKey);
+                            newMap.put((newFieldName), entry.getValue());
+                        }
+                    } else {
+                        newMap.put((arrayFieldName), item);
+                    }
```
wenshao, line 251: **[Critical]** Nested arrays (arrays containing arrays) are never flattened recursively, so `compare()` silently skips them and reports equal for objects that differ.

Failure scenario: `json1={"a":[[1,2],[3,4]]}`, `json2={"a":[[1,2],[3,5]]}`. `convertWithJsonPath` stores inner arrays as opaque values at `"a[0]"`, `"a[1]"` (the `else` branch — JSONArray is not JSONObject, so no recursion). In `compare()`, all paths (`"a"`, `"a[0]"`, `"a[1]"`) hit the `both instanceof JSONArray → continue` guard. Result is empty: `equals()` returns `true` despite `[3,4] ≠ [3,5]`.

Suggested fix: recurse into JSONArray items similarly to the JSONObject case, or fall back to `Objects.equals(value1, value2)` when both values are JSONArray and no flattened children exist.

嵌套数组（数组中包含数组）不会被递归展平，因此 `compare()` 会静默跳过它们，对实际不同的对象报告相等。

失败场景：`json1={"a":[[1,2],[3,4]]}`，`json2={"a":[[1,2],[3,5]]}`。`convertWithJsonPath` 将内部数组作为不透明值存储在 `"a[0]"`、`"a[1]"`（`else` 分支——JSONArray 不是 JSONObject，不会递归）。在 `compare()` 中，所有路径都命中 `both instanceof JSONArray → continue` 守卫。结果为空：尽管 `[3,4] ≠ [3,5]`，`equals()` 仍返回 `true`。

</details>

core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+        JSONObject jsonPath2 = convertWithJsonPath(json2);
+
+        JSONObject result = new JSONObject();
+
+        List<String> jsonPathList = new ArrayList<>(jsonPath1.keySet());
+        for (String path : jsonPath2.keySet()) {
+            if (!jsonPathList.contains(path)) {
+                jsonPathList.add(path);
+            }
+        }
```
wenshao, line 150: **[Suggestion]** O(n×m) path union via `ArrayList.contains()` — two JSON objects each with 10,000 leaf paths would perform ~50M string comparisons. A `LinkedHashSet` preserves insertion order with O(1) containment.

Concrete cost: quadratic scaling makes large-object comparison (config snapshots, wide API responses) take seconds instead of milliseconds.

```suggestion
        Set<String> jsonPathSet = new LinkedHashSet<>(jsonPath1.keySet());
        jsonPathSet.addAll(jsonPath2.keySet());
        List<String> jsonPathList = new ArrayList<>(jsonPathSet);
```

通过 `ArrayList.contains()` 进行 O(n×m) 的路径合并——两个各含 10,000 个叶子路径的 JSON 对象将执行约 5000 万次字符串比较。`LinkedHashSet` 保持插入顺序且 `contains` 为 O(1)。

具体开销：二次方缩放使大对象比较（配置快照、宽 API 响应）从毫秒级变为秒级。

</details>

core/src/main/java/com/alibaba/fastjson2/util/CompareUtils.java
```diff
@@ -0,0 +1,318 @@
...
+                pathResult.put(FIELD_NAME_OF_VALUE_EQUAL, false);
+                pathResult.put(FIELD_NAME_OF_DIFF_TYPE, DIFF_TYPE_OF_ADD);
+                Object value2 = json2.getByPath(path);
+                pathResult.put("value2", value2);
+                if (value2 instanceof JSONObject || value2 instanceof JSONArray) {
+                    ignorePathPrefixList.add(path);
+                }
+            } else if (json1Contain) {
+                Object value1 = json1.getByPath(path);
+                Object value2 = json2.getByPath(path);
```
wenshao, line 184: **[Suggestion]** `getByPath()` re-parses the path string and re-traverses the JSON tree on every call, even though the flattened maps `jsonPath1`/`jsonPath2` already hold the value at each path key.

Concrete cost: each call allocates a `JSONPath` object and walks the tree from root — O(depth) per call. `jsonPath1.get(path)` is an O(1) HashMap lookup returning the already-stored value. This applies to all three branches (REMOVE line 169, ADD line 177, both-exist lines 183-184).

```suggestion
                Object value1 = jsonPath1.get(path);
                Object value2 = jsonPath2.get(path);
```

`getByPath()` 每次调用都会重新解析路径字符串并重新遍历 JSON 树，尽管展平后的 map `jsonPath1`/`jsonPath2` 已经存储了每个路径键对应的值。

具体开销：每次调用分配一个 `JSONPath` 对象并从根节点遍历——每次 O(深度)。`jsonPath1.get(path)` 是 O(1) 的 HashMap 查找，直接返回已存储的值。此问题适用于所有三个分支。

</details>

core/src/test/java/com/alibaba/fastjson2/util/CompareUtilsTest.java
```diff
@@ -0,0 +1,965 @@
...
+                "'array': [{'number':123,'string':'abc',}]," +
+                "}");
+        boolean result = CompareUtils.equals(json1, json2);
+        System.out.println(result);
+        Object expected = false;
+        assertEquals(expected, result);
+    }
+
+    @Test
+    public void testSum() {
```
wenshao, line 881: **[Suggestion]** `sum()` is only tested with data that produces differences (`equal:false`); the `equal:true` path is never verified. If `sum()`'s `equal` field were initialised to `false`, `testSum()` would still pass green.

Concrete cost: a mutation `boolean equal = true` → `false` survives all tests. Add a test comparing two identical JSONs and asserting `"equal":true`.

`sum()` 仅使用产生差异的数据（`equal:false`）进行测试；`equal:true` 路径从未被验证。如果 `sum()` 的 `equal` 字段初始化为 `false`，`testSum()` 仍然会通过。

具体开销：变异 `boolean equal = true` → `false` 在所有测试中存活。建议添加一个比较两个相同 JSON 并断言 `"equal":true` 的测试。

</details>

core/src/test/java/com/alibaba/fastjson2/util/CompareUtilsPerfTest.java
```diff
@@ -0,0 +1,101 @@
...
+        //初始次数
+        long times = 1024;
+
+        // 2^N
+//        int MAX_SCALE = 21;
+        int MAX_SCALE = 15;
+
+        for (int j = 0; j < times * 10; j++) {
+            CompareUtils.compareToArray(json1, json2);
+        }
```
wenshao, line 88: **[Suggestion]** The performance test invokes `compareToArray` ~160k times but never asserts on the result — it passes regardless of whether the output is correct.

Concrete cost: if a refactor breaks `compareToArray` to return an empty array or wrong values (without throwing), this test still passes green. Add a one-line sanity assertion before the timing loop.

性能测试调用 `compareToArray` 约 16 万次但从未对结果进行断言——无论输出是否正确都会通过。

具体开销：如果重构导致 `compareToArray` 返回空数组或错误值（不抛异常），此测试仍会通过。建议在计时循环前添加一行基本断言。

</details>

wenshao (changes requested):

Issue #2310, opened 2024-03-09
[BUG]LocalTime日期反序列化


zacat:
fastjson 2.0.47

    @Data
    public static class TaskJobVO implements Serializable {
        LocalTime publishTime;
    }

    public static void main(String[] args) {
        String str = "{\"publishTime\": \"9:0:0\"}";
        TaskJobVO taskJobVO = JSONObject.parseObject(str,TaskJobVO.class);
        System.out.println(JSON.toJSONString(taskJobVO));
    }

Exception in thread "main" com.alibaba.fastjson2.JSONException: expect ':', but 125, offset 17, character ", line 1, column 17, fastjson-version 2.0.47 {"publishTime": "9:0:0"}
	at com.alibaba.fastjson2.JSONReaderASCII.readFieldNameHashCode(JSONReaderASCII.java:374)
	at com.alibaba.fastjson2.reader.ORG_1_1_TaskJobVO.readObject(Unknown Source)
	at com.alibaba.fastjson2.JSON.parseObject(JSON.java:786)
	at com.alibaba.fastjson2.JSONObject.parseObject(JSONObject.java:1984)

yanxutao89:
9:0:0这个时间格式有误，换成09:00:00试试看

wenshao:
@yanxutao89 fastjson2支持自动识别日期格式，这种格式也是要支持的

yanxutao89:
> @yanxutao89 fastjson2支持自动识别日期格式，这种格式也是要支持的

那我看着改改

wenshao:
https://oss.sonatype.org/content/repositories/snapshots/com/alibaba/fastjson2/fastjson2/2.0.48-SNAPSHOT/ 问题已修复，请帮忙用2.0.48-SNAPSHOT验证，2.0.48正式版本预计在月末发布

wenshao:
https://github.com/alibaba/fastjson2/releases/tag/2.0.48
请用新版本

PR #2322, opened 2024-03-13
fix support all kinds of LocalTime format for issue #2310

yanxutao89:

fix support all kinds of LocalTime format


wenshao:
不要通过异常做逻辑处理哈

yanxutao89:
> 不要通过异常做逻辑处理哈

行，那我就对现有方法进行调整了

> You can improve the DateUtils#parseLocalTime5 method to achieve this. Do not handle exceptions by catching them. The performance is not good and it is not easy to maintain.

copy that

Issue #2387, opened 2024-04-01
[bug]JSONPath.eval取不到值 


swapdewalkar:
Migrating bug from https://github.com/alibaba/fastjson/issues/4493 as fix need to be made here.
Original issue is from [Ryoui](https://github.com/Ryoui)
`public static void main(String[] args) { String data = "{\"userName\":\"zhangsan\"}"; Object userName1 = JSONPath.read(data, "$.userName"); Object userName2 = JSONPath.eval(data, "$.userName"); System.out.println(userName1); System.out.println(userName2); }`
上述代码，在1.2.83版本输出：
zhangsan
zhangsan
在2.0.47输出
zhangsan
null

PR #2388, opened 2024-04-01
2387: JSONPath eval(string,string) issue fix

swapdewalkar:
Address the issue:
https://github.com/alibaba/fastjson/issues/4493
https://github.com/alibaba/fastjson2/issues/2387

### Summary of your change
If you are passing string as root object to eval its fail as current eval(Object ..) method can not handle it.
Add new method and using extract method to handle this case.
Also added the test case

### DIFF
--- a/fastjson1-compatible/src/test/java/com/alibaba/fastjson/issue_2300/Issue2387Test.java
+++ b/fastjson1-compatible/src/test/java/com/alibaba/fastjson/issue_2300/Issue2387Test.java
@@ -1,0 +1,30 @@
+package com.alibaba.fastjson.issue_2300;
+
+import org.junit.jupiter.api.Test;
+
+
+import static org.junit.jupiter.api.Assertions.*;
+import com.alibaba.fastjson.JSONPath;
+
+public class Issue2387Test {
+    @Test
+    public void test_for_issue() throws Exception {
+        String data = "{\"userName\":\"testname\"}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertEquals(userName1, "testname");
+    }
+    
+    @Test
+    public void test_not_matching() {
+        String data = "{\"userName\":\"testname\"}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertNotEquals(userName1, "testname1");
+    }
+
+    @Test
+    public void test_null() throws Exception {
+        String data = "{\"userName\":null}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertNull(userName1);
+    }
+}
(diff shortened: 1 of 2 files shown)


fastjson1-compatible/src/test/java/com/alibaba/fastjson/issue_2300/Issue2387Test.java
```diff
@@ -0,0 +1,30 @@
...
+        String data = "{\"userName\":\"testname\"}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertEquals(userName1, "testname");
+    }
+    
+    @Test
+    public void test_not_matching() {
+        String data = "{\"userName\":\"testname\"}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertNotEquals(userName1, "testname1");
```
wenshao, line 21: **[Suggestion]** `test_not_matching` does not test a non-matching path — it re-tests the same successful extraction as `test_for_issue` with a trivially-true `assertNotEquals(userName1, "testname1")`. No non-matching path scenario (e.g., `$.nonExistent` → null) is ever exercised. — Concrete cost: if `eval(String, String)` mishandles a path that matches nothing (e.g., throws instead of returning null), this test would not catch it. Mutation analysis: changing `extract(rootObject)` → `eval(rootObject)` causes `test_for_issue` to fail but `test_not_matching` still passes because `assertNotEquals(null, "testname1")` is true.

```suggestion
    public void test_not_matching() {
        String data = "{\"userName\":\"testname\"}";
        Object result = JSONPath.eval(data, "$.nonExistent");
        assertNull(result);
```

wenshao (commented):

fastjson1-compatible/src/test/java/com/alibaba/fastjson/issue_2300/Issue2387Test.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+import com.alibaba.fastjson.JSONPath;
+
+public class Issue2387Test {
+    @Test
+    public void test_for_issue() throws Exception {
+        String data = "{\"userName\":\"testname\"}";
+        Object userName1 = JSONPath.eval(data, "$.userName");
+        assertEquals(userName1, "testname");
```
wenshao, line 14: **[Suggestion]** `assertEquals` arguments are reversed — the actual result (`userName1`) is in the `expected` slot and the literal (`"testname"`) is in the `actual` slot. JUnit 5 signature is `assertEquals(expected, actual)`. Codebase convention in neighboring tests (Issue2358, Issue2371, Issue2348) puts the literal first. — Concrete cost: on a regression where `eval` returns null, the failure message reads `expected: null but was: <testname>` instead of `expected: <testname> but was: null`, misleading about the regression direction. The same reversal applies to `assertNotEquals(userName1, "testname1")` on line 21.

```suggestion
        assertEquals("testname", userName1);
```

PR #2457, opened 2024-04-17
2413 fix

arjun12102019:

JSONObject.from(Object obj) 
Currently we have only method which expects type Object and getting class cast exception when we pass string

### Summary of your change

Overloaded method for String argument and invoking JSON.parseObject(text) to return JSONObject

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
@@ -2022,6 +2022,13 @@ public class JSONObject
     }
 
     /**
+     * See {@link JSON#parseObject} for details
+     */
+    public static JSONObject from(String text) {
+        return JSON.parseObject(text);
+    }
+
+    /**
      * See {@link JSON#toJSON} for details
      */
     public static JSONObject from(Object obj) {


wenshao:
please sign CLA

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
```diff
@@ -2021,6 +2021,13 @@ public static JSONObject parse(String text, JSONReader.Feature... features) {
         return JSON.parseObject(text, features);
     }
 
+    /**
+     * See {@link JSON#parseObject} for details
+     */
+    public static JSONObject from(String text) {
+        return JSON.parseObject(text);
+    }
```
wenshao, line 2029: **[Critical]** This overload does not fix issue #2413's reported scenario. The issue explicitly states the parameter is of `Object` type (dynamic): "传给JSONObject.from的参数值是Object类型，该类型是动态的". Java overload resolution is compile-time — when the caller writes `Object obj = someDynamicValue; JSONObject.from(obj);`, dispatch still goes to `from(Object)` → `(JSONObject) JSON.toJSON(obj)` → `ClassCastException`. The new `from(String)` is never reached.

Even for the literal repro `JSONObject.from("test")`, `JSON.parseObject("test")` throws `JSONException` since `test` is not valid JSON — the user gets a different exception, not a working result.

Additionally, this method is functionally identical to the already-existing `JSONObject.parseObject(String)` at line 2011.

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
```diff
@@ -2021,6 +2021,13 @@ public static JSONObject parse(String text, JSONReader.Feature... features) {
         return JSON.parseObject(text, features);
     }
 
+    /**
+     * See {@link JSON#parseObject} for details
+     */
+    public static JSONObject from(String text) {
```
wenshao, line 2027: **[Suggestion]** Sibling class `JSONArray` has the same `from(Object)` ClassCastException bug (`JSONArray.from("[1,2,3]")` → ClassCastException) but receives no matching `from(String)` overload in this PR. The fix is asymmetric across the two parallel classes.

Concrete cost: a user who sees this fix and applies the same pattern to arrays hits the identical exception.

wenshao, line 2027: **[Suggestion]** No test exercises the new `from(String)` overload. If a future refactor removes it, `JSONObject.from("{\"id\":1}")` silently re-resolves to `from(Object)`, reintroducing the ClassCastException with no test to catch the regression.

Concrete cost: the overload-resolution fix — the actual point of this PR — can regress silently.

Suggested test:
```java
@Test
public void testFromString() {
    JSONObject obj = JSONObject.from("{\"id\":123}");
    assertEquals(123, obj.getIntValue("id"));
    assertNull(JSONObject.from((String) null));
}
```

wenshao (changes requested):
Not reviewed: the executable-script lint — the report is stale or its diff could not be verified; re-run `qwen review script-lint`.

PR #2909, opened 2024-08-30
[Feature] add a DisableSingleQuote flag for JsonReader

wyTrivail:

We are building a function that relies on fastjson2 but in the meantime we are trying to achieve strict mode on parsing. Single quote should not be allowed in our case.

### Summary of your change

This PR adds a Feature in JSONReader to allow customers to disable single quote support, it throws an exception for single quote when this feature is enabled.

- [Done] Made sure tests are passing and test coverage is added if needed.
- [Done] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [Done] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -1974,6 +1974,9 @@ public abstract class JSONReader
     }
 
     public void read(Map object, long features) {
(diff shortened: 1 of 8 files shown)


addyg:
+1, would be good to get this flag.

wyTrivail:
may i know why we merged the PR despite CI was failing?

Issue #2944, opened 2024-09-09
[BUG] Duplicated elements in private final List member when converting JSONObject to Java object via `JSONObject.toJavaObject`


0xderek:
### 问题描述

Duplicated elements in private final List member when converting JSONObject to Java object via `JSONObject.toJavaObject`.

使用 `JSONObject.toJavaObject` 方法将 JSONObject 对象转换为 Java 对象时，private final List 成员中出现重复元素。

### 环境信息

 - OS信息：  Ubuntu 23.10 6.5.0-42-generic
 - JDK信息： Amazon Corretto 1.8.0_402
 - 版本信息：fastjson2 2.0.52

### 重现步骤
*如何操作可以重现该问题：*

```java
// Main.java
public class Main {
    public static void main(String[] args) {
        List<Employee> employees = new ArrayList<>();
        employees.add(new Employee(0, "John"));
        employees.add(new Employee(1, "Jane"));
        employees.add(new Employee(2, "Bob"));
        Department department = new Department("dev", employees);

        String payload = JSON.toJSONString(department);
        System.out.println("payload: " + payload);

        JSONObject jsonObject = JSON.parseObject(payload);
        Department parsed = jsonObject.toJavaObject(Department.class);
        System.out.println("parsed " + parsed);
    }
}

// Department.java
public class Department {
    private final String name;
    private final List<Employee> employees;

    public String getName() {
        return name;
    }

    public List<Employee> getEmployees() {
        return employees;
    }

    public Department(String name, List<Employee> employees) {
        this.name = name;
        this.employees = employees;
    }

    @Override
    public String toString() {
        return "Department{" +
                "name='" + name + '\'' +
                ", employees=" + employees +
                '}';
    }
}

// Employee.java
public class Employee {
    private final Integer id;
    private final String name;

    public Integer getId() {
        return id;
    }

    public String getName() {
        return name;
    }

    public Employee(int id, String name) {
        this.id = id;
        this.name = name;
    }

    @Override
    public String toString() {
        return "Employee{" +
                "id=" + id +
                ", name='" + name + '\'' +
                '}';
    }
}
```

### 期待的正确结果

Parsed object should have the same contents as the original object.

### 相关日志输出

```text
payload: {"employees":[{"id":0,"name":"John"},{"id":1,"name":"Jane"},{"id":2,"name":"Bob"}],"name":"dev"}
[...]

PR #2973, opened 2024-09-19
fix remove duplicated list elements when size are same, for issue #2944

yanxutao89:

fix remove duplicated list elements when size are same, for issue #2944


core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderCollectionMethodReadOnly.java
```diff
@@ -75,6 +75,11 @@ public void accept(T object, Object value) {
         }
 
         Collection values = (Collection) value;
+
+        if (values.size() == collection.size()) {
```
wenshao, line 79: 这个逻辑不对吧？？

Issue #3094, opened 2024-10-12
[BUG] Found 2 Flaky tests in JsonManagedReferenceTest.java


Mukta13:
### Problem Description
Found 2 flaky tests in the file: `JsonManagedReferenceTest.java`

The tests that were flaky within this file were: `testJsonBackReference()` and `test1()`. 

### Environment Info:

 - OS： Ubuntu 22.04.5 LTS, Release: 22.04
 - JDK： java-17-openjdk-amd64
 - Version：Latest version
 

### Steps to reproduce
 
**Run these commands:**

`mvn -pl core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.alibaba.fastjson2.jackson_support.JsonManagedReferenceTest#testJsonBackReference`

`mvn -pl core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.alibaba.fastjson2.jackson_support.JsonManagedReferenceTest#test1`

The results would be:

![image](https://github.com/user-attachments/assets/99c882da-52b5-43fa-98ae-cba4fe7cba8c)

### Expected Results
The tests should run without non-determinism issue and the build should be successful after running the above commands

### Relevant log output
![image](https://github.com/user-attachments/assets/1e15373d-1167-496b-a031-7fe7006ffe19)

Mukta13:
Fix in https://github.com/alibaba/fastjson2/pull/3095

PR #3095, opened 2024-10-12
Fix flaky tests in Issue#3049

Mukta13:

This is a fix for this flaky test issue:

https://github.com/alibaba/fastjson2/issues/3094

### Summary of your change

There were 2 non deterministic tests that were changed using a simple fix. The code compares the deserialization of two JSON strings (jsonJackson and jsonFastjson2) into Map<String, Object> objects using the mapper.readValue() method. Once both JSON strings are converted into maps, the assertEquals() method is used to check if the two resulting maps are identical. This ensures that the JSON content from both sources is equivalent, regardless of the order.

- [ Yes] Made sure tests are passing and test coverage is added if needed.
- [ yes] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [ yes] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.


Mukta13:
I have signed the CLA agreement

PR #3222, opened 2024-12-26
fix remove the field writer which only differ in first character that one is upper case the other is lower case

yanxutao89:

fix remove the field writer which only differ in first character that one is upper case the other is lower case


wenshao:
没有单测？

Issue #3224, opened 2024-12-28
[BUG] JSONWriter.writeNameValue is broken


mpfaff:
### 问题描述

The implementation of `JSONWriter.writeNameValue` is broken. It is missing a call to `writeColon` between the `writeName` and `writeAny`.

### 环境信息

 - OS信息：  macOS Sequoia
 - JDK信息： OpenJDK 21
 - 版本信息：Fastjson2 2.0.53
 

### 重现步骤

1. Call `writeNameValue("foo", "bar")` on a `JSONWriter`
2. Observe that the written JSON is missing a colon
   ```
   "foo""bar"
   ```

### 期待的正确结果

```
"foo": "bar"
```

### 相关日志输出

N/A

#### 附加信息

N/A

PR #3405, opened 2025-03-16
fix: add missing call to `writeColon` in `JSONWriter.writeNameValue`

mpfaff:

See #3224:
> The implementation of `JSONWriter.writeNameValue` is broken. It is missing a call to `writeColon` between the `writeName` and `writeAny`.

### Summary of your change

Add the aforementioned missing call, fixing the issue.

If I recall correctly, I tried running tests at the time I implemented this fix but they failed due to (seemingly) relying on the host system having a particular time zone.


wenshao:
请签署CLA

PR #3513, opened 2025-04-22
fix not write default value for primitive types, for issue #3186

yanxutao89:

fix not write default value for primitive types, for issue #3186


core/src/main/java/com/alibaba/fastjson2/internal/asm/MethodWriter.java
```diff
@@ -301,6 +301,19 @@ public void i2d() {
         visitInsn(Opcodes.I2D);
     }
 
+    public void o2i(Class<?> fieldClass) {
```
wenshao: 这个方法名不是很好

yanxutao89: 好的，后面我调整下

PR #3536, opened 2025-05-07
Fix 2478

yanxutao89:

fix enable serialize and deserialize when using PropertyNamingStrategy

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSON.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSON.java
@@ -896,7 +896,7 @@ public interface JSON {
 
         final JSONReader.Context context = JSONFactory.createReadContext(filter, features);
         boolean fieldBased = (context.features & JSONReader.Feature.FieldBased.mask) != 0;
-        ObjectReader<T> objectReader = context.provider.getObjectReader(clazz, fieldBased);
+        ObjectReader<T> objectReader = context.provider.getObjectReader(clazz, fieldBased, filter);
 
         try (JSONReader reader = JSONReader.of(text, context)) {
             if (reader.nextIfNull()) {
@@ -911,6 +911,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -947,7 +949,7 @@ public interface JSON {
         context.setDateFormat(format);
 
         boolean fieldBased = (context.features & JSONReader.Feature.FieldBased.mask) != 0;
-        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased);
+        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased, filters);
 
         try (JSONReader reader = JSONReader.of(text, context)) {
             if (reader.nextIfNull()) {
@@ -962,6 +964,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -1133,7 +1137,7 @@ public interface JSON {
         final JSONReader.Context context = JSONFactory.createReadContext(filter, features);
         Type type = typeReference.getType();
         boolean fieldBased = (context.features & JSONReader.Feature.FieldBased.mask) != 0;
-        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased);
+        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased, filter);
 
         try (JSONReader reader = JSONReader.of(text, context)) {
             T object = objectReader.readObject(reader, type, null, 0);
@@ -1144,6 +1148,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -1331,7 +1337,7 @@ public interface JSON {
         }
 
         final JSONReader.Context context = JSONFactory.createReadContext(filter, features);
-        ObjectReader<T> objectReader = context.getObjectReader(type);
+        ObjectReader<T> objectReader = context.getObjectReader(type, filter);
 
         try (JSONReader reader = JSONReader.of(text, context)) {
             T object = objectReader.readObject(reader, type, null, 0);
@@ -1342,6 +1348,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -1566,7 +1574,7 @@ public interface JSON {
 
         final JSONReader.Context context = JSONFactory.createReadContext(filter, features);
         boolean fieldBased = (context.features & JSONReader.Feature.FieldBased.mask) != 0;
-        ObjectReader<T> objectReader = context.provider.getObjectReader(clazz, fieldBased);
+        ObjectReader<T> objectReader = context.provider.getObjectReader(clazz, fieldBased, filter);
 
         try (JSONReader reader = JSONReader.of(bytes, context)) {
             T object = objectReader.readObject(reader, clazz, null, 0);
@@ -1577,6 +1585,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -1649,7 +1659,7 @@ public interface JSON {
         context.setDateFormat(format);
 
         boolean fieldBased = (context.features & JSONReader.Feature.FieldBased.mask) != 0;
-        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased);
+        ObjectReader<T> objectReader = context.provider.getObjectReader(type, fieldBased, filters);
 
         try (JSONReader reader = JSONReader.of(bytes, context)) {
             T object = objectReader.readObject(reader, type, null, 0);
@@ -1660,6 +1670,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
@@ -1805,7 +1817,7 @@ public interface JSON {
         }
 
         final JSONReader.Context context = JSONFactory.createReadContext(filter, features);
-        final ObjectReader<T> objectReader = context.getObjectReader(type);
+        final ObjectReader<T> objectReader = context.getObjectReader(type, filter);
 
         try (JSONReader reader = JSONReader.of(bytes, context)) {
             T object = objectReader.readObject(reader, type, null, 0);
@@ -1816,6 +1828,8 @@ public interface JSON {
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
         }
     }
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/alibaba/fastjson2/JSON.java
```diff
@@ -911,6 +911,8 @@ static <T> T parseObject(
                 throw new JSONException(reader.info("input not end"));
             }
             return object;
+        } finally {
+            context.provider.setNameFilter(null);
```
wenshao, line 915: 如果Context被多个线程同时使用，这里有问题的

wenshao:
我担心在ReaderProvider上增加NameFilter支持，对性能有很大影响

yanxutao89:
> 我担心在ReaderProvider上增加NameFilter支持，对性能有很大影响

这个也仅针对存在NameFilter的情况吧，影响范围还算可控。

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -764,6 +783,32 @@ public ObjectReader getObjectReader(Type objectType, boolean fieldBased) {
...
+                objectType = upperBounds[0];
+            }
+        }
+
+        synchronized (objectType.getClass()) {
+            Filter curFilter = fieldBased
+                    ? cacheFieldBasedFilter.get(objectType)
+                    : cacheFilter.get(objectType);
+            if (!filter.equals(curFilter)) {
+                cleanup((Class) objectType);
```
wenshao, line 801: **[Critical]** `cleanup((Class) objectType)` throws `ClassCastException` when `objectType` is a `ParameterizedType` (e.g., from `TypeReference<List<User>>`). The method accepts `Type` but casts to `Class` unconditionally. — Failure scenario: `JSON.parseObject(text, new TypeReference<List<User>>(){}.getType(), NameFilter.of(SnakeCase))` → `ClassCastException` at runtime. Multiple `parseObject` overloads accept `Type`/`TypeReference` with filters (JSON.java lines 1140, 1340, 1662, 1820).

```suggestion
                if (objectType instanceof Class) {
                    cleanup((Class) objectType);
                } else {
                    cache.remove(objectType);
                    cacheFieldBased.remove(objectType);
                    cacheFilter.remove(objectType);
                    cacheFieldBasedFilter.remove(objectType);
                }
```

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -764,6 +783,32 @@ public ObjectReader getObjectReader(Type objectType, boolean fieldBased) {
...
+            }
+            if (fieldBased) {
+                cacheFieldBasedFilter.put(objectType, filter);
+            } else {
+                cacheFilter.put(objectType, filter);
+            }
+            return getObjectReader(objectType, fieldBased);
+        }
+    }
```
wenshao, line 810: **[Critical]** Filter-built `ObjectReader` is cached in the shared `cache`/`cacheFieldBased` map. Subsequent no-filter calls to `getObjectReader(Type, boolean)` find and return the filter-transformed reader, causing silent field-name mismatch. — Failure scenario: (1) `JSON.parseObject(snakeJson, User.class, NameFilter.of(SnakeCase))` caches a snake_case reader. (2) `JSON.parseObject(camelJson, User.class)` (no filter) gets the cached snake_case reader → `firstName` doesn't match `first_name` → null fields. The `cacheFilter` map is never consulted on the no-filter path.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -148,6 +150,8 @@ public ObjectReaderCachePair(long hashCode, ObjectReader reader) {
 
     final ConcurrentMap<Type, ObjectReader> cache = new ConcurrentHashMap<>();
     final ConcurrentMap<Type, ObjectReader> cacheFieldBased = new ConcurrentHashMap<>();
+    final ConcurrentMap<Type, Filter> cacheFilter = new ConcurrentHashMap<>();
```
wenshao, line 153: **[Critical]** `clear()` (line 369) does not clear the new `cacheFilter` and `cacheFieldBasedFilter` maps. After `clear()`, stale filter entries cause `getObjectReaderByFilter` to skip `cleanup` (filter matches stale entry) and return an unfiltered reader from the rebuilt cache. — Failure scenario: filtered parse → `clear()` → unfiltered parse rebuilds cache → same-filter parse skips cleanup → gets unfiltered reader → null fields.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -764,6 +783,32 @@ public ObjectReader getObjectReader(Type objectType, boolean fieldBased) {
...
+            }
+        }
+
+        synchronized (objectType.getClass()) {
+            Filter curFilter = fieldBased
+                    ? cacheFieldBasedFilter.get(objectType)
+                    : cacheFilter.get(objectType);
+            if (!filter.equals(curFilter)) {
+                cleanup((Class) objectType);
+            }
```
wenshao, line 802: **[Critical]** `cleanup(Class)` permanently destroys mixin registrations (`mixInCache.remove`) and clears `BeanUtils` cache as side effects. Before this diff, `cleanup` was never called internally during ordinary deserialization. — Failure scenario: user registers a mixin → deserializes with NameFilter → `cleanup` removes mixin → all subsequent deserializations (even without filter) silently lose mixin behaviour. Consider replacing with targeted reader-cache invalidation only (`cache.remove`/`cacheFieldBased.remove`) instead of the broad `cleanup` call.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -764,6 +783,32 @@ public ObjectReader getObjectReader(Type objectType, boolean fieldBased) {
...
+        if (objectType == null) {
+            objectType = Object.class;
+        } else if (objectType instanceof WildcardType) {
+            Type[] upperBounds = ((WildcardType) objectType).getUpperBounds();
+            if (upperBounds.length == 1) {
+                objectType = upperBounds[0];
+            }
+        }
+
+        synchronized (objectType.getClass()) {
```
wenshao, line 796: **[Suggestion]** `synchronized (objectType.getClass())` acquires a global lock when `objectType` is a `Class` (the common case) — `objectType.getClass()` is `java.lang.Class`, a single monitor shared by every class in the JVM. All filtered reader creations for all types are serialized on this one lock during expensive reader construction (reflection + ASM). — Concrete cost: multi-threaded server deserializing User, Order, Payment with NameFilters → all serialize on one lock → throughput collapse.

```suggestion
            synchronized (objectType) {
```

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -764,6 +783,32 @@ public ObjectReader getObjectReader(Type objectType, boolean fieldBased) {
...
+            Filter curFilter = fieldBased
+                    ? cacheFieldBasedFilter.get(objectType)
+                    : cacheFilter.get(objectType);
+            if (!filter.equals(curFilter)) {
```
wenshao, line 800: **[Suggestion]** `NameFilter.of(PropertyNamingStrategy)` returns a new lambda per call. Lambdas use identity equality, so `filter.equals(curFilter)` is always `false` for distinct instances of the same strategy, defeating the cache guard and triggering `cleanup` + full reader rebuild on every call. — Concrete cost: per-request `NameFilter.of(SnakeCase)` → cache miss every time → reflection + ASM on every deserialization. Consider making `NameFilter.of` return a cached instance with strategy-based `equals`/`hashCode`.

core/src/main/java/com/alibaba/fastjson2/JSONFactory.java
```diff
@@ -434,6 +435,10 @@ public static JSONReader.Context createReadContext(Filter filter, JSONReader.Fea
             context.extraProcessor = (ExtraProcessor) filter;
         }
 
+        if (filter instanceof NameFilter) {
+            provider.setNameFilter((NameFilter) filter);
+        }
```
wenshao, line 440: **[Suggestion]** NameFilter is stored in a ThreadLocal on the JVM-wide singleton provider, but cleanup (`setNameFilter(null)`) only occurs in seven `JSON.parseObject` `finally` blocks. Other set-paths — direct `JSONReader.Context` construction, `config(Filter[])` — leave the ThreadLocal stale. — Concrete cost: user constructs `JSONReader.Context` directly with NameFilter → no cleanup → thread returns to pool → next request's reader creation picks up stale filter → wrong field names → silent data loss.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -4257,6 +4263,10 @@ public void config(Filter[] filters) {
                 if (filter instanceof ExtraProcessor) {
                     extraProcessor = (ExtraProcessor) filter;
                 }
+
+                if (filter instanceof NameFilter) {
+                    provider.setNameFilter((NameFilter) filter);
+                }
```
wenshao, line 4269: **[Suggestion]** NameFilter handling is added to `config(Filter[])` but not to the three sibling overloads: `config(Filter)`, `config(Filter, Feature...)`, and `config(Filter[], Feature...)`. The `Context(ObjectReaderProvider, Filter, Feature...)` constructor delegates to `config(Filter)`, so a NameFilter passed through it is silently ignored. — Concrete cost: `new JSONReader.Context(provider, NameFilter.of(SnakeCase))` → NameFilter not set → deserialization ignores naming strategy.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -148,6 +150,8 @@ public ObjectReaderCachePair(long hashCode, ObjectReader reader) {
 
     final ConcurrentMap<Type, ObjectReader> cache = new ConcurrentHashMap<>();
     final ConcurrentMap<Type, ObjectReader> cacheFieldBased = new ConcurrentHashMap<>();
+    final ConcurrentMap<Type, Filter> cacheFilter = new ConcurrentHashMap<>();
+    final ConcurrentMap<Type, Filter> cacheFieldBasedFilter = new ConcurrentHashMap<>();
```
wenshao, line 154: **[Suggestion]** `cleanup(ClassLoader)` (line 427) does not purge the new `cacheFilter`/`cacheFieldBasedFilter` maps. In app-server environments (Tomcat, WildFly), `Class` keys from the old classloader remain strongly reachable, preventing classloader GC and causing Metaspace leaks across redeploy cycles. — Concrete cost: N redeploys → N classloaders retained → `OutOfMemoryError: Metaspace`.

wenshao (changes requested):
**[Critical]** ThreadLocal NameFilter leak when getObjectReader throws (JSON.java:~899): getObjectReader is called BEFORE the try block whose finally clears the ThreadLocal. If getObjectReader throws (e.g., the ClassCastException from the ParameterizedType bug), the finally block is never reached, and the stale NameFilter contaminates subsequent reader creation on the same thread. All 7 modified parseObject overloads in JSON.java have this pattern.

**[Critical]** NameFilter overrides explicit @JSONField(name=...) annotations (ObjectReaderCreator.java:1380-1382): The NameFilter is applied unconditionally to ALL field names, including those explicitly set via @JSONField(name=...) annotations. The old namingStrategy was only applied inside the if (fieldInfo.fieldName == null) block, so annotations were respected. The new NameFilter is applied outside that block, transforming explicitly-named fields (e.g., "userId" → "user_id"), breaking deserialization for beans that combine annotations with NameFilter.

Issue #117, opened 2022-05-01
JSONPath eval  解析 不兼容 报异常


wsxe9988:
示例代码1
//1000000=>325,321,319
 //
 //1.加载json
String text = ("{code:1,msg:'Hello world',data:{list:[1,2,3,4,5], ary2:[{a:2},{a:3,b:{c:'ddd'}}]}}");
JSONObject obj = JSON.parseObject(text);

 JSONPath.eval(obj,"$..ary2[0].a");

 long start = System.currentTimeMillis();
for(int i=0,len=1000000; i<len; i++) {
      JSONPath.eval(obj,"$..ary2[0].a");
}

long times = System.currentTimeMillis() - start;

System.out.println(times);

assert times > 0;

com.alibaba.fastjson2.JSONException: not support : class java.lang.Integer

	at com.alibaba.fastjson2.JSONPath$NameSegment.eval(JSONPath.java:3218)
	at com.alibaba.fastjson2.JSONPath$MultiSegmentPath.eval(JSONPath.java:2620)
	at com.alibaba.fastjson2.JSONPath.eval(JSONPath.java:53)
	at com.alibaba.fastjson_perf.JSONPathTest2.test41(JSONPathTest2.java:106)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
[...]

PR #3561, opened 2025-05-16
fix enhance operator in and support nested json path, for issue #177

yanxutao89:

fix enhance operator in and support nested json path

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
@@ -7,6 +7,7 @@ import com.alibaba.fastjson2.writer.FieldWriter;
 import com.alibaba.fastjson2.writer.ObjectWriter;
 import com.alibaba.fastjson2.writer.ObjectWriterAdapter;
 
+import java.lang.reflect.Array;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.*;
@@ -1315,7 +1316,7 @@ abstract class JSONPathFilter
 
     static final class NameIntInSegment
             extends NameFilter {
-        private final long[] values;
+        private long[] values;
         private final boolean not;
 
         public NameIntInSegment(
@@ -1392,6 +1393,37 @@ abstract class JSONPathFilter
             }
 
             return not;
+        }
+
+        @Override
+        public void convert(Object root) {
+            JSONPath parentPath = getParentPath();
+            if (parentPath != null) {
+                Object eval = parentPath.eval(root);
+                int length = 1;
+                if (eval != null) {
+                    if (eval.getClass().isArray()) {
+                        length = Array.getLength(eval);
+                        values = new long[length];
+                        for (int i = 0; i < length; ++i) {
+                            values[i] = Long.parseLong(Array.get(eval, i).toString());
+                        }
+                    } else if (Collection.class.isAssignableFrom(eval.getClass())) {
+                        Long[] longs = new ArrayList<>((Collection<?>) eval).stream()
+                                .filter(Objects::nonNull)
+                                .map(v -> Long.parseLong(v.toString()))
+                                .toArray(Long[]::new);
+                        length = longs.length;
+                        values = new long[length];
+                        for (int i = 0; i < length; ++i) {
+                            values[i] = longs[i];
+                        }
+                    } else {
+                        values = new long[length];
+                        values[0] = Long.parseLong(eval.toString());
+                    }
+                }
+            }
         }
     }
 
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathMulti.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathMulti.java
@@ -188,6 +188,7 @@ final class JSONPathMulti
                     ((JSONPathFilter.NameFilter) segment).excludeArray();
                 }
             }
+            segment.convert(root);
             segment.eval(context);
         }
 
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathParser.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathParser.java
@@ -640,6 +640,7 @@ class JSONPathParser {
 
         String fieldName = null;
         long hashCode = 0;
+        boolean hasOperator = false;
         if (at) {
             if (jsonReader.ch == '[') {
                 JSONPathSegment segment = parseArrayAccess();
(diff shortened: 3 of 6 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
```diff
@@ -1393,6 +1394,37 @@ public boolean apply(Object fieldValue) {
...
+                        length = Array.getLength(eval);
+                        values = new long[length];
+                        for (int i = 0; i < length; ++i) {
+                            values[i] = Long.parseLong(Array.get(eval, i).toString());
+                        }
+                    } else if (Collection.class.isAssignableFrom(eval.getClass())) {
+                        Long[] longs = new ArrayList<>((Collection<?>) eval).stream()
+                                .filter(Objects::nonNull)
+                                .map(v -> Long.parseLong(v.toString()))
+                                .toArray(Long[]::new);
```
wenshao, line 1415: 这里可以去掉对Stream的使用么？

core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
```diff
@@ -1393,6 +1394,37 @@ public boolean apply(Object fieldValue) {
...
             return not;
         }
+
+        @Override
+        public void convert(Object root) {
+            JSONPath parentPath = getParentPath();
+            if (parentPath != null) {
+                Object eval = parentPath.eval(root);
+                int length = 1;
+                if (eval != null) {
```
wenshao, line 1404: **[Critical]** When the nested path matches nothing, `parentPath.eval(root)` returns `null`, this `if (eval != null)` block is skipped, and `values` stays `null` (the parser passes `null` for `values` when building the `$`-path segment). `apply()` then runs `for (long value : values)` → `NullPointerException`. — Failure scenario: `JSONPath.eval(object, "$.data.list[?(@ in $..nonexistent)]")` against an object with no `nonexistent` field → NPE instead of an empty result. A reused segment can also silently retain a stale non-null `values` from a previous eval when a later eval's path matches nothing.

```suggestion
                Object eval = parentPath.eval(root);
                int length = 1;
                if (eval != null) {
```
(add an `else { values = new long[0]; }` branch so `apply` matches nothing when the referenced path does not exist)

core/src/main/java/com/alibaba/fastjson2/JSONPathParser.java
```diff
@@ -820,12 +829,30 @@ JSONPathSegment parseFilter() {
...
             case IN:
             case NOT_IN: {
-                if (jsonReader.ch != '(') {
+                JSONPathSegment segment;
+                char ch = jsonReader.ch;
+                if (ch == '$') {
+                    StringBuilder sb = new StringBuilder();
+                    int filterNests = 0;
+                    while (ch != ')' || filterNests != 0) {
+                        sb.append(ch);
```
wenshao, line 838: **[Critical]** This `$`-path scanning loop has no EOI (end-of-input) guard. On malformed input missing the closing `)`, the reader reaches `EOI` (0x1A) and `JSONReader.next()` keeps `ch == EOI` without advancing; since `EOI != ')'`, the condition stays true forever and `sb.append(ch)` grows without bound → the thread hangs and the JVM eventually throws `OutOfMemoryError`. Every other scanning loop in this parser guards on `EOI` or tests a character `EOI` cannot match. — Failure scenario: `JSONPath.of("$.data.list[?(@ in $..ary2[0].a")` (no trailing `)`) → infinite loop. This is a denial-of-service vector for any application that parses user-supplied JSONPath expressions.

```suggestion
                    int filterNests = 0;
                    while (ch != ')' || filterNests != 0) {
                        if (ch == EOI) {
                            throw new JSONException(jsonReader.info("jsonpath syntax error"));
                        }
                        sb.append(ch);
```

core/src/main/java/com/alibaba/fastjson2/JSONPathMulti.java
```diff
@@ -188,6 +188,7 @@ public Object eval(Object root) {
                     ((JSONPathFilter.NameFilter) segment).excludeArray();
                 }
             }
+            segment.convert(root);
             segment.eval(context);
```
wenshao, line 192: **[Critical]** `convert(root)` — which populates the otherwise-`null` `values` for the `$`-path syntax — is added here and in `JSONPathTwoSegment.eval()`, but not in `JSONPathSingle.eval()`, `JSONPathTwoSegment.extract(JSONReader)`, or `JSONPathMulti.remove()` / `contains()` / `set()` / `extract()` / `extractScalar()`. For the `$` syntax `values` is `null` at construction, so any of those paths reaches `apply()` with null `values` → `NullPointerException`. — Failure scenario: `JSONPath.eval(root, "$[?(@ in $..ary[0].a)]")` produces a `JSONPathSingle` whose `eval()` calls `segment.eval(context)` without `convert(root)` → NPE; likewise `JSONPath.of("$.data.list[?(@ in $..ary2[0].a)]").extract(jsonReader)` via `JSONPathTwoSegment.extract()`.

Suggested fix: call `convert(root)` before every `eval()` that can reach a `NameIntInSegment`, or resolve lazily inside `apply()` when `parentPath != null && values == null`.

core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
```diff
@@ -1315,7 +1316,7 @@ public boolean apply(Object fieldValue) {
 
     static final class NameIntInSegment
             extends NameFilter {
-        private final long[] values;
+        private long[] values;
         private final boolean not;
```
wenshao, line 1320: **[Critical]** Removing `final` and overwriting `values` inside `convert()` during `eval()` makes this segment unsafe for concurrent evaluation of a shared compiled `JSONPath`. A `JSONPath` is typically parsed once (via `JSONPath.of`) and reused across threads; two threads evaluating the same instance with different roots race on this field — one thread's `apply()` can read the values another thread's `convert()` just wrote → silently incorrect filter results. Before this change the field was `final` and the segment was immutable after construction. — Failure scenario: threads A and B call `path.eval(rootA)` / `path.eval(rootB)` on the same `$.data.list[?(@ in $..allowed)]` concurrently → thread A filters with root B's values.

Suggested fix: compute the resolved `long[]` as a local and pass it through the eval `Context`, or create a per-evaluation copy of the segment, so the shared `JSONPath` stays immutable.

core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
```diff
@@ -1393,6 +1394,37 @@ public boolean apply(Object fieldValue) {
...
+            if (parentPath != null) {
+                Object eval = parentPath.eval(root);
+                int length = 1;
+                if (eval != null) {
+                    if (eval.getClass().isArray()) {
+                        length = Array.getLength(eval);
+                        values = new long[length];
+                        for (int i = 0; i < length; ++i) {
+                            values[i] = Long.parseLong(Array.get(eval, i).toString());
+                        }
```
wenshao, line 1410: **[Suggestion]** The native-array branch calls `Long.parseLong(Array.get(eval, i).toString())` with no null-element guard, while the sibling `Collection` branch below uses `.filter(Objects::nonNull)`. `Array.get(eval, i)` returns `null` for a null element, so `.toString()` throws `NullPointerException`. — Concrete cost: when the nested path resolves to a Java array containing a null (e.g. `new Integer[]{1, null, 3}`) this branch crashes, whereas a `List` with the same null is skipped — the two branches diverge on identical logical input.

```suggestion
                        for (int i = 0; i < length; ++i) {
                            Object item = Array.get(eval, i);
                            if (item != null) {
                                values[count++] = Long.parseLong(item.toString());
                            }
                        }
```
(skip nulls, mirroring the Collection branch; size `values` to `count` afterwards)

core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
wenshao, line 1415: **[Suggestion]** This conversion allocates an `ArrayList` copy of the collection, a Stream pipeline, `N` boxed `Long`s, a `Long[]`, and then a second copy into `long[]`. A single loop over the collection into a pre-sized `long[]` (skipping nulls) needs 1–2 allocations and zero boxing. — Concrete cost: ~N+3 extra allocations and N boxing operations per evaluation. This also addresses the maintainer's open question on this PR about removing the Stream usage.

```suggestion
                        Collection<?> coll = (Collection<?>) eval;
                        long[] tmp = new long[coll.size()];
                        int idx = 0;
                        for (Object v : coll) {
                            if (v != null) {
                                tmp[idx++] = Long.parseLong(v.toString());
                            }
                        }
                        values = idx == tmp.length ? tmp : Arrays.copyOf(tmp, idx);
```

core/src/test/java/com/alibaba/fastjson2/issues/Issue117.java
```diff
@@ -23,4 +25,14 @@ public void test() {
 
         System.out.println(times);
     }
+
+    @Test
+    public void test2() {
```
wenshao, line 30: **[Suggestion]** The new test leaves several new-code paths unexercised: (a) a non-matching nested path such as `@ in $..nonexistent` — this would catch the null-`values` NPE in `convert`; (b) `not in` with a nested `$` path — a mutation flipping the `not` flag (`operator == Operator.NOT_IN`) would pass undetected; (c) the native Java array branch of `convert` (`eval.getClass().isArray()`) is unreachable here because JSON-parsed arrays are `JSONArray` (a `Collection`), so a bug in the `Array.get` loop would ship undetected. — Concrete cost: each of these mutations/bugs ships with green tests.

Suggested fix: add assertions for a non-matching path (expect `[]`), a `not in $..ary2[0].a` case, and a bean with a native array field (e.g. `public int[] nums = {2, 4};`).

core/src/test/java/com/alibaba/fastjson2/issues/Issue117.java
```diff
@@ -23,4 +25,14 @@ public void test() {
...
+        String text = ("{code:1,msg:'Hello world',data:{list:[1,2,3,4,5], ary2:[{a:2},{a:[2]},{a:3,b:{c:'ddd'}}]}}");
+        JSONObject object = JSON.parseObject(text, JSONReader.Feature.AllowUnQuotedFieldNames);
+        assertEquals("[2]", JSONPath.eval(object, "$.data.list[?(@ in (2))]").toString());
+        assertEquals("[2]", JSONPath.eval(object, "$.data.list[?(@ in [2])]").toString());
+        assertEquals("[2]", JSONPath.eval(object, "$.data.list[?(@ in $..ary2[0].a)]").toString());
```
wenshao, line 35: **[Suggestion]** The PR title says "for issue #177", but the fix and this test target issue #117 ("JSONPath eval 解析 不兼容 报异常", sample 2: `@ in $..ary2[0].a` throwing `JSONException: syntax error, $`). Issue #177 is an unrelated `JSONObject.put()` generic-warning report, and `closingIssuesReferences` is empty, so neither issue is auto-linked or closed on merge. — Concrete cost: issue #117 stays open with no linked fix, and #177 receives a confusing cross-reference.

Suggested fix: reference #117 in the title and add `fixes #117` to the PR description so GitHub links and closes the correct issue.

core/src/main/java/com/alibaba/fastjson2/JSONPathParser.java
```diff
@@ -820,12 +829,30 @@ JSONPathSegment parseFilter() {
...
+                        if (ch == '(') {
+                            filterNests++;
+                        } else if (ch == ')') {
+                            filterNests--;
+                        }
+                        jsonReader.next();
+                        ch = jsonReader.ch;
+                    }
+                    jsonReader.next();
+                    JSONPath parentPath = JSONPath.of(sb.toString());
```
wenshao, line 848: **[Critical]** The `$`-path scanning loop only terminates on a balancing `)`, and this branch `return`s early — bypassing the `parseFilterRest` call that the literal IN branch uses for compound conditions. As a result a compound filter appends the operator and the entire right-hand expression into the path string. — Failure scenario: `JSONPath.eval(obj, "$.list[?(@ in $..ary2[0].a && @ > 2)]")` → the loop accumulates `$..ary2[0].a&&@>2` → `JSONPath.of(...)` throws `JSONException("not support &")`. The literal form `@ in (2) && @ > 2` works, so compound `in $path` is not merely unsupported but actively throws.

Suggested fix: also terminate the loop on `&` / `|` (and the `and` / `or` keywords) without appending them, then route through `parseFilterRest` the way the literal branch does.

core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
```diff
@@ -1393,6 +1394,37 @@ public boolean apply(Object fieldValue) {
...
+                                .map(v -> Long.parseLong(v.toString()))
+                                .toArray(Long[]::new);
+                        length = longs.length;
+                        values = new long[length];
+                        for (int i = 0; i < length; ++i) {
+                            values[i] = longs[i];
+                        }
+                    } else {
+                        values = new long[length];
+                        values[0] = Long.parseLong(eval.toString());
```
wenshao, line 1423: **[Critical]** `convert()` populates `values` with `Long.parseLong(element.toString())` in all three branches (array, Collection, and this scalar branch). For a non-integer numeric the `toString` is not a valid `long` (e.g. `BigDecimal`/`Double` `2.0` → `"2.0"`), so `Long.parseLong` throws an uncaught `NumberFormatException`; a non-numeric resolution (e.g. a string or object) fails the same way. The literal IN path uses `Number.longValue()` (truncating and Number-tolerant), so `@ in (2.0)` is accepted while the equivalent `$`-path form crashes. — Failure scenario: data `{"ary":[{"a":2.0}]}`, path `$.list[?(@ in $..ary[0].a)]` → `NumberFormatException`.

```suggestion
                        values = new long[length];
                        values[0] = eval instanceof Number ? ((Number) eval).longValue() : Long.parseLong(eval.toString());
```
(apply the same Number-tolerant conversion in the array and Collection branches)

wenshao (changes requested):
Not reviewed: build-and-test — no JDK installed on this machine, so the local build, test run, and test-efficacy probe could not execute (CI's single check reported all_pass).

Not reviewed: reverse audit — no auditor was launched with a prompt this skill builds — the pass that hunts what the rest of the review missed ran, if at all, without the method its brief carries.

PR #3694, opened 2025-08-04
fix: fix memory alignment issues on SPARC platform #3693

mengnankkkk:
SIGBUS (BUS_ADRALN) crash after upgrading from Fastjson2 2.0.32 to 2.0.57 on Oracle Solaris 11.4 SPARC platform
The IOUtils.getLongLE method uses Unsafe.getLong for unaligned memory access, and the SPARC architecture requires 8-byte aligned access.


wenshao:
I don't have a SPACE environment, so I can't verify that this is correct. 

Also, this PR makes getLongLE more complicated, which may affect performance.

PR #3745, opened 2025-08-21
Feature/issue 3690 default charset utf8

wenshao:
This PR sets the default charset to UTF-8 for FastJsonHttpMessageConverter in Spring5, Spring6, and Fastjson1-compatible versions.

  Changes made:
   1. Set defaultCharset to UTF-8 in FastJsonHttpMessageConverter constructors for all versions2. Updated documentation to reflect that
      manual setting of defaultCharset is no longer required from version 2.0.593. Added tests to verify defaultCharset is correctly set to
      UTF-8
   4. Refactored tests to use StandardCharsets.UTF_8 instead of Charset.forName("UTF-8")

  Benefits:
   - Users no longer need to manually configure defaultCharset in each project
   - Improves framework usability and reduces configuration overhead


VictorZeng:
LGTM

PR #3930, opened 2026-01-03
feat: add PrettyFormatInlineArrays option to keep arrays on single line #2972

### LINKED ISSUE #2972: [QUESTION]如何让fastjson2的PrettyFormat输出达到fastjson之前一样的效果？
`JSON.toJSONString(jsonObject, JSONWriter.Feature.PrettyFormat)`
```json
{
	"type":1001,
	"job_id":[
		0,
		45732
	],
	"id":[
		1297059520,
		4117498193
	],
	"is_client":false,
	"stream_count":0
}
```

> FastJson中格式化输出样式
`JSON.toJSONString(jsonObject, Boolean.TRUE)`
```json
{
	"type":1001,
	"stream_count":0,
	"job_id":[0,45732],
	"id":[1297059520,4117498193],
	"is_client":false
}
```
* 我该如何在fastjson2中做到fastjson之前的输出样式，不想让Json中数组的val展开换行
>因为项目有解析前后对比换行后样式错乱

coffeeandwork:

Fixes #2972

Added a new Context configuration option that, when used with PrettyFormat, keeps array elements on a single line instead of expanding each element to its own line. This matches the original fastjson behavior.

### Summary of your change

  - Added `Context.setPrettyFormatInlineArrays(boolean)` for per-call configuration
  - Added `JSONFactory.setDefaultWriterPrettyFormatInlineArrays(boolean)` for global default
  - Added `levelArray` bitmask to track array vs object context at each level
  - Modified `startArray`/`endArray`/`writeComma` in JSONWriterUTF8 and JSONWriterUTF16
  - Added comprehensive tests for edge cases

  ### Usage
  ```java
  JSONWriter.Context context = new JSONWriter.Context(JSONWriter.Feature.PrettyFormat);
  context.setPrettyFormatInlineArrays(true);
  String json = JSON.toJSONString(obj, context);

  // Or set global default
  JSONFactory.setDefaultWriterPrettyFormatInlineArrays(true);
```
  Example output
```json
  // PrettyFormat only:
  {
  	"id": 123,
  	"values": [
  		1,
  		2,
  		3
  	]
  }

  // PrettyFormat + setPrettyFormatInlineArrays(true):
  {
  	"id": 123,
  	"values": [1,2,3]
  }
```
  Uses Context configuration instead of a Feature bit to conserve the limited bits in the 64-bit Feature bitmask.


coffeeandwork:
I have signed the CLA. Not sure why its still showing as not signed yet. @wenshao help, what am i doing wrong?

jujn:
Hi @coffeeandwork ,
Thank you for submitting this PR! I see that the code implementation is very complete, and the test cases are also well-covered.

However, we are currently facing a tricky architectural limitation. In order to pursue ultimate performance, Fastjson2 uses a long (64-bit) to store all JSONWriter.Feature values. Currently, the available bits for this long are already very tight (only a dozen left).

Considering future scalability, we need to reserve these remaining bits for features that are practical, related to functional correctness, security, or core performance. Since PrettyFormatInlineArrays is primarily a formatting beautification feature, which is more of an "addition to excellence", we are hesitant to consume a permanent Feature bit.

coffeeandwork:
> Hi @coffeeandwork , Thank you for submitting this PR! Bringing back this behavior from Fastjson 1.x is very meaningful. I see that the code implementation is very complete, and the test cases are also well-covered.
> 
> However, we are currently facing a tricky architectural limitation. In order to pursue ultimate performance, Fastjson2 uses a long (64-bit) to store all JSONWriter.Feature values. Currently, the available bits for this long are already very tight (only a dozen left).
> 
> Considering future scalability, we need to reserve these remaining bits for features that are practical, related to functional correctness, security, or core performance. Since PrettyFormatInlineArrays is primarily a formatting beautification feature, which is more of an "addition to excellence", we are hesitant to consume a permanent Feature bit.
> 
> Is it possible to implement this functionality without adding new enumeration values to the JSONWriter.Feature? For instance, through some kind of Context configuration or a separate formatting strategy class?

 Hi @jujn,

  Thank you for the detailed feedback! Your concern about preserving the limited Feature bits makes complete sense.

  I've updated the PR to address this. The latest commit removes PrettyFormatInlineArrays from the Feature enum and instead uses Context configuration, following the same pattern as other non-Feature options like setDateFormat(), setZoneId(), and setMaxLevel().

  The new API:
```java
  // Per-call configuration
  JSONWriter.Context context = new JSONWriter.Context(JSONWriter.Feature.PrettyFormat);
  context.setPrettyFormatInlineArrays(true);
  String json = JSON.toJSONString(obj, context);

  // Or set global default
  JSONFactory.setDefaultWriterPrettyFormatInlineArrays(true);
```
  This approach preserves all existing Feature bits while still providing the functionality. Let me know if you'd like any further changes!

jujn:
Thank you for your quick modification.
Could you please explain in detail the specific scenarios where this format must be used?
Fastjson 2.x did intentionally remove many non core features from 1.x in its design to maintain simplicity. If it is only for the sake of compatibility with the behavior of 1. x, it seems unnecessary?

coffeeandwork:
> Thank you for your quick modification. Could you please explain in detail the specific scenarios where this format must be used? Fastjson 2.x did intentionally remove many non core features from 1.x in its design to maintain simplicity. If it is only for the sake of compatibility with the behavior of 1. x, it seems unnecessary?

Hi @jujn,

  Here are specific scenarios where inline array formatting provides significant value:

  ### 1. Primitive/Simple Value Arrays

  Arrays of IDs, numbers, or simple values become excessively verbose when expanded:

  **Without inline arrays (15 lines):**
  ```json
  {
      "type": 1001,
      "job_id": [
          0,
          45732
      ],
      "user_ids": [
          1297059520,
          4117498193,
          8291829301
      ],
      "is_active": true
  }
```
  With inline arrays (6 lines):
```json
  {
      "type": 1001,
      "job_id": [0,45732],
      "user_ids": [1297059520,4117498193,8291829301],
      "is_active": true
  }
```
  2. GeoJSON and Coordinate Data

  GeoJSON is a widely-used standard for geographic data. Coordinates are conventionally written inline - expanded format makes it nearly impossible to visualize the shape:

  Without inline arrays (25 lines for a simple polygon):
```json
  {
      "type": "Polygon",
      "coordinates": [
          [
              [
                  100.0,
                  0.0
              ],
              [
                  101.0,
                  0.0
              ],
              [
                  101.0,
                  1.0
              ],
              [
                  100.0,
                  0.0
              ]
          ]
      ]
  }
```
  With inline arrays (8 lines):
```json
  {
      "type": "Polygon",
      "coordinates": [
          [[100.0,0.0],[101.0,0.0],[101.0,1.0],[100.0,0.0]]
      ]
  }
```
  3. Matrix and Tabular Data

  Scientific computing, game development, and data processing often use matrices. Inline arrays preserve the visual structure:

  Without inline arrays (38 lines for a 3x3 matrix):
```json
  {
      "name": "rotation_matrix",
      "data": [
          [
              1,
              0,
              0
          ],
          [
              0,
              0.866,
              -0.5
          ],
          [
              0,
              0.5,
              0.866
          ]
      ]
  }
```
  With inline arrays (7 lines):
```json
  {
      "name": "rotation_matrix",
      "data": [
          [1,0,0],
          [0,0.866,-0.5],
          [0,0.5,0.866]
      ]
  }
```

  4. Configuration Files in Version Control

  When config files are stored in git, inline arrays produce cleaner diffs:

  Expanded format diff (adding one port):
```json
    "allowed_ports": [
        8080,
        8443,
  +     9000
    ]
```
  Shows 4 lines changed, harder to review in large configs.

  Inline format diff:
  - "allowed_ports": [8080,8443]
  + "allowed_ports": [8080,8443,9000]
  Single line change, immediately clear what changed.

  5. Logging and Debugging Output

  When logging JSON for debugging, inline arrays make logs scannable:

  Expanded (requires scrolling, loses context):
```json
  2024-01-15 10:23:45 DEBUG Response:
  {
      "status": "success",
      "error_codes": [
          0,
          0,
          0
      ],
      "processed_ids": [
          101,
          102,
          103,
          104,
          105
      ]
  }
  2024-01-15 10:23:46 DEBUG Next request...
```
  Inline (compact, preserves log context):
```json
  2024-01-15 10:23:45 DEBUG Response:
  {
      "status": "success",
      "error_codes": [0,0,0],
      "processed_ids": [101,102,103,104,105]
  }
  2024-01-15 10:23:46 DEBUG Next request...
```
  6. Test Fixtures and Documentation

  Test data and API documentation examples are more maintainable:

  Without inline arrays (test file becomes hundreds of lines):
```json
  {
      "test_input": {
          "rgb_color": [
              255,
              128,
              0
          ],
          "position": [
              10.5,
              20.3,
              5.0
          ],
          "tags": [
              "active",
              "premium",
              "verified"
          ]
      }
  }
```
  With inline arrays (concise, easy to maintain):
```json
  {
      "test_input": {
          "rgb_color": [255,128,0],
          "position": [10.5,20.3,5.0],
          "tags": ["active","premium","verified"]
      }
  }
```
  ---
  The common thread across all these scenarios is that arrays often represent atomic data units (a coordinate pair, a color, a row of values) that are more meaningful when kept together, while objects represent structured entities that benefit from expanded formatting.

  I understand the desire to keep fastjson2 lean. If you feel these use cases aren't common enough to warrant inclusion, I respect that decision.

wenshao:
This will cause problems if the array length is too large.

coffeeandwork:
> This will cause problems if the array length is too large.

Yeah that's a fair point. Though in practice, I think it's less of an issue than it might seem - the use cases where you'd want inline arrays (coordinates, RGB values, matrix rows, small ID lists) tend to be small arrays anyway. Anyone working with huge arrays probably wouldn't enable this option.

  Also worth noting that without pretty printing, arrays already end up on a single line regardless of size - so this isn't introducing new behavior, just giving users a middle ground between fully expanded and fully compact.

  That said, if you think it's important, I could add a max length threshold where arrays beyond N elements fall back to the expanded format. Let me know if you'd want that for the initial implementation or if it's fine as-is.

wenshao:
For primitive type arrays with a length less than 5, it might be better to put them on a single line by default.

coffeeandwork:
> For primitive type arrays with a length less than 5, it might be better to put them on a single line by default.

I can implement this if this is truly what we want. Are you sure we want to change the default behavior in this way? Wouldn't it impact projects that rely on it? Please advise.

coffeeandwork:
Short primitive arrays (< 5 elements) now render inline in pretty-print mode by default. Arrays with 5+ elements use multi-line formatting. The existing `PrettyFormatInlineArrays` option still forces all arrays inline regardless of length.

**Benchmarks** (JDK 22, Apple M4 Max, 1 fork, 3 warmup / 5 measurement iterations):

| Benchmark | Before | After |
|---|---|---|
| EishayWriteString (UTF16) | 5,946,941 ± 55,390 ops/s | 5,887,561 ± 34,412 ops/s |
| EishayWriteUTF8Bytes (UTF8) | 6,739,323 ± 40,674 ops/s | 6,700,563 ± 29,075 ops/s |

Differences are within error margins. No measurable impact on the non-pretty hot path.

core/src/main/java/com/alibaba/fastjson2/JSONWriter.java
```diff
@@ -1793,6 +1805,10 @@ public void writeInt16(short[] value) {
             writeArrayNull();
             return;
         }
+        boolean savedInline = prettyInlineArrays;
+        if (pretty != PRETTY_NON && !prettyInlineArrays && value.length < PRETTY_INLINE_ARRAY_MAX_LENGTH) {
+            prettyInlineArrays = true;
+        }
```
wenshao, line 1811: **[Critical]** The new auto-inline logic sets `prettyInlineArrays = true` for any primitive array with `value.length < 5` whenever PrettyFormat is active, changing short-array output from multi-line to inline. The existing test `JSONWriterPrettyTest.writeInt16()` asserts that `short[]{1, 2}` produces the multi-line form (`"[\n\t1,\n\t2\n]"`), but it now produces `[1,2]`, so the test fails. The diff does not update that test.

Failure scenario: `mvn -pl core test` → `JSONWriterPrettyTest.writeInt16()` fails (expected multi-line, actual `[1,2]`); the PR's CI is currently red across all JDKs/platforms, consistent with this breakage. The same behavior change applies to the other base-class writers modified here (`writeBool(boolean[])`, formatted `writeFloat`/`writeDouble`) — confirm no other existing tests assert multi-line output for short primitive arrays.

Fix: update `JSONWriterPrettyTest.writeInt16()` to expect `[1,2]`, or use an array of >= 5 elements to retain multi-line coverage.

core/src/main/java/com/alibaba/fastjson2/JSONWriterUTF8.java
```diff
@@ -265,6 +269,8 @@ public final void startArray() {
         if (++level > context.maxLevel) {
             overflowLevel();
         }
+        // Mark current level as array context
+        levelArray |= (1L << level);
```
wenshao, line 273: **[Suggestion]** `levelArray` is a 64-bit `long`, but `level` can reach `context.maxLevel` (default 2048). Java masks `long` shifts to `level & 63`, so `1L << level` aliases for `level >= 64` (level 64 collides with level 0). `endArray()` at level 64 runs `levelArray &= ~(1L << 64)` == `~(1L << 0)`, clearing the marker of a still-open level-0 array, so a later `writeComma()` at level 0 mis-indents.

Concrete cost: with inline arrays active (globally via `setDefaultWriterPrettyFormatInlineArrays(true)`, or transiently for short primitive arrays), a structure nested >= 64 levels deep gets corrupted pretty-print indentation at the aliased levels (output stays valid JSON).

Fix: guard the bitmask ops with `if (level < 64)` here and in `endArray()`, and use `boolean inArrayContext = level < 64 && (levelArray & (1L << level)) != 0;` in `writeComma()` (apply in both JSONWriterUTF8 and JSONWriterUTF16).

```suggestion
        // Mark current level as array context
        if (level < 64) {
            levelArray |= (1L << level);
        }
```

core/src/main/java/com/alibaba/fastjson2/JSONWriterUTF16.java
```diff
@@ -198,6 +202,8 @@ public final void startArray() {
         if (++level > context.maxLevel) {
             overflowLevel();
         }
+        // Mark current level as array context
+        levelArray |= (1L << level);
```
wenshao, line 206: **[Suggestion]** `levelArray` is a 64-bit `long`, but `level` can reach `context.maxLevel` (default 2048). Java masks `long` shifts to `level & 63`, so `1L << level` aliases for `level >= 64` (level 64 collides with level 0). `endArray()` at level 64 runs `levelArray &= ~(1L << 64)` == `~(1L << 0)`, clearing the marker of a still-open level-0 array, so a later `writeComma()` at level 0 mis-indents.

Concrete cost: with inline arrays active (globally via `setDefaultWriterPrettyFormatInlineArrays(true)`, or transiently for short primitive arrays), a structure nested >= 64 levels deep gets corrupted pretty-print indentation at the aliased levels (output stays valid JSON).

Fix: guard the bitmask ops with `if (level < 64)` here and in `endArray()`, and use `boolean inArrayContext = level < 64 && (levelArray & (1L << level)) != 0;` in `writeComma()` (apply in both JSONWriterUTF8 and JSONWriterUTF16).

```suggestion
        // Mark current level as array context
        if (level < 64) {
            levelArray |= (1L << level);
        }
```

core/src/main/java/com/alibaba/fastjson2/JSONWriterUTF8.java
```diff
@@ -1779,6 +1811,18 @@ public final void writeInt64(long[] values) {
             return;
         }
 
+        if (pretty != PRETTY_NON && !prettyInlineArrays && values.length >= PRETTY_INLINE_ARRAY_MAX_LENGTH) {
+            startArray();
+            for (int i = 0; i < values.length; i++) {
+                if (i != 0) {
+                    writeComma();
+                }
+                writeInt64(values[i]);
```
wenshao, line 1820: **[Suggestion]** This new multi-line path (length >= 5) delegates per-element to the single-value `writeInt64(long)`, which appends an `L` suffix under `Feature.WriteClassName` for int-range longs; the inline fast-path (length < 5) writes via `IOUtils.writeInt64` with no suffix. So under `PrettyFormat + WriteClassName`, `long[]{1,2,3}` → `[1,2,3]` but `long[]{1,2,3,4,5}` → `[1L,2L,3L,4L,5L]` — round-tripping the two as `Object` yields `int[]` vs `long[]` purely by the 5-element threshold. The same divergence exists for `writeFloat(float[])`/`writeDouble(double[])` NaN handling (inline writes unquoted `null`; the single-value method writes the `"null"` string under `WriteNonStringValueAsString`).

Concrete cost: length-dependent type markers and NaN quoting for the same feature set.

Fix: make both paths agree — replicate the `MASK_WRITE_CLASS_NAME` check in the inline bulk path (or drop the `L` suffix from this delegation loop), and align the NaN quoting (apply in both JSONWriterUTF8 and JSONWriterUTF16).

core/src/main/java/com/alibaba/fastjson2/JSONWriterUTF16.java
```diff
@@ -1888,6 +1920,18 @@ public final void writeInt64(long[] values) {
             return;
         }
 
+        if (pretty != PRETTY_NON && !prettyInlineArrays && values.length >= PRETTY_INLINE_ARRAY_MAX_LENGTH) {
+            startArray();
+            for (int i = 0; i < values.length; i++) {
+                if (i != 0) {
+                    writeComma();
+                }
+                writeInt64(values[i]);
```
wenshao, line 1929: **[Suggestion]** This new multi-line path (length >= 5) delegates per-element to the single-value `writeInt64(long)`, which appends an `L` suffix under `Feature.WriteClassName` for int-range longs; the inline fast-path (length < 5) writes via `IOUtils.writeInt64` with no suffix. So under `PrettyFormat + WriteClassName`, `long[]{1,2,3}` → `[1,2,3]` but `long[]{1,2,3,4,5}` → `[1L,2L,3L,4L,5L]` — round-tripping the two as `Object` yields `int[]` vs `long[]` purely by the 5-element threshold. The same divergence exists for `writeFloat(float[])`/`writeDouble(double[])` NaN handling (inline writes unquoted `null`; the single-value method writes the `"null"` string under `WriteNonStringValueAsString`).

Concrete cost: length-dependent type markers and NaN quoting for the same feature set.

Fix: make both paths agree — replicate the `MASK_WRITE_CLASS_NAME` check in the inline bulk path (or drop the `L` suffix from this delegation loop), and align the NaN quoting (apply in both JSONWriterUTF8 and JSONWriterUTF16).

wenshao (changes requested):

PR #3992, opened 2026-02-16
improved BigDecimal parsing performance

wenshao:
- Use bit-shift operations for faster digit parsing
- Simplify overflow detection to single check after digit loop
- Handle exponent with combined scale to avoid string-based parsing

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -4379,18 +4379,12 @@ final class JSONReaderUTF16
         }
 
         valueType = JSON_TYPE_INT;
-        boolean overflow = false;
+        int digitCount = 0; // Total digit count for overflow detection
         long longValue = 0;
         while (ch >= '0' && ch <= '9') {
             valid = true;
-            if (!overflow) {
-                long r = longValue * 10;
-                if ((longValue | 10) >>> 31 == 0L || (r / 10 == longValue)) {
-                    longValue = r + (ch - '0');
-                } else {
-                    overflow = true;
-                }
-            }
+            longValue = (longValue << 3) + (longValue << 1) + (ch & 0xf);
+            digitCount++;
 
             if (offset == end) {
                 ch = EOI;
@@ -4400,10 +4394,6 @@ final class JSONReaderUTF16
             ch = chars[offset++];
         }
 
-        if (longValue < 0) {
-            overflow = true;
-        }
-
         this.scale = 0;
         if (ch == '.') {
             valueType = JSON_TYPE_DEC;
@@ -4411,14 +4401,8 @@ final class JSONReaderUTF16
             while (ch >= '0' && ch <= '9') {
                 valid = true;
                 this.scale++;
-                if (!overflow) {
-                    long r = longValue * 10;
-                    if ((longValue | 10) >>> 31 == 0L || (r / 10 == longValue)) {
-                        longValue = r + (ch - '0');
-                    } else {
-                        overflow = true;
-                    }
-                }
+                longValue = (longValue << 3) + (longValue << 1) + (ch & 0xf);
+                digitCount++;
 
                 if (offset == end) {
                     ch = EOI;
@@ -4428,6 +4412,9 @@ final class JSONReaderUTF16
                 ch = chars[offset++];
             }
         }
+
+        // Check overflow: more than 18 digits, or 19 digits with negative value (sign bit set)
+        boolean overflow = digitCount > 18 && (digitCount > 19 || longValue < 0);
 
         int expValue = 0;
         if (ch == 'e' || ch == 'E') {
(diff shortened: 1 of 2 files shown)


wenshao (commented):
The same overflow check is duplicated in `core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java` and should be fixed there as well.

_— gpt-5.5 via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
```diff
@@ -6872,6 +6856,9 @@ public final BigDecimal readBigDecimal() {
             }
         }
 
+        // Check overflow: more than 18 digits, or 19 digits with negative value (sign bit set)
```
wenshao, line 6859: **[Critical]** This post-loop overflow check can accept corrupted 19-digit values. For exactly 19 digits, values greater than `Long.MAX_VALUE` may overflow during `longValue = longValue * 10 + digit` and still finish with a non-negative wrapped `longValue`, so they take the `BigDecimal.valueOf(longValue, combinedScale)` fast path instead of falling back to `TypeUtils.parseBigDecimal(...)`. That can silently return the wrong `BigDecimal` for valid JSON numbers outside signed `long` range.

A safer fix is to restore per-digit overflow detection, or add an exact 19-digit range check before using the accumulated `longValue`; otherwise force the existing string-based fallback.

_— gpt-5.5 via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
```diff
@@ -4429,6 +4413,9 @@ public final BigDecimal readBigDecimal() {
             }
         }
 
+        // Check overflow: more than 18 digits, or 19 digits with negative value (sign bit set)
+        boolean overflow = digitCount > 18 && (digitCount > 19 || longValue < 0);
```
wenshao, line 4417: **[Suggestion]** The comment misdescribes the overflow condition — it reads as "more than 18 digits" is sufficient for overflow, but the code only flags overflow unconditionally at 20+ digits; 19 digits overflows only when the sign bit is set. — Concrete cost: a maintainer trusting the comment could simplify to `digitCount > 18`, forcing every valid 19-digit long (e.g. `Long.MAX_VALUE`) onto the slow `TypeUtils.parseBigDecimal` path — a silent performance regression.

```suggestion
        // Check overflow: 20+ digits always overflow; exactly 19 digits overflow only if sign bit is set
        boolean overflow = digitCount > 18 && (digitCount > 19 || longValue < 0);
```

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
```diff
@@ -6872,6 +6856,9 @@ public final BigDecimal readBigDecimal() {
...
+        boolean overflow = digitCount > 18 && (digitCount > 19 || longValue < 0);
```
wenshao, line 6860: **[Suggestion]** Same misleading comment as in `JSONReaderUTF16.java` — it reads as "more than 18 digits" is sufficient for overflow, but the code only flags overflow unconditionally at 20+ digits; 19 digits overflows only when the sign bit is set. — Concrete cost: a maintainer trusting the comment could simplify to `digitCount > 18`, forcing every valid 19-digit long onto the slow path.

```suggestion
        // Check overflow: 20+ digits always overflow; exactly 19 digits overflow only if sign bit is set
        boolean overflow = digitCount > 18 && (digitCount > 19 || longValue < 0);
```

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
```diff
@@ -6959,8 +6946,12 @@ public final BigDecimal readBigDecimal() {
...
         if (!value) {
-            if (expValue == 0 && !overflow && longValue != 0) {
-                decimal = BigDecimal.valueOf(negative ? -longValue : longValue, scale);
+            if (!overflow && longValue != 0) {
+                // Combine scale and exponent: scale - expValue
+                // For 1.5E3: scale=1, exp=3 => combined=-2 => 1500
+                // For 1.5E-3: scale=1, exp=-3 => combined=4 => 0.0015
+                int combinedScale = scale - expValue;
+                decimal = BigDecimal.valueOf(negative ? -longValue : longValue, combinedScale);
                 value = true;
```
wenshao, line 6955: **[Suggestion]** The new exponent fast path in `JSONReaderUTF8.readBigDecimal()` has no test asserting the exact `BigDecimal` result (value and scale) from `byte[]` input with exponent notation. The only UTF8 test with exponents (`DoubleTest.parseDouble1`, input `"123.123E256"`) asserts only `.doubleValue()`, discarding scale. — Concrete cost: a future scale error in the UTF8 path that preserves the double value would go undetected. The UTF16 path is covered by `Issue492` with exact `BigDecimal` assertions.

Consider adding a test that parses exponent-notation numbers from `byte[]` and asserts exact `BigDecimal` equality (e.g. `assertEquals(new BigDecimal("1.1111111E7"), JSON.parseObject(bytes, BigDecimal.class))`), mirroring `Issue492`'s coverage.

PR #4036, opened 2026-03-25
fix: support negative year (BC date) deserialization

### LINKED ISSUE #3845: [BUG]公元前时间反序列化出错
### 问题描述
公元前时间反序列化出错

### 环境信息
*请填写以下信息：*

 - OS信息：  [e.g.：CentOS 8.4.2105 4Core 3.10GHz 16 GB]
 - JDK信息： [e.g.：Openjdk 1.8.0_312]
 - 版本信息：[e.g.：Fastjson2 2.0.60]
 

### 重现步骤
        LocalDateTime time=LocalDateTime.of(0,1,1,0,0,0,0).minusHours(8);
        JSONObject bean=new JSONObject().fluentPut("time",time);
        String s = bean.toJSONString();
        System.out.println(s);
        System.out.println(JSON.parseObject(s).getLocalDateTime("time"));

### 期待的正确结果
不抛出异常

### 相关日志输出
Exception in thread "main" com.alibaba.fastjson2.JSONException: read LocalDateTime error -0001-12-31 16:00:00, offset 22, character , line 1, column 22, fastjson-version 2.0.60 "-0001-12-31 16:00:00"
	at com.alibaba.fastjson2.JSONReader.readLocalDateTime(JSONReader.java:2506)
	at com.alibaba.fastjson2.reader.ObjectReaderImplLocalDateTime.readObject(ObjectReaderImplLocalDateTime.java:64)
	at com.alibaba.fastjson2.util.TypeUtils.cast(TypeUtils.java:1522)
	at com.alibaba.fastjson2.util.TypeUtils.cast(TypeUtils.java:1426)
	at com.alibaba.fastjson2.JSONObject.getLocalDateTime(JSONObject.java:1087)
	at com.alibaba.fastjson2.JSONObject.getLocalDateTime(JSONObject.java:1072)

#### 附加信息

daguimu:
## Problem

`LocalDateTime` with negative years (BC dates) serializes correctly (e.g. `"-0001-12-31 16:00:00"`) but fails to deserialize with:

```
JSONException: read LocalDateTime error -0001-12-31 16:00:00
```

Reproduction:
```java
LocalDateTime time = LocalDateTime.of(0, 1, 1, 0, 0, 0, 0).minusHours(8);
JSONObject bean = new JSONObject().fluentPut("time", time);
String s = bean.toJSONString();  // {"time":"-0001-12-31 16:00:00"}
JSON.parseObject(s).getLocalDateTime("time");  // throws JSONException
```

## Root Cause

The negative year date string `"-0001-12-31 16:00:00"` has length 20 (due to the leading `-`). The `readLocalDateTime()` method dispatches length-20 strings to `parseLocalDateTime20()`, which expects a completely different format (`dd Mon yyyy HH:mm:ss`). This returns null, and the string eventually falls through to an error.

## Fix

Added handling for the `-yyyy-MM-dd HH:mm:ss` format in the string fallback path of `readLocalDateTime()`. The fix:
- Checks if the string is 20 chars starting with `-`
- Validates the format matches `-yyyy-MM-dd HH:mm:ss` (supporting both ` ` and `T` separators, and `/` delimiters)
- Manually parses the year, month, day, hour, minute, second components
- Follows the same manual parsing style used throughout the codebase

## Tests Added

- `testNegativeYearRoundTrip` — Serialize then deserialize a BC date, verify equality
- `testNegativeYearDirectParse` — Parse `"-0001-12-31 16:00:00"` and verify all components
- `testNegativeYearWithT` — Parse with `T` separator: `"-0001-12-31T16:00:00"`
- `testNegativeYearWithSlash` — Parse with `/` delimiter: `"-0001/12/31 16:00:00"`
- `testNegativeYearLarger` — Year -100 (101 BC) round-trip
- `testPositiveYearStillWorks` — Regression test for normal dates

## Impact

Minimal — only affects deserialization of negative year datetime strings (20 chars starting with `-`). Normal date parsing is unaffected.

Fixes #3845

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -2522,6 +2522,32 @@ public abstract class JSONReader
             wasNull = true;
             return null;
         }
+
+        // Handle negative year dates: -yyyy-MM-dd HH:mm:ss or -yyyy-MM-ddTHH:mm:ss
+        if (str.length() == 20 && str.charAt(0) == '-') {
+            char c5 = str.charAt(5);
+            char c8 = str.charAt(8);
+            char c11 = str.charAt(11);
+            char c14 = str.charAt(14);
+            char c17 = str.charAt(17);
+            if ((c5 == '-' || c5 == '/') && (c8 == '-' || c8 == '/')
+                    && (c11 == ' ' || c11 == 'T') && c14 == ':' && c17 == ':') {
+                int year = -(((str.charAt(1) - '0') * 1000)
+                        + ((str.charAt(2) - '0') * 100)
+                        + ((str.charAt(3) - '0') * 10)
+                        + (str.charAt(4) - '0'));
+                int month = ((str.charAt(6) - '0') * 10) + (str.charAt(7) - '0');
+                int dom = ((str.charAt(9) - '0') * 10) + (str.charAt(10) - '0');
+                int hour = ((str.charAt(12) - '0') * 10) + (str.charAt(13) - '0');
+                int minute = ((str.charAt(15) - '0') * 10) + (str.charAt(16) - '0');
+                int second = ((str.charAt(18) - '0') * 10) + (str.charAt(19) - '0');
+                if (month >= 1 && month <= 12 && dom >= 1 && dom <= 31
+                        && hour <= 23 && minute <= 59 && second <= 59) {
+                    return LocalDateTime.of(year, month, dom, hour, minute, second);
+                }
+            }
+        }
+
         throw new JSONException(info("read LocalDateTime error " + str));
     }
 
--- a/core/src/test/java/com/alibaba/fastjson2/issues_3800/Issue3845.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_3800/Issue3845.java
@@ -1,0 +1,82 @@
+package com.alibaba.fastjson2.issues_3800;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDateTime;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/**
+ * #3845 BC date (negative year) deserialization error
+ */
+public class Issue3845 {
+    @Test
+    public void testNegativeYearRoundTrip() {
+        // Year 0 minus 8 hours = year -1, Dec 31, 16:00:00
+        LocalDateTime time = LocalDateTime.of(0, 1, 1, 0, 0, 0, 0).minusHours(8);
+        JSONObject bean = new JSONObject().fluentPut("time", time);
+        String s = bean.toJSONString();
+        LocalDateTime parsed = JSON.parseObject(s).getLocalDateTime("time");
+        assertNotNull(parsed);
+        assertEquals(time, parsed);
+    }
+
+    @Test
+    public void testNegativeYearDirectParse() {
+        // Parse a negative year datetime string directly
+        String json = "{\"time\":\"-0001-12-31 16:00:00\"}";
+        LocalDateTime parsed = JSON.parseObject(json).getLocalDateTime("time");
+        assertNotNull(parsed);
+        assertEquals(-1, parsed.getYear());
+        assertEquals(12, parsed.getMonthValue());
+        assertEquals(31, parsed.getDayOfMonth());
+        assertEquals(16, parsed.getHour());
+        assertEquals(0, parsed.getMinute());
+        assertEquals(0, parsed.getSecond());
+    }
+
+    @Test
+    public void testNegativeYearWithT() {
+        // Parse with T separator instead of space
+        String json = "{\"time\":\"-0001-12-31T16:00:00\"}";
+        LocalDateTime parsed = JSON.parseObject(json).getLocalDateTime("time");
+        assertNotNull(parsed);
+        assertEquals(-1, parsed.getYear());
+        assertEquals(12, parsed.getMonthValue());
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -2522,6 +2522,32 @@ public LocalDateTime readLocalDateTime() {
...
+        // Handle negative year dates: -yyyy-MM-dd HH:mm:ss or -yyyy-MM-ddTHH:mm:ss
+        if (str.length() == 20 && str.charAt(0) == '-') {
+            char c5 = str.charAt(5);
+            char c8 = str.charAt(8);
+            char c11 = str.charAt(11);
+            char c14 = str.charAt(14);
+            char c17 = str.charAt(17);
+            if ((c5 == '-' || c5 == '/') && (c8 == '-' || c8 == '/')
+                    && (c11 == ' ' || c11 == 'T') && c14 == ':' && c17 == ':') {
+                int year = -(((str.charAt(1) - '0') * 1000)
```
wenshao, line 2535: **[Critical]** The new negative-year parser computes numeric fields with `str.charAt(i) - '0'` but does not validate that those positions are actually digits, and the time fields only have upper-bound checks. Inputs with valid separators but non-digit numeric characters can be coerced into different valid values instead of being rejected; for example, `-0001-12-3/ 16:00:00` can parse the day as `29` because `'/' - '0' == -1`. Please validate every numeric position before computing the fields, or use existing digit helpers that reject non-digits, and require lower bounds for hour/minute/second as well.

_— gpt-5.5 via Qwen Code /review_

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -2522,6 +2522,32 @@ public LocalDateTime readLocalDateTime() {
...
+                        + ((str.charAt(3) - '0') * 10)
+                        + (str.charAt(4) - '0'));
+                int month = ((str.charAt(6) - '0') * 10) + (str.charAt(7) - '0');
+                int dom = ((str.charAt(9) - '0') * 10) + (str.charAt(10) - '0');
+                int hour = ((str.charAt(12) - '0') * 10) + (str.charAt(13) - '0');
+                int minute = ((str.charAt(15) - '0') * 10) + (str.charAt(16) - '0');
+                int second = ((str.charAt(18) - '0') * 10) + (str.charAt(19) - '0');
+                if (month >= 1 && month <= 12 && dom >= 1 && dom <= 31
+                        && hour <= 23 && minute <= 59 && second <= 59) {
+                    return LocalDateTime.of(year, month, dom, hour, minute, second);
```
wenshao, line 2546: **[Suggestion]** `dom <= 31` does not account for per-month day limits, so a negative-year date like Feb 31 passes validation and `LocalDateTime.of()` throws `java.time.DateTimeException` instead of this method's contractual `JSONException` — Concrete cost: `"-0001-02-31 16:00:00"` throws an unchecked `DateTimeException` that callers catching `JSONException` won't handle; before this change the same input fell through to the `JSONException` below. (The positive-year path shares this gap, but this diff makes it newly reachable for negative years.)

```suggestion
                if (month >= 1 && month <= 12 && dom >= 1 && dom <= 31
                        && hour <= 23 && minute <= 59 && second <= 59) {
                    try {
                        return LocalDateTime.of(year, month, dom, hour, minute, second);
                    } catch (java.time.DateTimeException ignored) {
                        // fall through to JSONException below
                    }
```

core/src/test/java/com/alibaba/fastjson2/issues_3800/Issue3845.java
```diff
@@ -0,0 +1,82 @@
...
+        LocalDateTime time = LocalDateTime.of(-100, 6, 15, 12, 30, 45);
+        JSONObject bean = new JSONObject().fluentPut("time", time);
+        String s = bean.toJSONString();
+        LocalDateTime parsed = JSON.parseObject(s).getLocalDateTime("time");
+        assertNotNull(parsed);
+        assertEquals(time, parsed);
+    }
+
+    @Test
+    public void testPositiveYearStillWorks() {
```
wenshao, line 73: **[Suggestion]** No test exercises the error path — out-of-range components or non-digit characters that should fall through to `JSONException` — Concrete cost: if a range check (e.g. `month <= 12`) were accidentally deleted, `"-0001-13-31 16:00:00"` would reach `LocalDateTime.of` and throw `DateTimeException`, and no test would fail. Adding negative tests would guard the contract (and pin the digit-validation fix from the Critical above). Illustrative (needs `assertThrows` / `JSONException` imports):

```java
@Test
public void testNegativeYearInvalid() {
    assertThrows(JSONException.class, () ->
        JSON.parseObject("{\"time\":\"-0001-13-31 16:00:00\"}").getLocalDateTime("time"));
    assertThrows(JSONException.class, () ->
        JSON.parseObject("{\"time\":\"-000a-01-01 00:00:00\"}").getLocalDateTime("time"));
}
```

wenshao (changes requested):
**[Critical]** core/src/main/java/com/alibaba/fastjson2/JSONReader.java (~lines 2536-2544): The negative-year parser computes year/month/day/hour/minute/second via raw `str.charAt(i) - '0'` arithmetic without validating that those positions are ASCII digits, and hour/minute/second have no lower-bound checks. Input with valid separators but non-digit numeric characters is coerced into a wrong-but-valid LocalDateTime instead of being rejected — e.g. `"-0001-12-3/ 16:00:00"` parses the day as 29 because `'/' - '0' == -1`, and `"-000a-01-01 00:00:00"` yields year -49. The sibling positive-year parser uses IOUtils.digit2/digit4 (which return -1 for non-digits) and rejects such input; this new path bypasses that pattern. This is the same defect already reported in the existing inline thread (comment 3141714528); re-checked against the code at this commit and it still stands — not re-posted inline to avoid a duplicate.

PR #7621, opened 2026-04-16
Fix Kotlin 2.3.20 compatibility in parameter annotation handling

utafrali:
What this PR does / why we need it?

Kotlin 2.3.20 broke serialization for data classes. The issue is that KotlinUtils.getKoltinConstructorParameters() can return null when reflection fails, but ObjectWriterBaseModule wasn't checking for it before iterating. Results in an NPE.

Summary of your change

Added a null check on paramNames before the loop. Also cleaned up that debug printStackTrace that was hiding the real error. Added a test for when Kotlin reflection fails.

Checklist

- [x] Tests passing
- [x] Commit format follows conventional commits
- [x] No documentation changes needed


utafrali:
Fair enough, closing this.

PR #7631, opened 2026-04-25
fix: honor IgnoreSetNullValue in specialized FieldReader subclasses, for issue #7618

### LINKED ISSUE #7618: IgnoreSetNullValue ignored by specialized FieldReader subclasses (FieldReaderString et al.)
## Summary

`JSONReader.Feature.IgnoreSetNullValue` is documented to skip setter calls when the incoming JSON value is `null`. It works for fields deserialized via `FieldReaderObject`, but is silently ignored for fields deserialized via the specialized subclasses — most commonly **`FieldReaderString`**, which is picked for every `String` field on a bean.

This means neither of the documented ways to enable the feature actually works for `String` fields:

1. `JSON.parseObject(json, Clazz.class, JSONReader.Feature.IgnoreSetNullValue)` — parse-time context feature.
2. `@JSONType(deserializeFeatures = JSONReader.Feature.IgnoreSetNullValue)` on the bean — annotation-level feature that correctly propagates into `FieldReader.features`.

I verified both against the current master (`2.0.62-SNAPSHOT`, commit `3697c2d37`) and against `2.0.53`.

## Reproducer

```java
public class Repro {
    @JSONType(deserializeFeatures = JSONReader.Feature.IgnoreSetNullValue)
    public static class Sample {
        private String name = "Default";
        public String getName() { return name; }
        public void setName(String n) { this.name = n; }
    }

        Sample s = JSON.parseObject("{\"name\":null}", Sample.class);
        System.out.println(s.getName()); // expected "Default", actual null
    }
}
```

Also fails with the context-feature form:

```java
JSON.parseObject("{\"name\":null}", Sample.class, JSONReader.Feature.IgnoreSetNullValue);
// expected name == "Default", actual null
```

## Root cause

Debug-inspecting the generated reader confirms the feature **does** flow correctly into the FieldReader when `@JSONType(deserializeFeatures=...)` is used:

```
Reader class: com.alibaba.fastjson2.reader.ORG_1_1_Sample
FieldReader class: com.alibaba.fastjson2.reader.FieldReaderString
IgnoreSetNullValue in FieldReader.features? true   <-- correctly propagated
Parsed name = null                                  <-- but ignored at parse time
```

The propagation chain works:
- `ObjectReaderBaseModule.getBeanInfo` reads `@JSONType.deserializeFeatures` and ORs them into `beanInfo.readerFeatures`.
- `ObjectReaderCreator` ORs `beanInfo.readerFeatures` into `fieldInfo.features`.
- The resulting `FieldReaderString.features` contains the `IgnoreSetNullValue` mask.

The feature is then **ignored at the FieldReader subclass level**. `FieldReaderObject.accept(T, Object)` correctly gates on it:

```java
// FieldReaderObject.java:300
if (isParameter() || (value == null && (features & JSONReader.Feature.IgnoreSetNullValue.mask) != 0)) {
    return;
}
```

But `FieldReaderString.accept(T, Object)` and `FieldReaderString.readFieldValue(JSONReader, T)` do not check the feature — they call `propertyAccessor.setObject(object, fieldValue)` regardless of `fieldValue == null`:

```java
// FieldReaderString.java:52-83  — no IgnoreSetNullValue guard
@Override
public void accept(T object, Object value) {
    String fieldValue = ...;
    ...
    propertyAccessor.setObject(object, fieldValue);
}

// FieldReaderString.java:87-     — no IgnoreSetNullValue guard
@Override
public void readFieldValue(JSONReader jsonReader, T object) {
    String fieldValue = jsonReader.readString();
    ...
    propertyAccessor.setObject(object, fieldValue);
}
```

A `grep -n IgnoreSetNullValue core/src/main/java/.../reader/` shows the feature is honored in only a handful of readers: `FieldReaderObject`, `FieldReaderZonedDateTime`, `FieldReaderLocalDateTime`, `FieldReaderInstant`, `FieldReaderStackTrace`, plus `ObjectReaderAdapter#createInstance(Map, features)`. All other specialized FieldReaders silently ignore it, including `FieldReaderString` and the primitive-boxed variants.

## Suggested fix

Add the same null-skip gate used in `FieldReaderObject` to every specialized FieldReader subclass's `accept(T, Object)` and `readFieldValue(JSONReader, T)` paths. Minimal per-class diff, e.g. for `FieldReaderString`:

```java
if (fieldValue == null && (features & JSONReader.Feature.IgnoreSetNullValue.mask) != 0) {
    return;
}
```

## Environment

- fastjson2: 2.0.53 and master (`2.0.62-SNAPSHOT`, commit `3697c2d37`)
- JDK: 21
- OS: macOS

## Context

Encountered while migrating a Spring `FastJsonHttpMessageConverter`-based service from Jackson (`@JsonSetter(nulls = Nulls.SKIP)`) to fastjson2. No combination of `FastJsonConfig.setReaderFeatures`, module-level `BeanInfo.readerFeatures` injection, or `@JSONType(deserializeFeatures=...)` restored `Nulls.SKIP` parity for `String` fields. Tracing confirmed the feature reaches the FieldReader correctly — it is simply not consulted by `FieldReaderString`.

wenshao:
## Summary
- The null-skip gate present in `FieldReaderObject` was missing from type-specialized `FieldReader` subclasses (`FieldReaderString`, boxed primitives, `BigDecimal`/`BigInteger`, `Number`, `Date`), so `JSONReader.Feature.IgnoreSetNullValue` was silently ignored for those fields — both when configured via `@JSONType(deserializeFeatures = ...)` and when passed at parse time.
- Adds `ignoreSetNullValue()` helpers on `FieldReader` so the gate is reused consistently, and applies it in `accept(T, Object)` and `readFieldValue(JSONReader, T)` for the affected specialized readers.
- `ObjectReaderCreatorASM.genReadFieldValue`:
  - extends the existing `NullOnError` fallback so it also falls back to `fieldReader.readFieldValue` when `IgnoreSetNullValue` is in `fieldFeatures` (annotation-driven case),
  - emits a runtime check on `FEATURES` for non-primitive fields on default-constructor beans, so the parse-time context feature also takes effect on the ASM hot path.
- The `NoneDefaultConstructor` ASM path uses temp locals (no OBJECT slot yet), so the runtime check is skipped there — that path needs a separate fix.

Fixes #7618.

## Test plan
- [x] New `Issue7618` test covering all four combinations: String / boxed primitives × annotation-level / context-level feature.
- [x] Existing related tests (`Issue3789`, `Issue1745`, `Issue2164`) still pass.
- [x] All 2189 `Issue*` regression tests pass.
- [x] All `FieldReader*` and `ObjectReader*Test` tests pass.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
@@ -69,6 +69,9 @@ final class FieldReaderDate<T>
 
     @Override
     protected void accept(T object, Date value) {
+        if (value == null && ignoreSetNullValue()) {
+            return;
+        }
         propertyAccessor.setObject(object, value);
     }
 
--- a/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
@@ -64,6 +64,9 @@ final class FieldReaderDate<T>
             }
             date = null;
         }
+        if (date == null && ignoreSetNullValue(jsonReader)) {
+            return;
+        }
         accept(object, date);
     }
 
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
@@ -3014,7 +3014,7 @@ public class ObjectReaderCreatorASM
 
         MethodWriter mw = mwc.mw;
 
-        if ((fieldFeatures & JSONReader.Feature.NullOnError.mask) != 0) {
+        if ((fieldFeatures & (JSONReader.Feature.NullOnError.mask | JSONReader.Feature.IgnoreSetNullValue.mask)) != 0) {
             mw.aload(THIS);
             mw.getfield(classNameType, fieldReader(fieldReaderIndex), DESC_FIELD_READER);
             mw.aload(JSON_READER);
@@ -3027,6 +3027,31 @@ public class ObjectReaderCreatorASM
         Method method = fieldReader.method;
 
         Label endSet_ = new Label();
+
+        // Honor JSONReader.Feature.IgnoreSetNullValue passed at parse time (context-level).
+        // Annotation-level IgnoreSetNullValue is handled by the fieldFeatures branch above.
+        // NoneDefaultConstructor uses temp locals (no OBJECT slot yet), so skip — that
+        // path needs a different fix.
+        if (!fieldClass.isPrimitive()
+                && !(context.objectReaderAdapter instanceof ObjectReaderNoneDefaultConstructor)) {
+            // if ((FEATURES & IgnoreSetNullValue.mask) != 0) { fieldReader.readFieldValue(jsonReader, object); goto endSet_; }
+            Label noFallback_ = new Label();
+            mw.lload(FEATURES);
+            mw.visitLdcInsn(JSONReader.Feature.IgnoreSetNullValue.mask);
+            mw.land();
+            mw.lconst_0();
+            mw.lcmp();
+            mw.ifeq(noFallback_);
+
+            mw.aload(THIS);
+            mw.getfield(classNameType, fieldReader(fieldReaderIndex), DESC_FIELD_READER);
+            mw.aload(JSON_READER);
+            mw.aload(OBJECT);
+            mw.invokevirtual(TYPE_FIELD_READE, "readFieldValue", METHOD_DESC_READ_FIELD_VALUE);
+            mw.goto_(endSet_);
+
+            mw.visitLabel(noFallback_);
+        }
 
         String TYPE_FIELD_CLASS = ASMUtils.type(fieldClass);
         String DESC_FIELD_CLASS = ASMUtils.desc(fieldClass);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
@@ -3014,7 +3014,9 @@ public class ObjectReaderCreatorASM
 
         MethodWriter mw = mwc.mw;
 
-        if ((fieldFeatures & (JSONReader.Feature.NullOnError.mask | JSONReader.Feature.IgnoreSetNullValue.mask)) != 0) {
+        if ((fieldFeatures & JSONReader.Feature.NullOnError.mask) != 0
+                || ((fieldFeatures & JSONReader.Feature.IgnoreSetNullValue.mask) != 0
+                    && !(context.objectReaderAdapter instanceof ObjectReaderNoneDefaultConstructor))) {
             mw.aload(THIS);
             mw.getfield(classNameType, fieldReader(fieldReaderIndex), DESC_FIELD_READER);
             mw.aload(JSON_READER);
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7600/Issue7618.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7600/Issue7618.java
@@ -1,0 +1,146 @@
+package com.alibaba.fastjson2.issues_7600;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONB;
+import com.alibaba.fastjson2.JSONReader;
+import com.alibaba.fastjson2.annotation.JSONType;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Date;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Issue 7618: JSONReader.Feature.IgnoreSetNullValue is silently ignored by
+ * specialized FieldReader subclasses (FieldReaderString, FieldReaderInt32 etc.)
+ * because the null-skip gate present in FieldReaderObject was not duplicated
+ * into them. The feature must work both via @JSONType.deserializeFeatures and
+ * via JSONReader.Feature passed at parse time.
+ */
+public class Issue7618 {
+    @JSONType(deserializeFeatures = JSONReader.Feature.IgnoreSetNullValue)
+    public static class StringSample {
+        private String name = "Default";
+
+        public String getName() {
+            return name;
+        }
+
+        public void setName(String n) {
+            this.name = n;
+        }
+    }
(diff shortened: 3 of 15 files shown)


core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
```diff
@@ -69,6 +69,9 @@ public void readFieldValue(JSONReader jsonReader, T object) {
 
     @Override
     protected void accept(T object, Date value) {
+        if (value == null && ignoreSetNullValue()) {
+            return;
+        }
```
wenshao, line 77: **[Critical]** `FieldReaderDate.readFieldValue(JSONReader, T)` delegates null-skip to `accept(T, Date)`, which calls `ignoreSetNullValue()` (field-level features only). Every other modified subclass checks `ignoreSetNullValue(jsonReader)` (context-aware, ORs in `jsonReader.getContext().getFeatures()`) directly in `readFieldValue`. Context-level `IgnoreSetNullValue` (passed at parse time) is silently ignored for `Date` fields. — Failure scenario: `JSON.parseObject("{\"d\":null}", Bean.class, JSONReader.Feature.IgnoreSetNullValue)` with `Date d = new Date(0L)` → `d` becomes `null` instead of retaining the default, because `accept` checks only `this.features` (zero for a plain bean). The ASM context-level fallback delegates to this same broken path. The test `testBoxedContextFeature` omits `Date`, so this ships undetected.

Fix: add the context-aware guard in `readFieldValue` before the `accept` call, matching every other subclass:

```java
if (date == null && ignoreSetNullValue(jsonReader)) {
    return;
}
accept(object, date);
```

core/src/test/java/com/alibaba/fastjson2/issues_7600/Issue7618.java
```diff
@@ -0,0 +1,146 @@
...
+        assertEquals(new BigDecimal("3.14"), s.getBd());
+        assertEquals(BigInteger.TEN, s.getBi());
+        assertEquals(new Date(0L), s.getD());
+    }
+
+    public static class BoxedSamplePlain {
+        private Integer i = 42;
+        private Long l = 100L;
+        private Boolean b = Boolean.TRUE;
+        private String s = "kept";
```
wenshao, line 134: **[Suggestion]** Five modified `FieldReader` subclasses — `Float`, `Double`, `Short` (`Int16`), `Byte` (`Int8`), and `Number` — have no test coverage in either the annotation-level or context-level test paths. — Concrete cost: a future null-guard misplacement in any of these readers (the exact class of error in the `FieldReaderDate` bug above) would pass CI undetected.

```java
// Add to BoxedSamplePlain (and BoxedSample):
private Float f = 1.5f;
private Double d = 2.5d;
private Short sh = 7;
private Byte by = 3;
private Number n = 99;
// ... with getters/setters and null-input assertions
```

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
```diff
@@ -3014,7 +3014,7 @@ private <T> void genReadFieldValue(
 
         MethodWriter mw = mwc.mw;
 
-        if ((fieldFeatures & JSONReader.Feature.NullOnError.mask) != 0) {
+        if ((fieldFeatures & (JSONReader.Feature.NullOnError.mask | JSONReader.Feature.IgnoreSetNullValue.mask)) != 0) {
```
wenshao: **[Critical]** The field-level early-return branch was extended to include `IgnoreSetNullValue.mask` but lacks the `ObjectReaderNoneDefaultConstructor` guard that the new context-level block (line ~3035) explicitly adds. For `NoneDefaultConstructor` readers, the `OBJECT` local is never assigned (only the `else` branch calls `genCreateObject` + `astore(OBJECT)`), so the generated bytecode calls `readFieldValue(jsonReader, null)` → `propertyAccessor.setObject(null, value)` → `NullPointerException`. — Failure scenario: a class without a default constructor (e.g. `@JSONCreator`) annotated with `@JSONType(deserializeFeatures = IgnoreSetNullValue)` → the field-level condition fires at codegen time → `aload(OBJECT)` loads null → NPE crashes deserialization. Before this diff, `IgnoreSetNullValue` on a `NoneDefaultConstructor` class was silently ignored (no crash); the diff makes it crash.

```suggestion
        if ((fieldFeatures & JSONReader.Feature.NullOnError.mask) != 0
                || ((fieldFeatures & JSONReader.Feature.IgnoreSetNullValue.mask) != 0
                && !(context.objectReaderAdapter instanceof ObjectReaderNoneDefaultConstructor))) {
```

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running. Reviewed diff-only — the PR’s existing discussion could not be fetched, so this is not an approval and not a no-blockers claim. Suggestions are inline. Not reviewed: issue-fidelity — pr-context could not be fetched (network failure). Not reviewed: the linked-issue fidelity pass — its prompt was built, but no agent on record was launched with it.

core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderDate.java
wenshao: Fixed in be9ea1475. `FieldReaderDate.readFieldValue` now checks `ignoreSetNullValue(jsonReader)` (context-aware) before calling `accept`, consistent with all other modified subclasses. The field-level check in `accept(T, Date)` is retained for callers that go through `accept` directly (e.g. `acceptNull`).

core/src/test/java/com/alibaba/fastjson2/issues_7600/Issue7618.java
wenshao: Done in be9ea1475. Added Float, Double, Short, Byte to both annotation-level (`BoxedSample`) and context-level (`BoxedSamplePlain`) tests. Number is covered at annotation-level only — context-level for `Number` fields goes through a different ASM codegen route (inlined generic-object path, not `genReadFieldValue`), which is a pre-existing gap out of scope for this PR. Also added Date to the context-level test to gate the FieldReaderDate fix above.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
wenshao: Fixed in be9ea1475. The field-level branch now separates the conditions: NullOnError retains its original behavior (pre-existing), while IgnoreSetNullValue is guarded with `!(context.objectReaderAdapter instanceof ObjectReaderNoneDefaultConstructor)`, matching the context-level block guard. For NoneDefaultConstructor readers with annotation-level IgnoreSetNullValue, the code falls through to the normal ASM path instead of the early-return that would load an unassigned OBJECT local.

PR #7632, opened 2026-04-25
fix: ISO8601 should take priority over yyyyMMddHHmmss in Instant write, for issue #4003

### LINKED ISSUE #4003: [BUG] 时间序列化的问题，在不同小版本之间序列化结果不一致！
### 问题描述
对于java.time.Instant对象（其他时间类没有测试过），在配置完全相同的情况下，fastjson2在v2.0.40和v2.0.61版本中序列化结果不一致。

配置：
1. json配置：config.setSerializerFeatures(SerializerFeature.UseISO8601DateFormat)
2. spring启动配置：TimeZone.setDefault(TimeZone.getTimeZone("UTC"));

v2.0.40中，Instant对象被序列化成："2026-02-24T14:50:00Z"（正确）
v2.0.61中，Instant对象被序列化成："2026-02-24 14:50:00" （错误）

### 环境信息
*请填写以下信息：*

 - OS信息：  Ubuntu 22; Windows 11
 - JDK信息：JDK 11
 - 版本信息：fastjson2 2.0.61
 

### 重现步骤
*如何操作可以重现该问题：*

见上文

### 期待的正确结果
*对您期望发生的结果进行清晰简洁的描述。*
对于已经设置UTC时区的情况，java.time.Instant应该被序列化为YYYY-MM-DDThh:mm:ss<TZDSuffix>的格式。

wenshao:
## Summary
- When the `JSONWriter.Context` has `dateFormat="iso8601"` (the runtime form that fastjson1's `SerializerFeature.UseISO8601DateFormat` resolves into) ISO8601 must take priority over any string-pattern shape flags such as `yyyyMMddhhmmss19`/`14`/`8`/`10`.
- Previously the codec evaluated those flags purely against the codec instance + context's pattern flags, so a setup that had been initialized with `dateFormat="yyyy-MM-dd HH:mm:ss"` and later switched to ISO8601 could still hit the `yyyyMMddhhmmss19` path and produce `"2026-02-24 14:50:00"` instead of `"2026-02-24T14:50:00Z"`.
- Compute a single `iso8601` effective flag and gate the `yyyyMMddhhmmss` shape paths on `!iso8601`, so an explicit ISO8601 request always wins.

Fixes #4003.

## Test plan
- [x] New `Issue4003` test covering: ISO8601 alone, ISO8601 after a default `yyyy-MM-dd HH:mm:ss`, default `yyyy-MM-dd HH:mm:ss` without ISO8601 (regression), and no-format default.
- [x] All `*Instant*`, `*ISO8601*`, `*DateTimeCodec*`, `Issue*` tests pass (37 tests).

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
@@ -36,13 +36,24 @@ final class ObjectWriterImplInstant
                 : context.getDateFormat();
 
         Instant instant = (Instant) object;
-        if (dateFormat == null) {
+
+        // Honor formatISO8601 / context.isDateFormatISO8601() ahead of any
+        // string-pattern shape flags (yyyyMMddHHmmss19 etc.) — when the caller
+        // explicitly requested ISO8601 (e.g. SerializerFeature.UseISO8601DateFormat
+        // or JSONWriter.Feature.WriteISO8601DateFormat) it must take priority over
+        // a default dateFormat that may have been set on the context (issue #4003).
+        boolean iso8601 = formatISO8601 || (this.format == null && context.isDateFormatISO8601());
+
+        if (dateFormat == null && !iso8601) {
             jsonWriter.writeInstant(instant);
             return;
         }
 
-        boolean yyyyMMddhhmmss19 = this.yyyyMMddhhmmss19 || (context.isFormatyyyyMMddhhmmss19() && this.format == null);
-        if (yyyyMMddhhmmss14 || yyyyMMddhhmmss19 || yyyyMMdd8 || yyyyMMdd10) {
+        boolean yyyyMMddhhmmss19 = !iso8601 && (this.yyyyMMddhhmmss19 || (context.isFormatyyyyMMddhhmmss19() && this.format == null));
+        boolean yyyyMMddhhmmss14local = !iso8601 && yyyyMMddhhmmss14;
+        boolean yyyyMMdd8local = !iso8601 && yyyyMMdd8;
+        boolean yyyyMMdd10local = !iso8601 && yyyyMMdd10;
+        if (yyyyMMddhhmmss14local || yyyyMMddhhmmss19 || yyyyMMdd8local || yyyyMMdd10local) {
             final long SECONDS_PER_DAY = 60 * 60 * 24;
             ZoneId zoneId = context.getZoneId();
             long epochSecond = instant.getEpochSecond();
@@ -173,7 +184,7 @@ final class ObjectWriterImplInstant
 
         int year = zdt.getYear();
         if (year >= 0 && year <= 9999) {
-            if (formatISO8601 || (format == null && context.isDateFormatISO8601())) {
+            if (iso8601) {
                 jsonWriter.writeDateTimeISO8601(
                         year,
                         zdt.getMonthValue(),
--- a/core/src/test/java/com/alibaba/fastjson2/issues_4000/Issue4003.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_4000/Issue4003.java
@@ -1,0 +1,75 @@
+package com.alibaba.fastjson2.issues_4000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONWriter;
+import org.junit.jupiter.api.Test;
+
+import java.time.Instant;
+import java.time.ZoneId;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Issue 4003: when WriteISO8601DateFormat is requested at runtime but the
+ * JSONWriter context already carries a default dateFormat such as
+ * "yyyy-MM-dd HH:mm:ss" (for example via Spring Boot's FastJsonConfig),
+ * Instant was being serialized as "yyyy-MM-dd HH:mm:ss" — the explicit
+ * ISO8601 feature should take priority.
+ */
+public class Issue4003 {
+    public static class Bean {
+        public Instant time;
+    }
+
+    private static JSONWriter.Context utcContext() {
+        JSONWriter.Context ctx = new JSONWriter.Context();
+        ctx.setZoneId(ZoneId.of("UTC"));
+        return ctx;
+    }
+
+    private static JSONWriter.Context utcContext(String dateFormat) {
+        JSONWriter.Context ctx = new JSONWriter.Context(dateFormat);
+        ctx.setZoneId(ZoneId.of("UTC"));
+        return ctx;
+    }
+
+    @Test
+    public void testIso8601ContextDateFormat() {
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        JSONWriter.Context ctx = utcContext();
+        ctx.setDateFormat("iso8601");
+        String json = JSON.toJSONString(b, ctx);
+        assertEquals("{\"time\":\"2026-02-24T14:50:00Z\"}", json);
+    }
+
+    @Test
+    public void testIso8601OverridesDefaultDateFormat() {
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        // Simulates Spring Boot's FastJsonConfig: default dateFormat is
+        // "yyyy-MM-dd HH:mm:ss", then the user later switches to ISO8601.
+        JSONWriter.Context ctx = utcContext("yyyy-MM-dd HH:mm:ss");
+        ctx.setDateFormat("iso8601");
+        String json = JSON.toJSONString(b, ctx);
+        assertEquals("{\"time\":\"2026-02-24T14:50:00Z\"}", json);
+    }
+
+    @Test
+    public void testDefaultDateFormatStillUsedWithoutIso8601() {
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        JSONWriter.Context ctx = utcContext("yyyy-MM-dd HH:mm:ss");
+        String json = JSON.toJSONString(b, ctx);
+        assertEquals("{\"time\":\"2026-02-24 14:50:00\"}", json);
+    }
+
+    @Test
+    public void testNoFeatureNoFormat() {
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        JSONWriter.Context ctx = utcContext();
+        String json = JSON.toJSONString(b, ctx);
+        assertEquals("{\"time\":\"2026-02-24T14:50:00Z\"}", json);
+    }
+}


core/src/test/java/com/alibaba/fastjson2/issues_4000/Issue4003.java
```diff
@@ -0,0 +1,75 @@
...
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        JSONWriter.Context ctx = utcContext();
+        ctx.setDateFormat("iso8601");
+        String json = JSON.toJSONString(b, ctx);
+        assertEquals("{\"time\":\"2026-02-24T14:50:00Z\"}", json);
+    }
+
+    @Test
+    public void testIso8601OverridesDefaultDateFormat() {
```
wenshao, line 47: **[Suggestion]** None of the four tests in this class fail without the production change — I ran `Issue4003` against the base `main` version of `ObjectWriterImplInstant.java` and got `Tests run: 4, Failures: 0, Errors: 0`. The suite passes with or without the fix, so it does not actually guard the issue-4003 regression.

Root cause: `Context.setDateFormat("iso8601")` recomputes all shape flags in one shot — it sets `dateFormatISO8601 = true` and `formatyyyyMMddhhmmss19 = false` in the same call, and both flags are only ever assigned inside `setDateFormat`, so they are mutually exclusive. After `ctx.setDateFormat("iso8601")` the context is equivalent to `new Context("iso8601")`, and the old code already took the ISO8601 branch because `yyyyMMddhhmmss19` was already `false`. The conflicting state this patch guards against (`iso8601` true **and** a `yyyyMMdd*` flag true) is not reachable through `setDateFormat`.

To make this a real regression test it needs to reproduce the actual trigger and fail on `main` without this change. If the real-world report originates from `SerializerFeature.UseISO8601DateFormat` / a feature bit rather than `setDateFormat("iso8601")`, note that the `iso8601` variable here only consults `formatISO8601` and `context.isDateFormatISO8601()` — it never inspects a feature flag — so this patch may not cover that path either. Worth confirming how 4003 reproduces before relying on this fix.

_— automated review via /qreview_

wenshao (commented):
Automated review via `/qreview`. 1 suggestion inline. Submitted as a Comment (not Approve / Request changes) because this is a self-authored PR and CI is still running.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
```diff
@@ -36,13 +36,24 @@ public void write(JSONWriter jsonWriter, Object object, Object fieldName, Type f
...
 
         Instant instant = (Instant) object;
-        if (dateFormat == null) {
+
+        // Honor formatISO8601 / context.isDateFormatISO8601() ahead of any
+        // string-pattern shape flags (yyyyMMddHHmmss19 etc.) — when the caller
+        // explicitly requested ISO8601 (e.g. SerializerFeature.UseISO8601DateFormat
+        // or JSONWriter.Feature.WriteISO8601DateFormat) it must take priority over
+        // a default dateFormat that may have been set on the context (issue #4003).
+        boolean iso8601 = formatISO8601 || (this.format == null && context.isDateFormatISO8601());
```
wenshao, line 45: **[Critical]** This change does not fix #4003 — it is behaviorally inert in core, and the real reproduction path erases the ISO8601 request before this writer runs.

**1. The gating is unreachable in core.** `iso8601 == true` requires `formatISO8601` (which implies `this.format == "iso8601"`, so all `this.yyyyMMdd*` shape flags derived from the same format string are false) or `context.isDateFormatISO8601()` (which implies `context.dateFormat == "iso8601"`). `JSONWriter.Context.setDateFormat` (JSONWriter.java:3635–3670) recomputes `dateFormatISO8601` and `formatyyyyMMddhhmmss19` atomically in one switch, and `dateFormatISO8601` has no other setter in core — so the conflicting state this patch guards (ISO8601 requested **and** a `yyyyMMdd*` shape flag set) cannot occur. Empirically: with this file reverted to the base commit, all four `Issue4003` tests still pass (`Tests run: 4, Failures: 0`).

**2. The real #4003 path clobbers the flag before write time.** The reporter's setup is fastjson1-compatible + Spring (`config.setSerializerFeatures(SerializerFeature.UseISO8601DateFormat)`). Failure scenario: `FastJsonHttpMessageConverter.writeInternal` → `JSON.writeJSONString(baos, object, fastJsonConfig.getDateFormat(), filters, features)`, and `FastJsonConfig.getDateFormat()` (FastJsonConfig.java:188–190) always returns `JSON.DEFAULT_DATE_FORMAT` = `"yyyy-MM-dd HH:mm:ss"`. In that overload (fastjson1-compatible JSON.java:2107–2116), `createWriteContext` first applies `UseISO8601DateFormat` → `context.setDateFormat("iso8601")` (JSON.java:1250), then `context.setDateFormat(dateFormat)` overwrites it → `dateFormatISO8601 = false`, `formatyyyyMMddhhmmss19 = true` → `iso8601` here evaluates to `false` → output `"2026-02-24 14:50:00"` — exactly the wrong output reported. The same pattern exists in the sibling overloads at JSON.java:1371, 1536, 1779, 1924, 2051.

Suggested fix (in fastjson1-compatible, not here): do not let the `dateFormat` parameter clobber an explicit ISO8601 feature:

```java
JSONWriter.Context context = createWriteContext(SerializeConfig.global, DEFAULT_GENERATE_FEATURE, features);
if (dateFormat != null && !context.isDateFormatISO8601()) {
    context.setDateFormat(dateFormat);
}
```

(or make `FastJsonConfig.getDateFormat()` return `null` when the user has not explicitly set a format). Then add a regression test in fastjson1-compatible that fails on `main` — e.g. `JSON.writeJSONString(os, bean, "yyyy-MM-dd HH:mm:ss", filters, SerializerFeature.UseISO8601DateFormat)` asserting ISO8601 output. The current `Issue4003` tests only exercise `ctx.setDateFormat("iso8601")` directly, which never hits the clobbering sequence and passes with or without this PR's production change.

_— Claude (Anthropic) via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
```diff
@@ -36,13 +36,24 @@ public void write(JSONWriter jsonWriter, Object object, Object fieldName, Type f
                 : context.getDateFormat();
 
         Instant instant = (Instant) object;
-        if (dateFormat == null) {
+
+        // Honor formatISO8601 / context.isDateFormatISO8601() ahead of any
+        // string-pattern shape flags (yyyyMMddHHmmss19 etc.) — when the caller
+        // explicitly requested ISO8601 (e.g. SerializerFeature.UseISO8601DateFormat
+        // or JSONWriter.Feature.WriteISO8601DateFormat) it must take priority over
```
wenshao, line 43: **[Suggestion]** `JSONWriter.Feature.WriteISO8601DateFormat` does not exist — a repo-wide grep for `WriteISO8601DateFormat` matches only this comment and this PR's test Javadoc. Concrete cost: a maintainer tracing how ISO8601 is requested (e.g. while investigating #4003) searches for this enum constant, finds nothing, and is pointed away from the real mechanisms (`context.setDateFormat("iso8601")` / `SerializerFeature.UseISO8601DateFormat`).

```suggestion
        // explicitly requested ISO8601 (e.g. context.setDateFormat("iso8601")
        // or fastjson1-compatible SerializerFeature.UseISO8601DateFormat) it must take priority over
```

_— Claude (Anthropic) via Qwen Code /review_

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running. Reviewed. Suggestions are inline.

_— Claude (Anthropic) via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
```diff
@@ -36,13 +36,24 @@ public void write(JSONWriter jsonWriter, Object object, Object fieldName, Type f
...
             jsonWriter.writeInstant(instant);
             return;
         }
 
-        boolean yyyyMMddhhmmss19 = this.yyyyMMddhhmmss19 || (context.isFormatyyyyMMddhhmmss19() && this.format == null);
-        if (yyyyMMddhhmmss14 || yyyyMMddhhmmss19 || yyyyMMdd8 || yyyyMMdd10) {
+        boolean yyyyMMddhhmmss19 = !iso8601 && (this.yyyyMMddhhmmss19 || (context.isFormatyyyyMMddhhmmss19() && this.format == null));
+        boolean yyyyMMddhhmmss14local = !iso8601 && yyyyMMddhhmmss14;
+        boolean yyyyMMdd8local = !iso8601 && yyyyMMdd8;
+        boolean yyyyMMdd10local = !iso8601 && yyyyMMdd10;
```
wenshao, line 55: **[Suggestion]** The `!iso8601` guard is applied via `*local` variables at the outer `if` condition, but the inner dispatch (~lines 140, 149) still references the ungated inherited fields `yyyyMMddhhmmss14` and `yyyyMMdd10`, while `yyyyMMddhhmmss19` accidentally uses the gated local because it shadows the field name. — Concrete cost: a future refactor that restructures the outer guard would silently re-introduce the ISO8601-vs-shape-flag priority bug for the 14/10 branches while the 19 branch remains correct — an asymmetric failure masked by the naming inconsistency.

```suggestion
                boolean yyyyMMddhhmmss14 = !iso8601 && this.yyyyMMddhhmmss14;
                boolean yyyyMMdd8 = !iso8601 && this.yyyyMMdd8;
                boolean yyyyMMdd10 = !iso8601 && this.yyyyMMdd10;
```

core/src/test/java/com/alibaba/fastjson2/issues_4000/Issue4003.java
```diff
@@ -0,0 +1,75 @@
...
+        Bean b = new Bean();
+        b.time = Instant.parse("2026-02-24T14:50:00Z");
+        // Simulates Spring Boot's FastJsonConfig: default dateFormat is
+        // "yyyy-MM-dd HH:mm:ss", then the user later switches to ISO8601.
+        JSONWriter.Context ctx = utcContext("yyyy-MM-dd HH:mm:ss");
+        ctx.setDateFormat("iso8601");
+        String json = JSON.toJSONString(b, ctx);
```
wenshao, line 54: **[Suggestion]** `testIso8601OverridesDefaultDateFormat` does not exercise the code path this PR changes. `setDateFormat("iso8601")` atomically replaces all format state (clearing `formatyyyyMMddhhmmss19`), making this test functionally identical to `testIso8601ContextDateFormat` — it passes identically on pre-change code. — Failure scenario: if a future regression breaks the `!iso8601` guards, this test still passes and the regression ships. The test verifies only that `setDateFormat("iso8601")` produces ISO8601 output, which was already true before this diff.

```suggestion
        // Note: setDateFormat("iso8601") atomically replaces all format state,
        // so this is a regression guard for setDateFormat replacement semantics
        // rather than a true priority-conflict test.
        JSONWriter.Context ctx = utcContext("yyyy-MM-dd HH:mm:ss");
        ctx.setDateFormat("iso8601");
        String json = JSON.toJSONString(b, ctx);
```

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplInstant.java
```diff
@@ -36,13 +36,24 @@ public void write(JSONWriter jsonWriter, Object object, Object fieldName, Type f
...
+        // Honor formatISO8601 / context.isDateFormatISO8601() ahead of any
+        // string-pattern shape flags (yyyyMMddHHmmss19 etc.) — when the caller
+        // explicitly requested ISO8601 (e.g. SerializerFeature.UseISO8601DateFormat
+        // or JSONWriter.Feature.WriteISO8601DateFormat) it must take priority over
+        // a default dateFormat that may have been set on the context (issue #4003).
+        boolean iso8601 = formatISO8601 || (this.format == null && context.isDateFormatISO8601());
+
+        if (dateFormat == null && !iso8601) {
             jsonWriter.writeInstant(instant);
             return;
```
wenshao, line 49: **[Suggestion]** The relaxed early-return condition (`dateFormat == null && !iso8601`) makes the stale `dateFormatISO8601` flag from `setDateFormat(null)` newly reachable. `setDateFormat(null)` sets `dateFormat = null` but skips the boolean-reset block (JSONWriter.java:3634 guards on `dateFormat != null`), leaving `dateFormatISO8601 = true`. — Failure scenario: a direct API user calls `ctx.setDateFormat("iso8601")` then `ctx.setDateFormat(null)` to get default Instant serialization; they now get ISO8601 instead of epoch-seconds. Pre-existing bug in `setDateFormat(null)`, but this diff makes it observable here.

Consider fixing the root cause in `JSONWriter.Context.setDateFormat` by resetting the boolean flags when `dateFormat` is null or empty.

core/src/test/java/com/alibaba/fastjson2/issues_4000/Issue4003.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static JSONWriter.Context utcContext(String dateFormat) {
+        JSONWriter.Context ctx = new JSONWriter.Context(dateFormat);
+        ctx.setZoneId(ZoneId.of("UTC"));
+        return ctx;
+    }
+
+    @Test
+    public void testIso8601ContextDateFormat() {
+        Bean b = new Bean();
```
wenshao, line 38: **[Suggestion]** `testIso8601ContextDateFormat` also passes identically on pre-change code — together with `testIso8601OverridesDefaultDateFormat`, neither ISO8601 test exercises the modified early-return guard or the `!iso8601` shape-flag gating. — Failure scenario: all four tests pass with the production change reverted, so the suite provides zero regression protection for the changed behavior. A test that makes `context.isDateFormatISO8601()` true while `context.getDateFormat()` returns null (e.g. via the stale-flag path: `setDateFormat("iso8601")` then `setDateFormat(null)`) would actually distinguish old from new code.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running. Reviewed. Suggestions are inline. 1 Suggestion-level finding(s) could not be anchored to a changed line and were dropped; nothing further to act on here. Not reviewed: build-and-test — no JDK installed on this machine; Maven compile and test could not run. **[Critical]** Existing blocker (comment 3655578895, ObjectWriterImplInstant.java:45) still stands: this change does not fix #4003 — the production change is behaviorally inert in core because the conflicting state (ISO8601 requested AND a yyyyMMdd* shape flag set) cannot occur through any code path in core. The real reproduction path in fastjson1-compatible (FastJsonHttpMessageConverter.writeInternal → JSON.writeJSONString with dateFormat parameter) clobbers the ISO8601 flag before this writer runs. Empirically verified: all four Issue4003 tests pass with the production change reverted.

PR #7633, opened 2026-04-25
fix: avoid NPE in trustedLookup on environments without IMPL_LOOKUP, for issue #4011

### LINKED ISSUE #4011: [BUG] NullPointerException: java.lang.invoke.MethodHandles$Lookup java.lang.invoke.MethodHandles$Lookup.in(java.lang.Class)
### 问题描述
在 Android 环境下使用 JSON.parseObject、JSON.toJSONString 等方法时抛出 NullPointerException：

   java.lang.NullPointerException: Attempt to invoke virtual method
   'java.lang.invoke.MethodHandles$Lookup java.lang.invoke.MethodHandles$Lookup.in(java.lang.Class)'
   on a null object reference
### 环境信息
*请填写以下信息：*

   - OS 信息：Android 16 (API 36)
   - JDK 信息：JDK 21 (编译)
   - 版本信息：Fastjson2 2.0.61
 
### 重现步骤
*如何操作可以重现该问题：*
```kotlin
 data class TestModel(
        var id: Int = 0,
        var name: String = "",
        var value: String = ""
    )

val jsonString = """
            {
                "id": 1,
                "name": "test",
                "value": "hello"
            }
        """.trimIndent()

val result = JSON.parseObject(jsonString, TestModel::class.java)
val result = jsonString.parseObject<TestModel>()
```

### 期待的正确结果
预期是返回
```kotlin
TestModel(id=1, name=test, value=hello)
```

wenshao:
## Summary
- On Android the static initializer leaves `JDKUtils.IMPL_LOOKUP` null (the Unsafe-based access to `MethodHandles.Lookup.IMPL_LOOKUP` is intentionally gated to `!ANDROID`). `trustedLookup(Class)` then NPE'd at `IMPL_LOOKUP.in(...)`, producing the user-visible \"Attempt to invoke virtual method `MethodHandles$Lookup.in(...)` on a null object reference\" when parsing Kotlin data classes via `JSON.parseObject` on Android 16 (API 36).
- `JDKUtils.trustedLookup`: short-circuit and return `null` when `IMPL_LOOKUP` is unavailable. Most existing callers already wrap usage in `try/catch` and fall back to reflection-based access, so a null `Lookup` is the cleanest signal.
- `ObjectReaderCreator.lambdaSetter` / `createBuildFunctionLambda`: handle the null lookup explicitly and return null so callers fall back to reflection instead of throwing `JSONException(\"create fieldReader error\")`.

Fixes #4011.

## Test plan
- [x] Existing `JDKUtilsTest` (uses `trustedLookup` with non-null `IMPL_LOOKUP`) still passes.
- [x] `ObjectReaderCreator*Test*`, `LambdaSetterTest*` — all 38 tests pass.
- [ ] Verified on Android (would need an Android device run; reproducer in issue body).

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreator.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreator.java
@@ -2186,6 +2186,9 @@ public class ObjectReaderCreator {
      */
     <T, R> Function<T, R> createBuildFunctionLambda(Method builderMethod) {
         MethodHandles.Lookup lookup = JDKUtils.trustedLookup(builderMethod.getDeclaringClass());
+        if (lookup == null) {
+            return null;
+        }
         try {
             MethodHandle target = lookup.findVirtual(builderMethod.getDeclaringClass(),
                     builderMethod.getName(),
@@ -3994,6 +3997,12 @@ public class ObjectReaderCreator {
 
     protected Object lambdaSetter(Class objectClass, Class fieldClass, Method method) {
         MethodHandles.Lookup lookup = JDKUtils.trustedLookup(objectClass);
+        if (lookup == null) {
+            // Trusted MethodHandles.Lookup not available (e.g. Android, where
+            // IMPL_LOOKUP cannot be obtained via Unsafe). Skip lambda creation
+            // and let the caller fall back to reflection-based access.
+            return null;
+        }
 
         Class<?> returnType = method.getReturnType();
         LambdaSetterInfo lambdaInfo = methodTypeMapping.get(fieldClass);
--- a/core/src/main/java/com/alibaba/fastjson2/util/JDKUtils.java
+++ b/core/src/main/java/com/alibaba/fastjson2/util/JDKUtils.java
@@ -470,6 +470,15 @@ public class JDKUtils {
     }
 
     public static MethodHandles.Lookup trustedLookup(Class objectClass) {
+        // IMPL_LOOKUP is null on Android (see static init guard) and may also be
+        // null on environments where neither the Unsafe-based access nor
+        // MethodHandles.lookup() were available. All callers wrap usage in a
+        // try/catch and fall back to plain reflection, so signal that here by
+        // returning null instead of NPE'ing on the .in(...) call below.
+        if (IMPL_LOOKUP == null) {
+            return null;
+        }
+
         if (!CONSTRUCTOR_LOOKUP_ERROR) {
             try {
                 int TRUSTED = -1;


core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreator.java
```diff
@@ -2186,6 +2186,9 @@ public <T, R> Function<T, R> createBuildFunction(Method builderMethod) {
      */
     <T, R> Function<T, R> createBuildFunctionLambda(Method builderMethod) {
         MethodHandles.Lookup lookup = JDKUtils.trustedLookup(builderMethod.getDeclaringClass());
+        if (lookup == null) {
```
wenshao, line 2189: **[Critical]** `createBuildFunctionLambda` returns `null` on Android, but the caller `createBuildFunction` (line ~2161) does `return createBuildFunctionLambda(builderMethod)` inside a `try` block. A `null` return completes the `try` block normally — no exception is thrown — so the `catch` block is skipped and the reflection-based fallback (after `catch`) is **never reached**.

Before this PR, the NPE from `trustedLookup` was caught by `catch (Throwable)`, which accidentally allowed the reflection fallback to run. After this PR, the reflection fallback is dead code on Android, causing builder-pattern deserialization to silently return the intermediate builder object instead of the final built object.

```suggestion
        if (lookup == null) {
            return null; // caller createBuildFunction must check for null and fall back to reflection
        }
```

Fix `createBuildFunction` to handle null return:
```java
try {
    Function<T, R> result = createBuildFunctionLambda(builderMethod);
    if (result != null) {
        return result;
    }
    // null means trustedLookup unavailable — fall through to reflection below
} catch (Throwable e) {
    jitErrorCount.incrementAndGet();
    jitErrorLast = e;
}
```

core/src/main/java/com/alibaba/fastjson2/util/JDKUtils.java
```diff
@@ -470,6 +470,15 @@ public static char[] getCharArray(String str) {
     }
 
     public static MethodHandles.Lookup trustedLookup(Class objectClass) {
+        // IMPL_LOOKUP is null on Android (see static init guard) and may also be
+        // null on environments where neither the Unsafe-based access nor
+        // MethodHandles.lookup() were available. All callers wrap usage in a
+        // try/catch and fall back to plain reflection, so signal that here by
```
wenshao, line 476: **[Suggestion]** The comment claims "All callers wrap usage in a try/catch and fall back to plain reflection" but this is factually inaccurate:

- `ObjectWriterCreator.lambdaGetter` catches Throwable and **re-throws** as `JSONException` (no graceful fallback)
- Several `PropertyAccessorFactoryLambda` methods (`getObject`, `getBoolean`, `setInt`, etc.) catch Throwable and **re-throw** as `RuntimeException`

Suggested revision:
```java
        // Returns null when IMPL_LOOKUP is unavailable (e.g. Android).
        // Most callers handle this gracefully via try/catch with reflection
        // fallbacks; callers that re-throw rely on their upstream callers
        // to catch and fall back.
```

wenshao (commented):
⚠️ Downgraded from Request Changes to Comment: self-PR; CI still running.

**[Suggestion]** PR adds 3 new null-return early-exit paths (`trustedLookup`, `createBuildFunctionLambda`, `lambdaSetter`) but no test verifies behavior when `trustedLookup` returns null. A test that mocks `IMPL_LOOKUP = null` would have caught the `createBuildFunction` regression. Consider adding:
1. Test `trustedLookup` returns null when `IMPL_LOOKUP` is null
2. Test `createBuildFunction` still produces a working reflection-based Function when `createBuildFunctionLambda` returns null

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreator.java
```diff
@@ -2186,6 +2186,9 @@ public <T, R> Function<T, R> createBuildFunction(Method builderMethod) {
...
+            return null;
+        }
```
wenshao, line 2191: **[Critical]** Returning `null` here makes `createBuildFunctionLambda` return `null` on Android (where `IMPL_LOOKUP` is null), but its sole caller `createBuildFunction` does `return createBuildFunctionLambda(builderMethod);` inside a `try` block. A `null` return completes the `try` normally — no exception is thrown — so the `catch (Throwable)` is skipped and the reflection-based fallback after it is never reached. — Failure scenario: on Android, deserialize any builder-pattern class (`@JSONType(builder=...)`). Before this PR the `trustedLookup` NPE was caught by `catch (Throwable)` and the reflection fallback ran; after this PR `createBuildFunction` returns `null`, so `builderFunction` is null at line 525 and `ObjectReaderAdapter` (guarded by `if (buildFunction != null)`) returns the intermediate builder object instead of the built object — a silent wrong result. Fix the caller to handle a null return:

```java
public <T, R> Function<T, R> createBuildFunction(Method builderMethod) {
    try {
        Function<T, R> function = createBuildFunctionLambda(builderMethod);
        if (function != null) {
            return function;
        }
    } catch (Throwable e) {
        jitErrorCount.incrementAndGet();
        jitErrorLast = e;
    }
    // ... existing reflection fallback below ...
}
```

core/src/main/java/com/alibaba/fastjson2/util/JDKUtils.java
```diff
@@ -470,6 +470,15 @@ public static char[] getCharArray(String str) {
...
+        // returning null instead of NPE'ing on the .in(...) call below.
```
wenshao, line 477: **[Suggestion]** The claim "All callers wrap usage in a try/catch and fall back to plain reflection" is inaccurate. A `null` return does not trigger any `try/catch` (no exception is thrown) — which is exactly the mechanism behind the `createBuildFunction` regression above — and several callers re-throw as `JSONException` rather than falling back (e.g. `ObjectWriterCreator.lambdaGetter`, `MoneySupport`). — Concrete cost: a future maintainer trusts this invariant when adding a caller, omits a null check, and the null silently propagates past a try/catch to a deferred NPE far from the root cause.

```suggestion
        // IMPL_LOOKUP is null on Android (see static init guard) and may also be
        // null on environments where neither the Unsafe-based access nor
        // MethodHandles.lookup() were available. Return null so callers can
        // detect the unavailable case; callers must handle null (a null return
        // does NOT trigger their try/catch).
```

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI failing: Test Reflect JDK 8 on windows-latest. Reviewed. Suggestions are inline.

PR #7657, opened 2026-06-13
fix: global naming strategy not applied to Java records (#7656)

### LINKED ISSUE #7656: [BUG] global SnakeCase naming strategy is not applied to Java records
## Description

When fastjson2 global writer naming strategy is configured with `PropertyNamingStrategy.SnakeCase`, serialization applies snake_case naming to normal Java class fields, but not to Java records.

This causes DTOs implemented as records to serialize camelCase fields into final JSON even though global snake_case output is configured.

## Environment

- Verified versions: `2.0.60`, `2.0.61`, `2.0.62`
- Java: 21

## Reproduction

```java
import com.alibaba.fastjson2.JSONFactory;
import com.alibaba.fastjson2.PropertyNamingStrategy;

record RecordDto(String currencyCode) {}

class ClassDto {
    public String currencyCode = "840";
}

JSONFactory.getDefaultObjectWriterProvider()
        .setNamingStrategy(PropertyNamingStrategy.SnakeCase);

System.out.println(JSON.toJSONString(new RecordDto("840")));
System.out.println(JSON.toJSONString(new ClassDto()));
```

## Expected Result

Both DTOs serialize using snake_case:

```json
{"currency_code":"840"}
{"currency_code":"840"}
```

## Actual Result

Only the normal class is snake-cased. The record remains camelCase:

```json
{"currencyCode":"840"}
{"currency_code":"840"}
```

## Workaround

Use a normal Java class instead of a record, or annotate each affected record component with `@JSONField(name = "snake_case_name")`.

## Additional Notes

Also verified that `@JSONType(naming = PropertyNamingStrategy.SnakeCase)` and `JSONWriter.Feature.FieldBased` did not change the record output in the local probe.

1919chichi:
## Summary

Fixes #7656

When serializing Java records, `ObjectWriterCreator` was calling `method.getName()` directly to get the field name, bypassing `BeanUtils.getterName(method, kotlin, namingStrategy)`. This meant that both the global naming strategy (set via `JSONFactory.getDefaultObjectWriterProvider().setNamingStrategy(...)`) and the per-type `@JSONType(naming = ...)` annotation were silently ignored for record components.

**Root cause** (`ObjectWriterCreator.java`):
```java
// Before — records bypassed the naming strategy entirely
if (record) {
    fieldName = method.getName();
} else {
    fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
    ...
}

// After — records go through the same naming pipeline as regular classes
fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
if (!record) {
    // NAME_COMPATIBLE_WITH_FILED logic (not applicable to records)
    ...
}
```

## Test plan

- [x] `JSONTypeNamingSnake#testRecord` — verifies `@JSONType(naming = SnakeCase)` applies to records
- [x] `JSONTypeNamingSnake#testRecordGlobalNaming` — verifies global `setNamingStrategy(SnakeCase)` applies to records
- [x] `./mvnw -pl core clean test` — all tests pass
- [x] `./mvnw -pl core -Dfastjson2.creator=reflect clean test` — reflect mode tests pass
- [x] `./mvnw -pl core validate` — Checkstyle passes

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
@@ -753,8 +753,10 @@ public class ObjectWriterCreator {
     ) {
         String fieldName;
         if (fieldInfo.fieldName == null || fieldInfo.fieldName.isEmpty()) {
-            fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
-            if (!record) {
+            if (record) {
+                fieldName = BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy);
+            } else {
+                fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
                 Field field;
                 if ((provider.userDefineMask & NAME_COMPATIBLE_WITH_FILED) != 0
                         && (field = BeanUtils.getField(objectClass, method)) != null) {
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
@@ -754,7 +754,9 @@ public class ObjectWriterCreator {
         String fieldName;
         if (fieldInfo.fieldName == null || fieldInfo.fieldName.isEmpty()) {
             if (record) {
-                fieldName = BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy);
+                fieldName = beanInfo.namingStrategy != null
+                        ? BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy)
+                        : method.getName();
             } else {
                 fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
                 Field field;
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
@@ -753,11 +753,8 @@ public class ObjectWriterCreator {
     ) {
         String fieldName;
         if (fieldInfo.fieldName == null || fieldInfo.fieldName.isEmpty()) {
-            if (record) {
-                fieldName = method.getName();
-            } else {
-                fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
-
+            fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
+            if (!record) {
                 Field field;
                 if ((provider.userDefineMask & NAME_COMPATIBLE_WITH_FILED) != 0
                         && (field = BeanUtils.getField(objectClass, method)) != null) {
--- a/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
+++ b/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.annotation;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONFactory;
 import com.alibaba.fastjson2.PropertyNamingStrategy;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -50,6 +51,33 @@ public class JSONTypeNamingSnake {
 
         public void setBeanName(String beanName) {
             this.beanName = beanName;
+        }
+    }
+
+    @JSONType(naming = PropertyNamingStrategy.SnakeCase)
+    public record RecordDto(String currencyCode) {}
+
+    @JSONType(naming = PropertyNamingStrategy.SnakeCase)
+    static class ClassDto {
+        public String currencyCode = "840";
+    }
+
+    @Test
+    public void testRecord() {
+        assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new RecordDto("840")));
+        assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new ClassDto()));
+    }
+
+    @Test
+    public void testRecordGlobalNaming() {
+        var provider = JSONFactory.getDefaultObjectWriterProvider();
+        var prev = provider.getNamingStrategy();
+        try {
+            provider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
+            record GlobalRecordDto(String currencyCode) {}
+            assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
+        } finally {
+            provider.setNamingStrategy(prev);
         }
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
+++ b/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.annotation;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.PropertyNamingStrategy;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
--- a/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
+++ b/core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
@@ -1,8 +1,6 @@
 package com.alibaba.fastjson2.annotation;
 
 import com.alibaba.fastjson2.JSON;
-import com.alibaba.fastjson2.JSONFactory;
-import com.alibaba.fastjson2.PropertyNamingStrategy;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
@@ -55,29 +53,12 @@ public class JSONTypeNamingSnake {
     }
 
     @JSONType(naming = PropertyNamingStrategy.SnakeCase)
-    public record RecordDto(String currencyCode) {}
-
-    @JSONType(naming = PropertyNamingStrategy.SnakeCase)
     static class ClassDto {
         public String currencyCode = "840";
     }
 
     @Test
-    public void testRecord() {
-        assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new RecordDto("840")));
+    public void testClass() {
         assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new ClassDto()));
-    }
-
-    @Test
-    public void testRecordGlobalNaming() {
-        var provider = JSONFactory.getDefaultObjectWriterProvider();
-        var prev = provider.getNamingStrategy();
-        try {
-            provider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
-            record GlobalRecordDto(String currencyCode) {}
-            assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
-        } finally {
-            provider.setNamingStrategy(prev);
-        }
     }
 }
--- a/test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java
+++ b/test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2;
 
 import com.alibaba.fastjson2.annotation.JSONType;
+import com.alibaba.fastjson2.reader.ObjectReaderProvider;
 import com.alibaba.fastjson2.writer.ObjectWriterProvider;
 import org.junit.jupiter.api.Test;
 
@@ -29,13 +30,18 @@ public class RecordNamingStrategyTest {
 
     @Test
     public void testRecordGlobalNaming() {
-        ObjectWriterProvider provider = JSONFactory.getDefaultObjectWriterProvider();
-        PropertyNamingStrategy prev = provider.getNamingStrategy();
+        ObjectWriterProvider writerProvider = JSONFactory.getDefaultObjectWriterProvider();
+        ObjectReaderProvider readerProvider = JSONFactory.getDefaultObjectReaderProvider();
+        PropertyNamingStrategy prevWriter = writerProvider.getNamingStrategy();
+        PropertyNamingStrategy prevReader = readerProvider.getNamingStrategy();
         try {
-            provider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
+            writerProvider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
+            readerProvider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
             assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
+            assertEquals(new GlobalRecordDto("840"), JSON.parseObject("{\"currency_code\":\"840\"}", GlobalRecordDto.class));
         } finally {
-            provider.setNamingStrategy(prev);
+            writerProvider.setNamingStrategy(prevWriter);
+            readerProvider.setNamingStrategy(prevReader);
         }
     }
 }
--- a/test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java
+++ b/test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java
@@ -1,0 +1,41 @@
+package com.alibaba.fastjson2;
+
+import com.alibaba.fastjson2.annotation.JSONType;
+import com.alibaba.fastjson2.writer.ObjectWriterProvider;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class RecordNamingStrategyTest {
+    @JSONType(naming = PropertyNamingStrategy.SnakeCase)
+    public record RecordDto(String currencyCode) {}
+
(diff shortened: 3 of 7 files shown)


core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
```diff
@@ -753,11 +753,8 @@ protected static String getFieldName(
     ) {
         String fieldName;
         if (fieldInfo.fieldName == null || fieldInfo.fieldName.isEmpty()) {
-            if (record) {
-                fieldName = method.getName();
-            } else {
-                fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
-
+            fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
```
wenshao: **[Critical] `is`/`get` prefix stripping corrupts record field names**

`BeanUtils.getterName(method, kotlin, namingStrategy)` applies JavaBean `is`/`get` prefix-stripping before the naming strategy transform. Record accessor methods have no such prefix — the method name IS the field name. For records with boolean `is`-prefixed components (very common: `isActive`, `isEnabled`, `isValid`), the prefix is silently stripped:

| Record | Before PR (correct) | After PR (broken) |
|---|---|---|
| `record Foo(boolean isActive)` | `{"isActive":true}` | `{"active":true}` |
| `record Bar(String getValue)` | `{"getValue":"x"}` | `{"value":"x"}` |

**Root cause:** `BeanUtils.getterName(Method, boolean, String)` has an early-return guard for `is`-prefixed methods, but it only fires when the return type is NOT `boolean`. For boolean-returning `isActive()` on a record, the guard does NOT fire, and execution falls through to `getterName("isActive", namingStrategy)` which strips the prefix.

The existing tests use `currencyCode` (no `is`/`get` prefix) so this regression is invisible.

**Suggested fix:** For records, apply the naming strategy to the raw method name without JavaBean prefix detection. A dedicated helper like `BeanUtils.applyNamingStrategy(String rawName, String namingStrategy)` with `prefixLength = 0` would be safest:

```java
if (record) {
    fieldName = method.getName();
    if (beanInfo.namingStrategy != null) {
        // apply naming strategy without is/get prefix stripping
        fieldName = BeanUtils.applyNamingStrategy(fieldName, beanInfo.namingStrategy);
    }
} else {
    fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
}
```

core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
```diff
@@ -52,4 +53,31 @@ public void setBeanName(String beanName) {
...
+        assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new ClassDto()));
+    }
+
+    @Test
+    public void testRecordGlobalNaming() {
+        var provider = JSONFactory.getDefaultObjectWriterProvider();
+        var prev = provider.getNamingStrategy();
+        try {
+            provider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
+            record GlobalRecordDto(String currencyCode) {}
```
wenshao: **[Critical] Test compilation failure on JDK 11**

`record GlobalRecordDto(String currencyCode) {}` is a local record (Java 16+, JEP 395), but the project uses `maven.compiler.source=8` and CI runs JDK 11. The CI check `Test JDK 11 on ubuntu-24.04` is already failing with:

```
JSONTypeNamingSnake.java:[77,35] ';' expected
```

**Suggested fix:** Move `GlobalRecordDto` to a static inner record at the class level (like `RecordDto` on line 58), and replace `var` with explicit types:

```java
@JSONType(naming = PropertyNamingStrategy.SnakeCase)
public record GlobalRecordDto(String currencyCode) {}

@Test
public void testRecordGlobalNaming() {
    ObjectWriterProvider provider = JSONFactory.getDefaultObjectWriterProvider();
    PropertyNamingStrategy prev = provider.getNamingStrategy();
    try {
        provider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
        assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
    } finally {
        provider.setNamingStrategy(prev);
    }
}
```

Note: Even static inner records require Java 16+ — if CI must pass on JDK 11 with `-source 8`, the record-based tests need to be in a separate module or profile compiled with a higher source level.

wenshao (changes requested):
**[Critical] Deserialization round-trip is broken for records with naming strategies**

This PR fixes only the writer side. The reader side (`ObjectReaderCreator` line ~194) assigns `fieldName = paramName` from the canonical constructor parameter names with no naming-strategy transformation. After this PR, `JSON.toJSONString(new RecordDto("840"))` produces `{"currency_code":"840"}`, but `JSON.parseObject("{\"currency_code\":\"840\"}", RecordDto.class)` fails to match the key `currency_code` against the raw parameter name `currencyCode` — the value is silently dropped.

The existing test only asserts serialization, not round-trip. The reader-side `createFieldReaders(provider, objectClass, objectType, owner, parameters, paramNames)` overload needs to apply `beanInfo.namingStrategy` to parameter names, similar to how the field-based overload applies it via `BeanUtils.fieldName(fieldName, namingStrategy)`.

artpaym:
@wenshao @1919chichi As it seems related to the changes, please also ensure code gen code also gets the same fix (if needed).

1919chichi:
Thanks for the thorough review! Here's a summary of the changes made to address all three points:

**1. `is`/`get` prefix stripping (Writer side)**
Restored the `if (record) / else` branch and replaced `BeanUtils.getterName()` with the existing `BeanUtils.fieldName(method.getName(), namingStrategy)` for records. This applies the naming strategy without any prefix stripping, so a field like `isActive` correctly serializes to `is_active` instead of `active`.

**2. Deserialization round-trip (Reader side)**
Fixed `ObjectReaderCreator`: after fetching record field names via `BeanUtils.getRecordFieldNames()`, the naming strategy is now applied to each parameter name so that incoming JSON keys (e.g. `currency_code`) hash-match correctly during constructor binding. Added round-trip deserialization assertions to the tests to cover this.

**3. JDK 11 compilation failure (Tests)**
Moved all record-based tests to the `test-jdk17` module (compiled with `source 17`). Replaced the local record and `var` with a class-level `static record` and explicit types. Also added a `BooleanRecordDto(boolean isActive)` test case to explicitly verify the `is` prefix is preserved.

core/src/test/java/com/alibaba/fastjson2/annotation/JSONTypeNamingSnake.java
```diff
@@ -1,7 +1,6 @@
 package com.alibaba.fastjson2.annotation;
 
 import com.alibaba.fastjson2.JSON;
-import com.alibaba.fastjson2.PropertyNamingStrategy;
 import org.junit.jupiter.api.Tag;
```
wenshao, line 5: **[Critical]** Missing import causes compilation failure

The import `com.alibaba.fastjson2.PropertyNamingStrategy` was removed, but the class is still referenced at lines 27, 33, and 55 (`PropertyNamingStrategy.SnakeCase`). Since this file is in package `com.alibaba.fastjson2.annotation` and `PropertyNamingStrategy` lives in `com.alibaba.fastjson2`, the explicit import is required.

This causes `core` module test compilation to fail:
```
JSONTypeNamingSnake.java:[27,24] cannot find symbol
JSONTypeNamingSnake.java:[33,24] cannot find symbol
JSONTypeNamingSnake.java:[55,24] cannot find symbol
```

```suggestion
import com.alibaba.fastjson2.JSON;
import com.alibaba.fastjson2.PropertyNamingStrategy;
import org.junit.jupiter.api.Tag;
```

test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test
+    public void testBooleanFieldNotStripped() {
+        assertEquals("{\"is_active\":true}", JSON.toJSONString(new BooleanRecordDto(true)));
+        assertEquals(new BooleanRecordDto(true), JSON.parseObject("{\"is_active\":true}", BooleanRecordDto.class));
+    }
+
+    @Test
+    public void testRecordGlobalNaming() {
```
wenshao, line 32: **[Suggestion]** `testRecordGlobalNaming` only tests serialization — deserialization with global naming strategy is untested

The reader-side fix in `ObjectReaderCreator.java:1282-1286` applies naming strategy to record constructor parameter names when `beanInfo.namingStrategy` is non-null. For global strategies, this value comes from `ObjectReaderProvider` via the `BeanInfo` constructor. This entire code path has zero test coverage.

A regression in reader-side global naming support for records would go undetected.

```suggestion
    @Test
    public void testRecordGlobalNaming() {
        ObjectWriterProvider writerProvider = JSONFactory.getDefaultObjectWriterProvider();
        ObjectReaderProvider readerProvider = JSONFactory.getDefaultObjectReaderProvider();
        PropertyNamingStrategy prevWriter = writerProvider.getNamingStrategy();
        PropertyNamingStrategy prevReader = readerProvider.getNamingStrategy();
        try {
            writerProvider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
            readerProvider.setNamingStrategy(PropertyNamingStrategy.SnakeCase);
            assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
            assertEquals(new GlobalRecordDto("840"), JSON.parseObject("{\"currency_code\":\"840\"}", GlobalRecordDto.class));
        } finally {
            writerProvider.setNamingStrategy(prevWriter);
            readerProvider.setNamingStrategy(prevReader);
        }
    }
```

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java
```diff
@@ -754,10 +754,9 @@ protected static String getFieldName(
         String fieldName;
         if (fieldInfo.fieldName == null || fieldInfo.fieldName.isEmpty()) {
             if (record) {
-                fieldName = method.getName();
+                fieldName = BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy);
```
wenshao: **[Suggestion]** Missing null guard for `namingStrategy` — inconsistent with reader and sibling code

The record branch unconditionally calls `BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy)`. While `BeanUtils.fieldName()` handles null by defaulting to `"CamelCase"` (which is a no-op for standard lowercase-starting record names), this is inconsistent with:

- **Reader** (`ObjectReaderCreator.java:1282`): `if (beanInfo.namingStrategy != null && parameterNames != null)`
- **Writer field path** (line 243): `if (beanInfo.namingStrategy != null)`

Adding a null guard makes the code robust against future changes to `BeanUtils.fieldName()`'s null-handling and keeps the three paths consistent.

```suggestion
                fieldName = beanInfo.namingStrategy != null
                        ? BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy)
                        : method.getName();
```

wenshao (changes requested):

1919chichi:
Updated the PR to address the latest review comments:

- Restored the missing `PropertyNamingStrategy` import in `JSONTypeNamingSnake`.
- Kept record writer field-name handling on the raw accessor name and added an explicit `namingStrategy != null` guard, so record components such as `isActive` are not treated as JavaBean getters.
- Extended `RecordNamingStrategyTest#testRecordGlobalNaming` to cover global naming strategy deserialization via `ObjectReaderProvider` as well as serialization via `ObjectWriterProvider`.
- Restored the accidentally removed `lombok.var` imports in the `issue3601` tests, which caused the same kind of test compilation failure.

Validation run locally:

- `./mvnw -pl core -Dtest=JSONTypeNamingSnake test -q`
- `../mvnw test -Dtest=RecordNamingStrategyTest -q` from `test-jdk17/`
- `git diff --check`

wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI still running.

The implementation correctly integrates naming strategy support for Java records with proper reader/writer symmetry. `@JSONField` annotations are correctly resolved via constructor parameter annotations with proper precedence over naming strategy. Test coverage is adequate — serialization, deserialization, boolean prefix handling, and global naming strategy are all verified.

artpaym:
Hi. What's the status of this PR?

wenshao:
CI is all green, but the following issues need to be addressed before merging:

### 1. Reader-side missing naming strategy support (round-trip broken)
`ObjectReaderCreator` ~L194 does not apply naming strategy transformation to record canonical constructor parameter names. Serialization produces `currency_code`, but deserialization fails to match against `currencyCode` — the value is silently dropped.

Apply `beanInfo.namingStrategy` to parameter names (same as the field-based overload), and add a deserialization assertion in `testRecordGlobalNaming` to verify round-trip.

### 2. `is`/`get` prefix incorrectly stripped
Routing records through `BeanUtils.getterName()` strips `isActive` to `active`. Records need a separate path: apply naming strategy only, skip JavaBean prefix detection.

### 3. `namingStrategy` null guard
The writer record branch calls `BeanUtils.fieldName()` without a null check on `namingStrategy`, inconsistent with the reader side and neighboring code paths.

---
Once these three are fixed, this is good to merge. Thanks for the contribution!

<details>
<summary>中文版本</summary>

CI 已经全绿，但以下几个问题需要解决后才能合入：

#### 1. Reader 侧缺少 naming strategy 支持（round-trip 断掉）
`ObjectReaderCreator` ~L194 对 record canonical constructor 参数名没有做 naming strategy 转换。序列化产出 `currency_code`，反序列化时无法匹配 `currencyCode`，值被静默丢弃。

需要像 field-based overload 那样对参数名应用 `beanInfo.namingStrategy`，并在 `testRecordGlobalNaming` 中补充反序列化断言验证 round-trip。

#### 2. `is`/`get` 前缀误剥离
Record 直接走 `BeanUtils.getterName()` 会把 `isActive` 剥成 `active`。Record 应走独立路径：只做 naming strategy 转换，跳过 JavaBean 前缀检测。

#### 3. `namingStrategy` null guard
Writer record 分支调用 `BeanUtils.fieldName()` 前缺少 null 判断，与 reader 侧及相邻代码路径不一致。

修好这三点就可以合了，感谢贡献！
</details>

1919chichi:
> CI is all green, but the following issues need to be addressed before merging:
> 
> ### 1. Reader-side missing naming strategy support (round-trip broken)
> `ObjectReaderCreator` ~L194 does not apply naming strategy transformation to record canonical constructor parameter names. Serialization produces `currency_code`, but deserialization fails to match against `currencyCode` — the value is silently dropped.
> 
> Apply `beanInfo.namingStrategy` to parameter names (same as the field-based overload), and add a deserialization assertion in `testRecordGlobalNaming` to verify round-trip.
> 
> ### 2. `is`/`get` prefix incorrectly stripped
> Routing records through `BeanUtils.getterName()` strips `isActive` to `active`. Records need a separate path: apply naming strategy only, skip JavaBean prefix detection.
> 
> ### 3. `namingStrategy` null guard
> The writer record branch calls `BeanUtils.fieldName()` without a null check on `namingStrategy`, inconsistent with the reader side and neighboring code paths.
> 
> Once these three are fixed, this is good to merge. Thanks for the contribution!
> 
> 中文版本

Hi  All three issues have been addressed in `db6666405` ("fix: address record naming review comments"), on top of the reader-side fix already landed in `dc90fd47a`. Details below.

### 1. Reader-side missing naming strategy support (round-trip broken)
Fixed in `ObjectReaderCreator#createObjectReader` (`core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreator.java:1280-1287`):

```java
if (record && parameterNames == null) {
    parameterNames = BeanUtils.getRecordFieldNames(objectClass);
    if (beanInfo.namingStrategy != null && parameterNames != null) {
        for (int i = 0; i < parameterNames.length; i++) {
            parameterNames[i] = BeanUtils.fieldName(parameterNames[i], beanInfo.namingStrategy);
        }
    }
}
```

The record canonical-constructor parameter names now go through the same naming-strategy transform as the writer side, so a value serialized to `currency_code` deserializes correctly.

Covered by `testRecordGlobalNaming` in `test-jdk17/src/test/java/com/alibaba/fastjson2/RecordNamingStrategyTest.java:31-46`, which asserts both directions:
```java
assertEquals("{\"currency_code\":\"840\"}", JSON.toJSONString(new GlobalRecordDto("840")));
assertEquals(new GlobalRecordDto("840"), JSON.parseObject("{\"currency_code\":\"840\"}", GlobalRecordDto.class));
```

### 2. `is`/`get` prefix incorrectly stripped
Fixed in `ObjectWriterCreator#getFieldName` (`core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterCreator.java:756-759`). Records no longer route through `BeanUtils.getterName()` (which does JavaBean prefix detection); they apply the naming strategy directly to the raw accessor name:

```java
if (record) {
    fieldName = beanInfo.namingStrategy != null
            ? BeanUtils.fieldName(method.getName(), beanInfo.namingStrategy)
            : method.getName();
} else {
    fieldName = BeanUtils.getterName(method, beanInfo.kotlin, beanInfo.namingStrategy);
    ...
```

Added regression test `testBooleanFieldNotStripped` (`RecordNamingStrategyTest.java:25-29`) covering exactly the `isActive` case from the review:
```java
public record BooleanRecordDto(boolean isActive) {}
...
assertEquals("{\"is_active\":true}", JSON.toJSONString(new BooleanRecordDto(true)));
assertEquals(new BooleanRecordDto(true), JSON.parseObject("{\"is_active\":true}", BooleanRecordDto.class));
```

### 3. `namingStrategy` null guard
Fixed at the same location (`ObjectWriterCreator.java:757`), shown in the snippet above — the record branch now short-circuits to `method.getName()` when `namingStrategy` is null, matching the reader side (`ObjectReaderCreator.java:1282`) and the writer field-based path (`ObjectWriterCreator.java:243`).

---

All three tests pass locally:
```
$ mvn -f test-jdk17/pom.xml -am test -Dtest=RecordNamingStrategyTest
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

Let me know if anything still looks off.

wenshao (approved):
No issues found. LGTM! ✅

PR #7664, opened 2026-06-22
fix: respect custom readers for map subtypes

### LINKED ISSUE #7636: [BUG]Custom reader is not executed when deserializing a class that implements the Map interface
When deserializing a class that implements the Map interface, the custom ObjectReader is not executed.
In the following code, if Entity removes extends HashMap<String, Object>, the custom reader will be executed as expected.

```java
import com.alibaba.fastjson2.JSONReader;
import com.alibaba.fastjson2.reader.ObjectReader;
import java.lang.reflect.Type;

public class EntityReader implements ObjectReader<Entity> {
    @Override
    public Entity readObject(JSONReader jsonReader, Type type, Object o, long l) {
        return null;
    }
}
```
```java
import com.alibaba.fastjson2.annotation.JSONType;
import java.util.HashMap;

@JSONType(deserializer = EntityReader.class)
public class Entity extends HashMap<String, Object> {
    public int id;
}
```
```java
public class Test {
    public Entity entity;
}
```
```java

public class Main {
        Test obj = JSON.parseObject("{\"entity\":{\"id\":123}}", Test.class);
    }
}
```

After debugging, I found that the issue is caused by the FieldReaderObject<T> class, which uses ObjectReaderImplMap when the field type implements the Map interface:

```java
public ObjectReader getObjectReader(JSONReader jsonReader) {
    if (this.initReader != null) {
        return this.initReader;
    } else if (this.reader != null) {
        return this.reader;
    } else {
        ObjectReader formattedObjectReader = createFormattedObjectReader(this.fieldType, this.fieldClass, this.format, this.locale);
        if (formattedObjectReader != null) {
            return this.reader = formattedObjectReader;
        } else if (this.fieldClass != null && Map.class.isAssignableFrom(this.fieldClass)) {
            return this.reader = ObjectReaderImplMap.of(this.fieldType, this.fieldClass, this.features);
        } else {
            return this.fieldClass != null && Collection.class.isAssignableFrom(this.fieldClass) ? (this.reader = ObjectReaderImplList.of(this.fieldType, this.fieldClass, this.features)) : (this.reader = jsonReader.getObjectReader(this.fieldType));
        }
    }
}
```

hutiefang76:

Fixes #7636.

When a class extends `Map` and declares `@JSONType(deserializer = ...)`, parsing it as a bean field was still handled by the default map reader. This skipped the custom `ObjectReader`.

### Summary of your change

- Let explicitly annotated custom deserializers bypass the base Map/Collection reader shortcut.
- Make object field reader prefer a provider-resolved custom reader before falling back to default Map/List readers.
- Add regression coverage for the field case, direct parse case, and an unannotated map subtype to keep the existing default behavior unchanged.

Docs impact: none, this preserves the documented custom deserializer behavior.

Tests:

- `mvn -pl core -Dtest=Issue7636 test -DskipITs -ntp`
- `mvn -pl core -Dtest=Issue7636,JSONType_deserializer,Issue1168,Issue2329,Issue2726 test -DskipITs -ntp`
- `mvn -pl core -Dfastjson2.creator=reflect -Dtest=Issue7636,JSONType_deserializer,Issue1168,Issue2329,Issue2726 test -DskipITs -ntp`

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
@@ -9,11 +9,15 @@ import com.alibaba.fastjson2.writer.ObjectWriter;
 import java.lang.reflect.*;
 import java.time.temporal.Temporal;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 public class FieldReaderObject<T>
         extends FieldReader<T> {
+    static final ConcurrentMap<Class<?>, Boolean> JSONB_SUPPORTED_CACHE = new ConcurrentHashMap<>();
+
     protected ObjectReader initReader;
 
     public FieldReaderObject(
@@ -139,24 +143,27 @@ public class FieldReaderObject<T>
         return objectReader != null
                 && !(objectReader instanceof ObjectReaderImplMap)
                 && !(objectReader instanceof ObjectReaderImplMapTyped)
+                && !(objectReader instanceof ObjectReaderImplMapMultiValueType)
                 && !(objectReader instanceof ObjectReaderImplList)
                 && !(objectReader instanceof ObjectReaderImplListStr)
                 && !(objectReader instanceof ObjectReaderImplListInt64);
     }
 
     static boolean isReadJSONBObjectSupported(ObjectReader objectReader) {
-        try {
-            Method method = objectReader.getClass().getMethod(
-                    "readJSONBObject",
-                    JSONReader.class,
-                    Type.class,
-                    Object.class,
-                    long.class
-            );
-            return method.getDeclaringClass() != ObjectReader.class;
-        } catch (NoSuchMethodException e) {
-            return false;
-        }
+        return JSONB_SUPPORTED_CACHE.computeIfAbsent(objectReader.getClass(), objectReaderClass -> {
+            try {
+                Method method = objectReaderClass.getMethod(
+                        "readJSONBObject",
+                        JSONReader.class,
+                        Type.class,
+                        Object.class,
+                        long.class
+                );
+                return method.getDeclaringClass() != ObjectReader.class;
+            } catch (NoSuchMethodException e) {
+                return false;
+            }
+        });
     }
 
     public static Object readJSONBObject(
@@ -268,7 +275,7 @@ public class FieldReaderObject<T>
                         value = jsonReader.readAny();
                     }
                 } else {
-                    value = objectReader.readJSONBObject(jsonReader, fieldType, fieldName, features);
+                    value = readJSONBObject(objectReader, jsonReader, fieldType, fieldClass, fieldName, features);
                 }
             } else {
                 value = objectReader.readObject(jsonReader, fieldType, fieldName, features);
@@ -394,7 +401,7 @@ public class FieldReaderObject<T>
         }
 
         Object object = jsonReader.jsonb
-                ? initReader.readJSONBObject(jsonReader, fieldType, fieldName, features)
+                ? readJSONBObject(initReader, jsonReader, fieldType, fieldClass, fieldName, features)
                 : initReader.readObject(jsonReader, fieldType, fieldName, features);
 
         Function builder = initReader.getBuildFunction();
--- a/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
@@ -92,8 +92,16 @@ public class FieldReaderObject<T>
         }
 
         if (fieldClass != null && Map.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = jsonReader.getObjectReader(fieldType);
+            if (isCustomReader(fieldObjectReader)) {
+                return reader = fieldObjectReader;
+            }
             return reader = ObjectReaderImplMap.of(fieldType, fieldClass, features);
         } else if (fieldClass != null && Collection.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = jsonReader.getObjectReader(fieldType);
+            if (isCustomReader(fieldObjectReader)) {
+                return reader = fieldObjectReader;
+            }
             return reader = ObjectReaderImplList.of(fieldType, fieldClass, features);
         }
 
@@ -111,12 +119,27 @@ public class FieldReaderObject<T>
         }
 
         if (Map.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = context.getObjectReader(fieldType);
+            if (isCustomReader(fieldObjectReader)) {
+                return reader = fieldObjectReader;
+            }
             return reader = ObjectReaderImplMap.of(fieldType, fieldClass, features);
         } else if (Collection.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = context.getObjectReader(fieldType);
+            if (isCustomReader(fieldObjectReader)) {
+                return reader = fieldObjectReader;
+            }
             return reader = ObjectReaderImplList.of(fieldType, fieldClass, features);
         }
 
         return reader = context.getObjectReader(fieldType);
+    }
+
+    private static boolean isCustomReader(ObjectReader objectReader) {
+        return objectReader != null
+                && !(objectReader instanceof ObjectReaderImplMap)
+                && !(objectReader instanceof ObjectReaderImplMapTyped)
+                && !(objectReader instanceof ObjectReaderImplList);
     }
 
     @Override
--- a/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
@@ -135,11 +135,67 @@ public class FieldReaderObject<T>
         return reader = context.getObjectReader(fieldType);
     }
 
-    private static boolean isCustomReader(ObjectReader objectReader) {
+    static boolean isCustomReader(ObjectReader objectReader) {
         return objectReader != null
                 && !(objectReader instanceof ObjectReaderImplMap)
                 && !(objectReader instanceof ObjectReaderImplMapTyped)
-                && !(objectReader instanceof ObjectReaderImplList);
+                && !(objectReader instanceof ObjectReaderImplList)
+                && !(objectReader instanceof ObjectReaderImplListStr)
+                && !(objectReader instanceof ObjectReaderImplListInt64);
+    }
+
+    static boolean isReadJSONBObjectSupported(ObjectReader objectReader) {
+        try {
+            Method method = objectReader.getClass().getMethod(
+                    "readJSONBObject",
+                    JSONReader.class,
+                    Type.class,
+                    Object.class,
+                    long.class
+            );
+            return method.getDeclaringClass() != ObjectReader.class;
+        } catch (NoSuchMethodException e) {
+            return false;
+        }
+    }
+
+    public static Object readJSONBObject(
+            ObjectReader objectReader,
+            JSONReader jsonReader,
+            Type fieldType,
+            Object fieldName,
+            long features
+    ) {
+        return readJSONBObject(
+                objectReader,
+                jsonReader,
+                fieldType,
+                TypeUtils.getClass(fieldType),
+                fieldName,
+                features
+        );
+    }
+
+    public static Object readJSONBObject(
+            ObjectReader objectReader,
+            JSONReader jsonReader,
+            Type fieldType,
+            Class fieldClass,
+            Object fieldName,
+            long features
+    ) {
+        if (isCustomReader(objectReader)
+                && !isReadJSONBObjectSupported(objectReader)
+                && fieldClass != null
+        ) {
+            if (Map.class.isAssignableFrom(fieldClass)) {
+                objectReader = ObjectReaderImplMap.of(fieldType, fieldClass, features);
+            } else if (Collection.class.isAssignableFrom(fieldClass)) {
+                objectReader = ObjectReaderImplList.of(fieldType, fieldClass, features);
+            }
+        }
+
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
```diff
@@ -111,14 +119,29 @@ public ObjectReader getObjectReader(JSONReader.Context context) {
...
             return reader = ObjectReaderImplList.of(fieldType, fieldClass, features);
         }
 
         return reader = context.getObjectReader(fieldType);
     }
 
+    private static boolean isCustomReader(ObjectReader objectReader) {
+        return objectReader != null
+                && !(objectReader instanceof ObjectReaderImplMap)
+                && !(objectReader instanceof ObjectReaderImplMapTyped)
```
wenshao, line 145: **[Critical]** JSONB deserialization will crash for custom readers that only implement `readObject()`.

When `isCustomReader` returns true, the custom reader is cached as `this.reader`. Later, `readFieldValueJSONB` calls `initReader.readJSONBObject(...)`. Custom `ObjectReader` implementations (like the `EntityReader` in the test) that only override `readObject()` will fall through to the default interface method `readJSONBObject()`, which calls `createInstance(features)` — and the default `createInstance` throws `UnsupportedOperationException`.

Before this PR, the built-in `ObjectReaderImplMap` handled JSONB deserialization correctly. After this PR, the JSONB path regresses for any custom reader that doesn't also implement `readJSONBObject()`.

**Suggested fix:** In `readFieldValueJSONB`, check whether the resolved reader properly supports JSONB (e.g., by checking if it overrides `readJSONBObject`). If not, fall back to `ObjectReaderImplMap.of(fieldType, fieldClass, features)` for Map subtypes, or document that custom deserializers for Map/Collection subtypes MUST implement `readJSONBObject`.

core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
```diff
@@ -92,8 +92,16 @@ public ObjectReader getObjectReader(JSONReader jsonReader) {
         }
 
         if (fieldClass != null && Map.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = jsonReader.getObjectReader(fieldType);
+            if (isCustomReader(fieldObjectReader)) {
+                return reader = fieldObjectReader;
+            }
             return reader = ObjectReaderImplMap.of(fieldType, fieldClass, features);
         } else if (fieldClass != null && Collection.class.isAssignableFrom(fieldClass)) {
+            ObjectReader fieldObjectReader = jsonReader.getObjectReader(fieldType);
```
wenshao, line 105: **[Critical]** `FieldReaderList.readFieldValue()` bypasses the custom reader for Collection subtypes.

`FieldReaderList` (which extends `FieldReaderObject`) overrides `readFieldValue()` entirely. When JSON input is an array `[...]`, it calls `createList(context)` and manually parses array elements — it never calls `getObjectReader()` to resolve the custom reader. This means:

1. If the custom reader doesn't implement `createInstance(long)`, `createList()` throws `UnsupportedOperationException`
2. If it does, the custom deserializer is silently bypassed and built-in array parsing is used instead

The Collection-side fix in this PR is therefore non-functional for the most common case (a Collection subtype field receiving array JSON). The Map-side fix works because `FieldReaderMap.readFieldValue()` delegates to `super.readFieldValue()`.

**Suggested fix:** In `FieldReaderList.readFieldValue()`, after resolving the reader, check if it's a custom reader (not `ObjectReaderImplList`). If so, delegate directly:
```java
ObjectReader objectReader = getObjectReader(context);
if (!(objectReader instanceof ObjectReaderImplList)) {
    Object value = jsonReader.jsonb
            ? objectReader.readJSONBObject(jsonReader, fieldType, fieldName, features)
            : objectReader.readObject(jsonReader, fieldType, fieldName, features);
    accept(object, value);
    return;
}
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7636.java
```diff
@@ -0,0 +1,80 @@
...
+    public void testCustomReaderForMapSubtype() {
+        Entity entity = JSON.parseObject("{\"id\":123}", Entity.class);
+
+        assertEquals(123, entity.id);
+        assertEquals("custom", entity.get("reader"));
+    }
+
+    @Test
+    public void testMapSubtypeWithoutCustomReader() {
+        RawBean bean = JSON.parseObject("{\"entity\":{\"id\":123}}", RawBean.class);
```
wenshao, line 35: **[Suggestion]** No test coverage for Collection subtypes with custom deserializers.

The PR modifies the Collection branch in both `FieldReaderObject` and `ObjectReaderBaseModule`, but all three test methods only exercise Map subtypes (`Entity extends HashMap<String, Object>`). The Collection-side fix is entirely untested.

**Suggested fix:** Add analogous tests for a Collection subtype:
```java
@JSONType(deserializer = MyListReader.class)
public static class MyList extends ArrayList<String> {
    public int count;
}
```
With tests for: field-level deserialization, direct parse, and a raw Collection subtype without custom reader (regression test).

wenshao (changes requested):

hutiefang76:
Addressed the requested changes in 036911538.

What changed:
- avoid treating fastjson2 built-in List readers as custom readers
- fall back to the built-in Map/List readers for JSONB when a custom Map/Collection subtype reader does not implement readJSONBObject
- avoid the ASM List fast path for Collection subtypes with @JSONType(deserializer = ...), so field-level custom readers are honored
- added coverage for JSONB Map subtype fallback, Collection subtype custom readers, direct Collection subtype parsing, and raw Collection subtype regression

Verification:
- ./mvnw -pl core -Dtest=Issue7636 test -DskipITs -Dcheckstyle.skip -Drat.skip=true
- git diff --check

core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
```diff
@@ -111,14 +119,85 @@ public ObjectReader getObjectReader(JSONReader.Context context) {
...
+                    Object.class,
+                    long.class
+            );
+            return method.getDeclaringClass() != ObjectReader.class;
+        } catch (NoSuchMethodException e) {
+            return false;
+        }
+    }
+
+    public static Object readJSONBObject(
```
wenshao, line 169: **[Critical]** Multiple `readFieldValue` JSONB branches bypass this static `readJSONBObject` helper

This static helper correctly falls back to `ObjectReaderImplMap`/`ObjectReaderImplList` when a custom reader doesn't override `readJSONBObject`. However, several other JSONB code paths still call `objectReader.readJSONBObject(...)` directly on the interface, producing incorrect results or exceptions for Map/Collection subtypes with custom deserializers that only implement `readObject()`:

1. **`readFieldValue(JSONReader, T)`** (line ~270) — calls `objectReader.readJSONBObject(...)` directly
2. **`readFieldValue(JSONReader)`** (1-arg overload, line ~397) — same direct call
3. **`FieldReaderMapReadOnly.readFieldValue`** (line ~119) — same direct call
4. **`FieldReaderCollectionReadOnly.readFieldValue`** (line ~121) — same direct call

`ObjectReaderAdapter.readJSONBObject()` calls `fieldReader.readFieldValue(jsonReader, object)` on the JSONB path, making location (1) directly reachable. Locations (3) and (4) affect read-only Map/Collection fields.

**Suggested fix:** Route all JSONB branches through this static helper:
```java
// e.g., in readFieldValue(JSONReader, T) around line 270:
} else {
    value = readJSONBObject(objectReader, jsonReader, fieldType, fieldClass, fieldName, features);
}
```

core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
```diff
@@ -111,14 +119,85 @@ public ObjectReader getObjectReader(JSONReader.Context context) {
...
             return reader = ObjectReaderImplList.of(fieldType, fieldClass, features);
         }
 
         return reader = context.getObjectReader(fieldType);
     }
 
+    static boolean isCustomReader(ObjectReader objectReader) {
+        return objectReader != null
+                && !(objectReader instanceof ObjectReaderImplMap)
+                && !(objectReader instanceof ObjectReaderImplMapTyped)
```
wenshao, line 145: **[Critical]** `isCustomReader()` does not exclude `ObjectReaderImplMapMultiValueType`

`ObjectReaderImplMapMultiValueType` implements `ObjectReader` directly (not extending any of the five checked types). It would be misclassified as "custom," causing two problems:

1. The JSONB fallback in `readJSONBObject()` would replace it with a plain `ObjectReaderImplMap.of()`, losing multi-value semantics
2. `FieldReaderList.readFieldValue()` would take the custom-reader shortcut instead of the standard list-iteration path

```suggestion
    static boolean isCustomReader(ObjectReader objectReader) {
        return objectReader != null
                && !(objectReader instanceof ObjectReaderImplMap)
                && !(objectReader instanceof ObjectReaderImplMapTyped)
                && !(objectReader instanceof ObjectReaderImplMapMultiValueType)
                && !(objectReader instanceof ObjectReaderImplList)
                && !(objectReader instanceof ObjectReaderImplListStr)
                && !(objectReader instanceof ObjectReaderImplListInt64);
    }
```

core/src/main/java/com/alibaba/fastjson2/reader/FieldReaderObject.java
```diff
@@ -111,14 +119,85 @@ public ObjectReader getObjectReader(JSONReader.Context context) {
...
+                && !(objectReader instanceof ObjectReaderImplListStr)
+                && !(objectReader instanceof ObjectReaderImplListInt64);
+    }
+
+    static boolean isReadJSONBObjectSupported(ObjectReader objectReader) {
+        try {
+            Method method = objectReader.getClass().getMethod(
+                    "readJSONBObject",
+                    JSONReader.class,
+                    Type.class,
```
wenshao: **[Suggestion]** Uncached reflection on the JSONB hot path

`isReadJSONBObjectSupported` calls `objectReader.getClass().getMethod(...)` on every invocation — one of the most expensive JDK reflection operations. Additionally, when the fallback triggers, `ObjectReaderImplMap.of(...)` / `ObjectReaderImplList.of(...)` allocates new reader objects each time.

Both results are deterministic per `ObjectReader` class. Cache them:

```java
private static final ConcurrentHashMap<Class<?>, Boolean> JSONB_SUPPORTED_CACHE = new ConcurrentHashMap<>();

static boolean isReadJSONBObjectSupported(ObjectReader objectReader) {
    return JSONB_SUPPORTED_CACHE.computeIfAbsent(objectReader.getClass(), clazz -> {
        try {
            Method method = clazz.getMethod("readJSONBObject",
                    JSONReader.class, Type.class, Object.class, long.class);
            return method.getDeclaringClass() != ObjectReader.class;
        } catch (NoSuchMethodException e) {
            return false;
        }
    });
}
```

Similarly, consider caching the fallback reader created in `readJSONBObject()` rather than calling `ObjectReaderImplMap.of()` / `ObjectReaderImplList.of()` on every invocation.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
```diff
@@ -3560,10 +3562,24 @@ private void genReadObject(
         gwGetFieldType(classNameType, mw, i, fieldType);
         mw.visitLdcInsn(fieldReader.fieldName);
         mw.visitLdcInsn(fieldFeatures);
-        mw.invokeinterface(
-                TYPE_OBJECT_READER,
-                jsonb ? "readJSONBObject" : "readObject",
-                METHOD_DESC_READ_OBJECT);
+        if (jsonb && (Map.class.isAssignableFrom(fieldReader.fieldClass)
```
wenshao: **[Suggestion]** ASM bytecode redirect applies to ALL Map/Collection JSONB fields, not just custom ones

The condition routes every Map/Collection JSONB field through the static `FieldReaderObject.readJSONBObject` helper, adding unnecessary overhead (`isCustomReader` check + potential `getMethod` reflection) for the vast majority of fields that have no custom deserializer.

Gate it with `hasCustomDeserializer` to match the JSON-side pattern at line 3211:

```suggestion
        if (jsonb && hasCustomDeserializer(fieldReader.fieldClass)
                && (Map.class.isAssignableFrom(fieldReader.fieldClass)
                || Collection.class.isAssignableFrom(fieldReader.fieldClass))) {
```

This restricts the static helper indirection to fields that actually need it, eliminating the reflection overhead for standard fields.

wenshao (changes requested):

hutiefang76:
Follow-up pushed in `c7487d489`.

What changed:
- route the remaining JSONB field-read paths through `FieldReaderObject.readJSONBObject(...)` instead of calling `ObjectReader.readJSONBObject(...)` directly
  - `FieldReaderObject.readFieldValue(JSONReader, T)`
  - `FieldReaderObject.readFieldValue(JSONReader)`
  - `FieldReaderMapReadOnly.readFieldValue(...)`
  - `FieldReaderCollectionReadOnly.readFieldValue(...)`
  - fallback branch in `FieldReaderList.readFieldValue(...)`
- exclude `ObjectReaderImplMapMultiValueType` from `isCustomReader(...)`
- cache the `readJSONBObject` override check per ObjectReader class
- gate the ASM JSONB helper indirection with `hasCustomDeserializer(fieldReader.fieldClass)`
- add JSONB coverage for Collection subtype fields and read-only Map/List subtype fields
- add a package-level regression test for the built-in `ObjectReaderImplMapMultiValueType` classification

Validation:
```bash
git diff --check
./mvnw -pl core -Dtest=FieldReaderObjectTest,Issue7636,MapMultiValueTypeTest test -DskipITs -Dcheckstyle.skip -Drat.skip=true
./mvnw -pl core test -DskipITs -Dcheckstyle.skip -Drat.skip=true
```

The full `core` module run passed 7894 tests with 0 failures/errors.

hutiefang76:
Thanks for the detailed review. After checking the remaining JSONB/custom-reader comments again, this patch has grown beyond the narrow Map subtype issue into broader reader-dispatch semantics. I don't want to keep expanding a risky reader hot path in this PR, so I'm closing this attempt instead of pushing more speculative changes. If there is a narrower direction you prefer for #7636, I can revisit it with a smaller patch and focused tests.

PR #7665, opened 2026-06-22
fix: support JSONCompiled records

### LINKED ISSUE #7659: [BUG] @JSONCompiled fails compilation on Java records
## Summary

Using `@JSONCompiled` on a Java record fails compilation with FastJSON2 codegen.
The generated reader path appears to assign values directly into record component
fields, but record fields are final.

## Minimal Reproduction

`ReproJsonCompiledOnly.java`

```java
import com.alibaba.fastjson2.annotation.JSONCompiled;

public class ReproJsonCompiledOnly {
    @JSONCompiled
    public record Person(String name) {
    }
}
```
## Actual Result

```text
warning: Supported source version 'RELEASE_8' from annotation processor 'com.alibaba.fastjson2.internal.processor.JSONBaseAnnotationProcessor' less than -source '21'
warning: Supported source version 'RELEASE_8' from annotation processor 'com.alibaba.fastjson2.internal.processor.JSONCompiledAnnotationProcessor' less than -source '21'
ReproJsonCompiledOnly.java:5: error: cannot assign a value to final variable name
    public record Person(String name) {
           ^
Note: ReproJsonCompiledOnly.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
2 warnings
```

## Expected Result
`@JSONCompiled` supports records by generating code that calls the canonical constructor.

## Environment

- JDK: `javac 21.0.5`
- FastJSON runtime: `com.alibaba.fastjson2:fastjson2:2.0.62`
- FastJSON codegen: `com.alibaba.fastjson2:fastjson2-codegen:2.0.62`

hutiefang76:
### What changed

This fixes `@JSONCompiled` on Java records. The generated reader path used to create a default instance and assign field values afterwards, which does not work for record components because the backing fields are final.

For records, the generated reader now collects component values into local variables and returns a new record instance through the canonical constructor. The normal bean path is unchanged.

### Details

- keep record reader attributes in declaration order so constructor arguments match record component order
- avoid generating field-writer lambdas that assign into final record fields
- skip the no-arg `createInstance()` override for records
- generate record-specific `readObject` / `readJSONBObject` methods that construct the record after field parsing
- reuse primitive-aware type generation for generated field class references

### Tests

```bash
./mvnw -s /tmp/fastjson2-empty-settings.xml -Penable-codegen -pl codegen -am -Dtest=JSONCompiledRecordTest -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -s /tmp/fastjson2-empty-settings.xml -pl core -DskipTests install
./mvnw -s /tmp/fastjson2-empty-settings.xml -N install
cd codegen && ../mvnw -s /tmp/fastjson2-empty-settings.xml test
```

The regression test compiles a `@JSONCompiled` record with `javac`, then loads it and verifies `JSON.parseObject` can construct the record values.

Fixes #7659

### DIFF
--- a/codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
+++ b/codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
@@ -665,6 +665,7 @@ public class JSONCompiledAnnotationProcessor
         if (genericStart != -1) {
             fieldClass = fieldClass.substring(0, genericStart);
         }
+        JCTree.JCExpression fieldTypeRef = genTypeReference(fieldType);
         JCTree.JCExpression fieldClassRef = field(getFieldValueType(fieldClass), names._class);
         JCTree.JCFieldAccess beanClassRef = field(beanType, names._class);
         return method(
@@ -676,13 +677,48 @@ public class JSONCompiledAnnotationProcessor
                 literal(0),
                 literal(0L),
                 defNull(),
-                fieldClassRef,
+                fieldTypeRef,
                 fieldClassRef,
                 literal(attributeInfo.name),
                 beanClassRef,
                 defNull(),
                 defNull()
         );
+    }
+
+    private JCTree.JCExpression genTypeReference(String type) {
+        int genericStart = type.indexOf('<');
+        if (genericStart == -1) {
+            return field(getFieldValueType(type), names._class);
+        }
+
+        String rawType = type.substring(0, genericStart);
+        String typeArguments = type.substring(genericStart + 1, type.length() - 1);
+        ListBuffer<JCTree.JCExpression> args = new ListBuffer<>();
+        args.append(field(getFieldValueType(rawType), names._class));
+        for (String argument : splitTypeArguments(typeArguments)) {
+            args.append(genTypeReference(argument.trim()));
+        }
+        return method(qualIdent("com.alibaba.fastjson2.TypeReference"), "parametricType", args.toList());
+    }
+
+    private java.util.List<String> splitTypeArguments(String typeArguments) {
+        java.util.List<String> result = new ArrayList<>();
+        int depth = 0;
+        int start = 0;
+        for (int i = 0; i < typeArguments.length(); i++) {
+            char ch = typeArguments.charAt(i);
+            if (ch == '<') {
+                depth++;
+            } else if (ch == '>') {
+                depth--;
+            } else if (ch == ',' && depth == 0) {
+                result.add(typeArguments.substring(start, i));
+                start = i + 1;
+            }
+        }
+        result.add(typeArguments.substring(start));
+        return result;
     }
 
     private JCTree.JCMethodInvocation genFieldReaderList(
@@ -3138,6 +3174,7 @@ public class JSONCompiledAnnotationProcessor
             if (referenceDetect) {
                 referenceDetect = isReference(itemType);
             }
+            JCTree.JCLabeledStatement itemLoopLabel = referenceDetect ? label("_list" + i) : null;
 
             JCTree.JCVariableDecl for_iVar;
             if ("i".equals(attributeInfo.name)) {
@@ -3183,7 +3220,7 @@ public class JSONCompiledAnnotationProcessor
                 );
                 isReferenceStmts.append(exec(addResolveTaskMethod));
                 isReferenceStmts.append(exec(method(fieldValueVar, "add", defNull())));
-                isReferenceStmts.append(defContinue(loopLabel));
+                isReferenceStmts.append(defContinue(itemLoopLabel));
                 whileStmts.append(defIf(method(jsonReaderIdent, "isReference"), block(isReferenceStmts.toList())));
                 whileStmts.append(listItemVar);
                 item = ident(listItemVar);
@@ -3192,25 +3229,32 @@ public class JSONCompiledAnnotationProcessor
             whileStmts.append(exec(method(fieldValueVar, "add", item)));
 
             ListBuffer<JCTree.JCStatement> condStmts = new ListBuffer<>();
+            JCTree.JCStatement itemLoop;
             if (isJsonb) {
                 JCTree.JCVariableDecl itemCntVar = defVar(attributeInfo.name + "_item_cnt", TypeTag.INT, method(jsonReaderIdent, "startArray"));
-                condStmts.append(forLoop(
+                itemLoop = forLoop(
                         List.of(for_iVar, itemCntVar),
                         lt(ident(for_iVar), ident(itemCntVar)),
                         List.of(exec(unary(JCTree.Tag.PREINC, ident(for_iVar)))),
-                        block(whileStmts.toList())));
+                        block(whileStmts.toList()));
             } else if (referenceDetect) {
-                condStmts.append(forLoop(
+                itemLoop = forLoop(
                         List.of(for_iVar),
                         unary(JCTree.Tag.NOT, nextIfArrayEndMethod),
                         List.of(exec(unary(JCTree.Tag.PREINC, ident(for_iVar)))),
-                        block(whileStmts.toList())));
+                        block(whileStmts.toList()));
             } else {
-                condStmts.append(whileLoop(unary(JCTree.Tag.NOT, nextIfArrayEndMethod), block(whileStmts.toList())));
+                itemLoop = whileLoop(unary(JCTree.Tag.NOT, nextIfArrayEndMethod), block(whileStmts.toList()));
+            }
+            if (itemLoopLabel != null) {
+                itemLoopLabel.body = itemLoop;
+                condStmts.append(itemLoopLabel);
+            } else {
+                condStmts.append(itemLoop);
             }
 
             if (isJsonb) {
-                stmts.appendList(condStmts.toList());
+                stmts.append(defIf(ne(fieldValueVar, defNull()), block(condStmts.toList())));
             } else {
                 stmts.append(defIf(nextIfArrayStartMethod, block(condStmts.toList())));
             }
--- a/codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
+++ b/codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
@@ -138,9 +138,10 @@ public class JSONCompiledAnnotationProcessor
                         innerReadClass.defs = innerReadClass.defs.append(genReadConstructor(beanType, beanNew, attributeInfos, readSuperClass, generatedFields, record));
 
                         // generate createInstance
-                        if (!record) {
-                            innerReadClass.defs = innerReadClass.defs.append(genCreateInstance(objectType, beanNew));
-                        }
+                        innerReadClass.defs = innerReadClass.defs.append(
+                                record
+                                        ? genCreateRecordInstance(objectType, beanType, attributeInfos)
+                                        : genCreateInstance(objectType, beanNew));
 
                         // generate readObject
                         innerReadClass.defs = innerReadClass.defs.append(
@@ -311,7 +312,7 @@ public class JSONCompiledAnnotationProcessor
         int fieldsSize = fields.size();
         for (int i = 0; i < fieldsSize; ++i) {
             JCTree.JCMethodInvocation readerMethod = record
-                    ? genRecordFieldReader(fields.get(i), objectReadersType)
+                    ? genRecordFieldReader(beanType, fields.get(i))
                     : genFieldReader(beanType, fields.get(i), objectReadersType);
             if (readerMethod != null) {
                 fieldReaders.append(readerMethod);
@@ -655,8 +656,8 @@ public class JSONCompiledAnnotationProcessor
     }
 
     private JCTree.JCMethodInvocation genRecordFieldReader(
-            AttributeInfo attributeInfo,
-            JCTree.JCExpression objectReadersType
+            JCTree.JCExpression beanType,
+            AttributeInfo attributeInfo
     ) {
         String fieldType = attributeInfo.type.toString();
         String fieldClass = fieldType;
@@ -664,11 +665,23 @@ public class JSONCompiledAnnotationProcessor
         if (genericStart != -1) {
             fieldClass = fieldClass.substring(0, genericStart);
         }
+        JCTree.JCExpression fieldClassRef = field(getFieldValueType(fieldClass), names._class);
+        JCTree.JCFieldAccess beanClassRef = field(beanType, names._class);
         return method(
-                objectReadersType,
-                "fieldReader",
+                field(qualIdent("com.alibaba.fastjson2.reader.ObjectReaderCreator"), "INSTANCE"),
+                "createFieldReaderParam",
+                beanClassRef,
+                beanClassRef,
                 literal(attributeInfo.name),
-                field(getFieldValueType(fieldClass), names._class)
+                literal(0),
+                literal(0L),
+                defNull(),
+                fieldClassRef,
+                fieldClassRef,
+                literal(attributeInfo.name),
+                beanClassRef,
+                defNull(),
+                defNull()
         );
     }
 
@@ -797,6 +810,25 @@ public class JSONCompiledAnnotationProcessor
     private JCTree.JCMethodDecl genCreateInstance(JCTree.JCIdent objectType, JCTree.JCNewClass beanNew) {
         JCTree.JCVariableDecl featuresVar = defVar("features", TypeTag.LONG);
         return defMethod(Flags.PUBLIC, "createInstance", objectType, List.of(featuresVar), block(defReturn(beanNew)));
+    }
+
+    private JCTree.JCMethodDecl genCreateRecordInstance(
+            JCTree.JCIdent objectType,
+            JCTree.JCExpression beanType,
+            java.util.List<AttributeInfo> attributeInfos
+    ) {
+        JCTree.JCVariableDecl featuresVar = defVar("features", TypeTag.LONG);
+        ListBuffer<JCTree.JCExpression> args = new ListBuffer<>();
+        for (AttributeInfo attributeInfo : attributeInfos) {
+            args.append(defaultValue(attributeInfo.type.toString()));
+        }
+        return defMethod(
+                Flags.PUBLIC,
+                "createInstance",
+                objectType,
+                List.of(featuresVar),
+                block(defReturn(newClass(null, null, beanType, args.toList(), null)))
+        );
     }
 
     private JCTree.JCMethodDecl genReadObject(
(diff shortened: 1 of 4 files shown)


codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -134,18 +135,24 @@ public void visitClassDef(JCTree.JCClassDecl beanClassDecl) {
                         }
 
                         // generate constructor
-                        innerReadClass.defs = innerReadClass.defs.append(genReadConstructor(beanType, beanNew, attributeInfos, readSuperClass, generatedFields));
+                        innerReadClass.defs = innerReadClass.defs.append(genReadConstructor(beanType, beanNew, attributeInfos, readSuperClass, generatedFields, record));
 
                         // generate createInstance
-                        innerReadClass.defs = innerReadClass.defs.append(genCreateInstance(objectType, beanNew));
+                        if (!record) {
+                            innerReadClass.defs = innerReadClass.defs.append(genCreateInstance(objectType, beanNew));
```
wenshao: **[Critical]** Skipping `genCreateInstance` for records leaves `creator` as `null` (set at line 307 via `record ? defNull() : constructorRef(beanType)`). `ObjectReaderAdapter.createInstance0` throws `JSONException("create instance error, " + objectClass)` when `creator` is null.

Code paths that call `createInstance()` instead of `readObject()` — e.g., `JSON.parseObject("{}", Person.class)` with an empty JSON object, collection item factories, or `@JSONField(defaultValue)` resolution — will throw at runtime for records.

Consider either generating a `createInstance` that throws a more descriptive error, or generating one that invokes the canonical constructor with default values (nulls/zeros).

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -965,6 +993,182 @@ private JCTree.JCMethodDecl genReadObject(
...
+                        fieldValueVar,
+                        loopLabel,
+                        stmts,
+                        i,
+                        structInfo.referenceDetect && isReference(type),
+                        fieldReaderField,
+                        beanType,
+                        isJsonb);
+            } else if (type.startsWith("java.util.Map<java.lang.String,")) {
+                valueExpr = genFieldValueMap(type, attributeInfo, jsonReaderIdent, fieldValueVar, loopLabel, stmts, i, false, isJsonb);
```
wenshao: **[Critical]** Two reference-detection issues in the record read path:

1. The `genFieldValueMap` call hardcodes `false` for the `referenceDetect` parameter. The non-record `genReadFieldValue` passes `structInfo.referenceDetect && isReference(type)`.

2. The entire `genReadRecordFieldValue` method is missing the top-level `$ref` detection block present in `genReadFieldValue` (which checks `jsonReader.isReference()`, reads the ref string, and calls `fieldReader.addResolveTask(...)`).

**Impact:** If JSON contains `{"$ref":"..."}` markers for record fields (of any object/collection type), the generated code will attempt to deserialize them as regular objects rather than resolving references. This can cause incorrect deserialization or infinite loops on circular references.

```suggestion
            } else if (type.startsWith("java.util.Map<java.lang.String,")) {
                valueExpr = genFieldValueMap(type, attributeInfo, jsonReaderIdent, fieldValueVar, loopLabel, stmts, i, structInfo.referenceDetect && isReference(type), isJsonb);
```

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -965,6 +993,182 @@ private JCTree.JCMethodDecl genReadObject(
...
+                            jsonReaderIdent,
+                            i,
+                            structInfo,
+                            loopLabel,
+                            beanType,
+                            isJsonb))
+            ));
+        }
+
+        loopBody.append(exec(method(jsonReaderIdent, "skipValue")));
```
wenshao, line 1124: **[Suggestion]** The record read path falls through directly to `skipValue()` for unrecognized field names. The non-record `genReadObject` has two fallback mechanisms before skipping:

1. `readFieldValueWithLCase` — case-insensitive field name matching (e.g., `{"Name":"John"}` matching a `name` field)
2. `processExtra` — respects features like `ErrorOnUnknownField`, `SupportAutoType`, etc.

Without these, records silently discard extra fields and won't match field names with different casing, even when `ErrorOnUnknownField` or `SupportSmartMatch` features are enabled. This is a behavioral gap vs. non-record classes.

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -644,6 +654,24 @@ private JCTree.JCMethodInvocation genFieldReader(
...
     }
 
+    private JCTree.JCMethodInvocation genRecordFieldReader(
+            AttributeInfo attributeInfo,
+            JCTree.JCExpression objectReadersType
+    ) {
+        String fieldType = attributeInfo.type.toString();
+        String fieldClass = fieldType;
+        int genericStart = fieldClass.indexOf('<');
+        if (genericStart != -1) {
```
wenshao, line 665: **[Suggestion]** `genRecordFieldReader` strips generic parameters (e.g., `List<String>` → `List`) and always calls the simple `ObjectReaders.fieldReader(name, class)` overload. The non-record `genFieldReader` preserves generic type info by using `fieldReaderWithMethod` or `fieldReaderWithField` for parameterized types.

For record fields with generic types like `Set<MyEnum>`, `Collection<MyPojo>`, or other non-`List`/non-`Map` generic types, the generated `FieldReader` will only see the raw type. The element type information is lost, potentially deserializing elements as `Object`/`LinkedHashMap` instead of the intended type.

Consider using the same generic-aware approach as `genFieldReader` for record fields.

codegen/src/test/java/com/alibaba/fastjson2/internal/processor/JSONCompiledRecordTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JSONCompiledRecordTest {
+    @TempDir
+    public Path tempDir;
+
+    @Test
+    public void record() throws Exception {
+        Path sourceDir = tempDir.resolve("src");
+        Path classesDir = tempDir.resolve("classes");
```
wenshao: **[Suggestion]** The single test only covers JSON deserialization of `record Person(int id, String name)`. The PR introduces several new code-generation branches that remain entirely untested:

- **JSONB deserialization** (`readJSONBObject` generated via `genReadRecordObject(..., isJsonb=true)`)
- **Serialization** (`JSON.toJSONString(person)` — the writer path for records)
- **`List<T>` fields** (the `java.util.List<` branch in `genReadRecordFieldValue`)
- **`Map<String, V>` fields** (the `java.util.Map<java.lang.String,` branch)
- **Other primitive types** (`defaultValue()` handles boolean, byte, short, long, char, float, double — only int is tested)
- **Null JSON** (`JSON.parseObject("null", personClass)`)
- **Nested object/record fields** (the generic ObjectReader fallback path)

Consider adding tests for at least JSONB round-trip, serialization, and collection/Map field types.

wenshao (changes requested):

hutiefang76:
Thanks for the detailed review. Updated the PR with a narrower follow-up:

- generate `createInstance(long features)` for records using default component values
- build record field readers through the constructor-parameter reader path instead of the simple bean-field helper
- pass `referenceDetect` through the record `Map<String, V>` branch
- fix JSONB list/map generated read paths that were still using text-JSON array/comma helpers
- expand `JSONCompiledRecordTest` for createInstance defaults, serialization, JSONB round-trip, and List/Map record components
- make the dynamic javac test helper JDK 8 source-compatible (`Path.of` / `readAllBytes` removed, record tests skip below JDK 17)

Local check:

```bash
./mvnw -Penable-codegen -pl codegen -Dtest=JSONCompiledRecordTest test -DskipITs -Dcheckstyle.skip -Drat.skip=true
git diff --check
```

I kept broader delayed `$ref` assignment for record object fields out of this commit because record components are constructor-bound/final; I can follow up separately if you want that handled with a dedicated design.

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -2957,7 +3209,11 @@ jsonReaderIdent, defNull(),
                 condStmts.append(whileLoop(unary(JCTree.Tag.NOT, nextIfArrayEndMethod), block(whileStmts.toList())));
             }
 
-            stmts.append(defIf(nextIfArrayStartMethod, block(condStmts.toList())));
+            if (isJsonb) {
+                stmts.appendList(condStmts.toList());
+            } else {
```
wenshao, line 3258: **[Critical]** JSONB `startArray()` has no null guard — stream corruption on null List fields

The JSONB branch appends the for-loop with `startArray()` **unconditionally** to `stmts` (`stmts.appendList(condStmts.toList())`). In the non-JSONB path, the loop is wrapped inside `defIf(nextIfArrayStartMethod, ...)` which acts as a null guard.

When the JSON value is null, `nextIfNull()` (line 3125) consumes the BC_NULL byte and sets `field=null`, but execution continues into the for-loop. `startArray()` then reads the **next** byte in the stream (belonging to the next field or object-end marker), causing stream corruption or `JSONException`.

This affects both record and non-record `@JSONCompiled` classes since `genFieldValueList` is a shared method.

```suggestion
            if (isJsonb) {
                JCTree.JCVariableDecl itemCntVar = defVar(attributeInfo.name + "_item_cnt", TypeTag.INT, method(jsonReaderIdent, "startArray"));
                ListBuffer<JCTree.JCStatement> forBlock = new ListBuffer<>();
                forBlock.append(forLoop(
                        List.of(for_iVar, itemCntVar),
                        lt(ident(for_iVar), ident(itemCntVar)),
                        List.of(exec(unary(JCTree.Tag.PREINC, ident(for_iVar)))),
                        block(whileStmts.toList())));
                stmts.append(defIf(nextIfArrayStartMethod, block(forBlock.toList())));
            } else {
                stmts.append(defIf(nextIfArrayStartMethod, block(condStmts.toList())));
            }
```

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -965,6 +1025,191 @@ private JCTree.JCMethodDecl genReadObject(
...
+        return defMethod(
+                Flags.PUBLIC,
+                isJsonb ? "readJSONBObject" : "readObject",
+                objectType,
+                List.of(jsonReaderVar, fieldTypeVar, fieldNameVar, featuresVar),
+                block(readObjectBody.toList())
+        );
+    }
+
+    private List<JCTree.JCStatement> genReadRecordFieldValue(
```
wenshao, line 1148: **[Critical]** Missing reference detection in record read path

`genReadRecordFieldValue` has no `$ref` / reference-detection guard. The non-record equivalent `genReadFieldValue` checks `structInfo.referenceDetect` and generates:
```java
if (jsonReader.isReference()) {
    String ref = jsonReader.readReference();
    fieldReader.addResolveTask(jsonReader, object, ref);
    continue;
}
```

This is entirely absent from the record path. Additionally, in `genFieldValueList`, the JSONB branch (`if (isJsonb)` at line 3195) is checked **before** `referenceDetect`, so when both are true, the JSONB count-based for-loop runs without any `isReference` check inside the loop body.

Records with circular or shared-object references will produce incorrect deserialization silently.

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -965,6 +1025,191 @@ private JCTree.JCMethodDecl genReadObject(
...
+                JCTree.JCMethodInvocation getObjectReaderMethod = method(fieldReaderField, "getObjectReader", jsonReaderIdent);
+                JCTree.JCAssign objectReaderAssign = assign(objectReaderIdent, getObjectReaderMethod);
+                stmts.append(defIf(eq(objectReaderIdent, defNull()), block(exec(objectReaderAssign))));
+                JCTree.JCMethodInvocation objectMethod = method(
+                        field(ident(names._this), fieldObjectReader(i)),
+                        isJsonb ? "readJSONBObject" : "readObject",
+                        jsonReaderIdent,
+                        field(fieldReaderField, "fieldType"),
+                        literal(attributeInfo.name),
+                        literal(0L)
```
wenshao, line 1206: **[Suggestion]** Dead `features2Var` — reader features never propagated to record path

`features2Var` is computed (`features | this.features`) in `genReadRecordObject` and appended to the generated method body, but is never referenced afterward. Here `literal(0L)` is used instead of `ident(features2Var)` for the fallback ObjectReader call. All reader features (`SupportAutoType`, `ErrorOnUnknownProperties`, `SupportSmartMatch`, etc.) are silently ignored for record deserialization.

Either wire `features2Var` through (replace `literal(0L)` with `ident(features2Var)` and add `processExtra`/`readFieldValueWithLCase` calls to the dispatch loop), or remove the dead variable from `genReadRecordObject`.

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -965,6 +1025,191 @@ private JCTree.JCMethodDecl genReadObject(
...
+        JCTree.JCLabeledStatement loopLabel = label("_while");
+        JCTree.JCWhileLoop loopHead = whileLoop(unary(JCTree.Tag.NOT, method(jsonReaderIdent, "nextIfObjectEnd")), null);
+        ListBuffer<JCTree.JCStatement> loopBody = new ListBuffer<>();
+
+        JCTree.JCFieldAccess readFieldNameHashCode = field(jsonReaderIdent, "readFieldNameHashCode");
+        JCTree.JCVariableDecl hashCode64Var = defVar("hashCode64", TypeTag.LONG, method(readFieldNameHashCode));
+        JCTree.JCExpression hashCode64 = ident(hashCode64Var);
+        loopBody.append(hashCode64Var);
+
+        for (int i = 0; i < attributeInfos.size(); ++i) {
```
wenshao, line 1109: **[Suggestion]** Missing `genRead243` fast-path and switch dispatch for records

Two performance gaps vs. the non-record path:

1. **genRead243**: The non-record `genReadObject` calls `genRead243` for field names with byte-length in [2,43], using direct byte comparison (`nextIfName4Match*`) instead of `readFieldNameHashCode` + linear if-chain. Since most Java record component names are 2-43 characters, this is the hottest path.

2. **Switch dispatch**: For >6 fields, the non-record path uses switch-based dispatch on `hashCode32` (O(1) JVM `tableswitch`/`lookupswitch`). Records always use a linear if-chain, degrading to O(N) per JSON field.

codegen/src/test/java/com/alibaba/fastjson2/internal/processor/JSONCompiledRecordTest.java
```diff
@@ -0,0 +1,185 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class JSONCompiledRecordTest {
+    @TempDir
+    public Path tempDir;
+
+    @Test
+    public void record() throws Exception {
```
wenshao, line 32: **[Suggestion]** Test coverage gaps for critical code paths

The 3 existing tests cover basic JSON parse, `createInstance` defaults, and JSONB round-trip with collections. Missing scenarios directly relevant to the bugs found:

- **Null List field** (triggers the JSONB `startArray()` stream corruption)
- **Partial JSON** (missing record components — exercises `defaultValue` initialization)
- **Extra/unknown fields** (exercises the `skipValue` path)
- **Nested records / custom object fields** (exercises ObjectReader fallback)
- **Reference detection** (`@JSONType(referenceDetect=true)`)

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
```diff
@@ -2947,7 +3192,14 @@ jsonReaderIdent, defNull(),
             whileStmts.append(exec(method(fieldValueVar, "add", item)));
 
             ListBuffer<JCTree.JCStatement> condStmts = new ListBuffer<>();
-            if (referenceDetect) {
+            if (isJsonb) {
```
wenshao, line 3233: **[Suggestion]** JSONB changes to shared methods affect non-record classes (undocumented blast radius)

The `isJsonb` branching in `genFieldValueList` (this line) and `genFieldValueMap` (line 3212) applies to both `genReadFieldValue` (non-record) and `genReadRecordFieldValue` (record). These are correctness fixes (JSONB `nextIfArrayEnd`/`nextIfComma` throw `UnsupportedOperationException`), but no non-record JSONB test exists to prevent regression. Consider adding a non-record `@JSONCompiled` test with `List` and `Map` fields.

wenshao (changes requested):

hutiefang76:
Addressed the Reflect JDK 17 CI failure in `9b10e4ab5`.

Root cause:
- record component accessors such as `tags()` were not registered as getters during analysis
- generated writer code could fall back to record fields and fail first-round javac compilation for generic record components under JDK 17 reflect CI

What changed:
- recognize zero-arg record component accessor methods when the matching record field attribute already exists
- this lets generated writers use the accessor method path for record components

Validation:
```
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Penable-codegen -pl codegen -Dfastjson2.creator=reflect -Dtest=JSONCompiledRecordTest#recordJsonbSerializationAndCollections clean test -DskipITs -Dcheckstyle.skip -Drat.skip=true
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Penable-codegen -pl codegen -Dfastjson2.creator=reflect -Dtest=JSONCompiledRecordTest test -DskipITs -Dcheckstyle.skip -Drat.skip=true
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Penable-codegen -pl codegen -Dtest=JSONCompiledRecordTest test -DskipITs -Dcheckstyle.skip -Drat.skip=true
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Penable-codegen -pl codegen -Dfastjson2.creator=reflect test -DskipITs -Dcheckstyle.skip -Drat.skip=true
git diff --check
```

The final codegen reflect run passed 8 tests with 0 failures/errors.

wenshao (approved):
LGTM overall. The core approach (local variables + canonical constructor for records) is sound. The existing inline comments cover the substantive issues. One minor observation: `getReaderAttributes()` for records relies on `getEnclosedElements()` insertion order rather than `getRecordComponents()` for canonical constructor argument ordering — works in practice with javac but could be made more robust.

codegen/src/main/java/com/alibaba/fastjson2/internal/processor/JSONCompiledAnnotationProcessor.java
artpaym: @hutiefang76 please confirm if this was handled. Considering fastjson2 is about performance, would be nice if there is no degradation in records vs classes.

PR #7666, opened 2026-06-23
fix: preserve runtime subtype fields in JSON.toJSON, for issue #7655

### LINKED ISSUE #7655: [BUG] com.alibaba.fastjson.JSON.toJSON 方法在属性声明为父类时丢失子类属性
### 问题描述
在 fastjson2 的 com.alibaba:fastjson 兼容包中，使用 com.alibaba.fastjson.JSON.toJSON(Object) 方法将一个复合对象转换为 JSONObject 时，如果某个属性的声明类型（编译期类型）是父类，但实际运行期类型是子类，那么子类特有的属性会在转换过程中丢失。

该问题似乎是因为兼容层的序列化逻辑严格依据了字段的声明类型（Compile-time Type）来解析 Bean 属性，而没有动态识别运行期的实际多态类型（Runtime Type）。

### 环境信息
 - OS信息：  Any
 - JDK信息： JDK21
 - 版本信息：>= 2.0.57 

### 重现步骤
```java
package com.example.demo;

import com.alibaba.fastjson.JSONObject;

public class Fastjson2IssueTest {

        // 1. 创建子类 ClassB 对象并赋值
        ClassB classB = new ClassB();
        classB.setField1("value1");
        classB.setField2("value2");
        classB.setField3("value3");
        classB.setField4("value4");

        // 2. 创建 ClassC 对象（其内部属性声明为父类 ClassA）
        ClassC classC = new ClassC();
        classC.setFieldClassA(classB); // 将子类实例赋给父类声明的字段

        // 3. 通过 fastjson 1.x 兼容接口转换为 JSONObject
        JSONObject json1 = (JSONObject) JSON.toJSON(classC);

        // 4. 输出结果
        System.out.println("转换后的 JSON 字符串: " + json1.toJSONString());

        // 5. 验证子类属性是否存在
        JSONObject fieldClassAJson = json1.getJSONObject("fieldClassA");
        if (!fieldClassAJson.containsKey("field3") || !fieldClassAJson.containsKey("field4")) {
            System.err.println("❌ BUG 触发: 子类特有属性 'field3' 或 'field4' 丢失了！");
        } else {
            System.out.println("✅ 验证通过: 属性完整。");
        }
    }

    // --- 测试实体类 ---
    public static class ClassA {
        private String field1;
        private String field2;

        public String getField1() { return field1; }
        public void setField1(String field1) { this.field1 = field1; }
        public String getField2() { return field2; }
        public void setField2(String field2) { this.field2 = field2; }
    }

    public static class ClassB extends ClassA {
        private String field3;
        private String field4;

        public String getField3() { return field3; }
        public void setField3(String field3) { this.field3 = field3; }
        public String getField4() { return field4; }
        public void setField4(String field4) { this.field4 = field4; }
    }

    public static class ClassC {
        // 声明类型为父类 ClassA
        private ClassA fieldClassA;

        public ClassA getFieldClassA() { return fieldClassA; }
        public void setFieldClassA(ClassA fieldClassA) { this.fieldClassA = fieldClassA; }
    }
}
```
### 期望结果
由于实际传入的是 ClassB 实例，转换后的 JSON 应该保留子类的属性（即包含 field3 和 field4）：
```json
{
  "fieldClassA": {
    "field1": "value1",
    "field2": "value2",
    "field3": "value3",
    "field4": "value4"
  }
}
```
### 实际结果
子类 ClassB 的特有属性被裁剪掉，仅剩下了父类 ClassA 的属性：
```
{
  "fieldClassA": {
    "field1": "value1",
    "field2": "value2"
  }
}
```

#### 附加信息
实际测试2.0.54没有问题，2.0.57、5.0.61、2.0.62试了都有问题

hutiefang76:

Fixes #7655.

When `com.alibaba.fastjson.JSON.toJSON` converts a bean whose field is declared as a parent type but currently holds a subclass instance, `ObjectWriterAdapter#toJSONObject` used the declared field type writer. As a result, subclass-only properties were omitted from the nested `JSONObject`.

### Summary of your change

Use the runtime value class writer for non-null object fields when the runtime class differs from the declared field class, so `toJSON` keeps the same subtype fields that normal serialization can see.

Added a fastjson1-compatible regression test covering a `ClassA` field that holds a `ClassB extends ClassA` instance and verifies all parent and child fields are present.

- [x] Made sure commit message follow the rule of Conventional Commits specification.

Tests run:

- `./mvnw -s /tmp/codex-empty-maven-settings.xml -pl fastjson1-compatible -am -Dtest=com.alibaba.fastjson.issue_7600.Issue7655 -Dsurefire.failIfNoSpecifiedTests=false test -DskipITs -Dcheckstyle.skip -Drat.skip=true`
- `./mvnw -s /tmp/codex-empty-maven-settings.xml -pl fastjson1-compatible -am -Dtest=com.alibaba.fastjson.JSONFromObjectTest,com.alibaba.fastjson.issue_7600.Issue7655 -Dsurefire.failIfNoSpecifiedTests=false test -DskipITs -Dcheckstyle.skip -Drat.skip=true`
- `./mvnw -s /tmp/codex-empty-maven-settings.xml -pl fastjson1-compatible -am test -DskipITs -Dcheckstyle.skip -Drat.skip=true`

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
@@ -611,7 +611,17 @@ public class ObjectWriterAdapter<T>
     }
 
     public JSONObject toJSONObject(T object, long features) {
+        return toJSONObject(object, features, new IdentityHashMap<Object, JSONObject>());
+    }
+
+    private JSONObject toJSONObject(T object, long features, IdentityHashMap refs) {
+        JSONObject existing = (JSONObject) refs.get(object);
+        if (existing != null) {
+            return existing;
+        }
+
         JSONObject jsonObject = new JSONObject();
+        refs.put(object, jsonObject);
 
         for (int i = 0, size = fieldWriters.size(); i < size; i++) {
             FieldWriter fieldWriter = fieldWriters.get(i);
@@ -681,16 +691,19 @@ public class ObjectWriterAdapter<T>
                 }
             }
             if (fieldWriter instanceof FieldWriterObject && fieldValue != null && !(fieldValue instanceof Map)) {
-                ObjectWriter valueWriter = fieldWriter.getInitWriter();
-                if (fieldValue.getClass() != fieldWriter.fieldClass) {
-                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
+                FieldWriterObject fieldWriterObject = (FieldWriterObject) fieldWriter;
+                Class<?> fieldValueClass = fieldValue.getClass();
+                ObjectWriter valueWriter = fieldWriterObject.getInitWriter();
+                if (fieldValueClass != fieldWriter.fieldClass
+                        && !fieldWriterObjectWriteUsing(fieldWriterObject, valueWriter, fieldValueClass)) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValueClass, this.features | features);
                 } else if (valueWriter == null) {
                     valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
                 }
                 if (valueWriter instanceof ObjectWriterAdapter) {
                     ObjectWriterAdapter objectWriterAdapter = (ObjectWriterAdapter) valueWriter;
                     if (!objectWriterAdapter.getFieldWriters().isEmpty()) {
-                        fieldValue = objectWriterAdapter.toJSONObject(fieldValue);
+                        fieldValue = objectWriterAdapter.toJSONObject(fieldValue, features, refs);
                     } else {
                         fieldValue = JSON.toJSON(fieldValue);
                     }
@@ -698,8 +711,20 @@ public class ObjectWriterAdapter<T>
             }
             jsonObject.put(fieldWriter.fieldName, fieldValue);
         }
-
+        refs.remove(object);
         return jsonObject;
+    }
+
+    private static boolean fieldWriterObjectWriteUsing(
+            FieldWriterObject fieldWriterObject,
+            ObjectWriter valueWriter,
+            Class<?> fieldValueClass
+    ) {
+        Class<?> initValueClass = fieldWriterObject.initValueClass;
+        return valueWriter != null
+                && fieldWriterObject.writeUsing
+                && initValueClass != null
+                && initValueClass.isAssignableFrom(fieldValueClass);
     }
 
     @Override
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
@@ -682,7 +682,9 @@ public class ObjectWriterAdapter<T>
             }
             if (fieldWriter instanceof FieldWriterObject && fieldValue != null && !(fieldValue instanceof Map)) {
                 ObjectWriter valueWriter = fieldWriter.getInitWriter();
-                if (valueWriter == null) {
+                if (fieldValue.getClass() != fieldWriter.fieldClass) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
+                } else if (valueWriter == null) {
                     valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
                 }
                 if (valueWriter instanceof ObjectWriterAdapter) {
--- a/core/src/test/java/com/alibaba/fastjson2/issues_2200/Issue2286.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_2200/Issue2286.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.issues_2200;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
 import com.alibaba.fastjson2.JSONWriter;
 import com.alibaba.fastjson2.annotation.JSONField;
 import com.alibaba.fastjson2.writer.ObjectWriter;
@@ -10,6 +11,7 @@ import org.junit.jupiter.api.Test;
 import java.lang.reflect.Type;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
 
 @Tag("regression")
 @Tag("annotation")
@@ -19,6 +21,18 @@ public class Issue2286 {
         Bean bean = new Bean();
         bean.animal = new Cat();
         assertEquals("{\"animal\":{\"type\":\"Cat\"}}", JSON.toJSONString(bean));
+    }
+
+    @Test
+    public void toJSONShouldNotBypassSerializeUsingForSubtype() {
+        Bean bean = new Bean();
+        Cat cat = new Cat();
+        cat.color = "black";
+        bean.animal = cat;
+
+        JSONObject jsonObject = (JSONObject) JSON.toJSON(bean);
+
+        assertSame(cat, jsonObject.get("animal"));
     }
 
     public static class Bean {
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -682,7 +682,9 @@ public JSONObject toJSONObject(T object, long features) {
             }
             if (fieldWriter instanceof FieldWriterObject && fieldValue != null && !(fieldValue instanceof Map)) {
                 ObjectWriter valueWriter = fieldWriter.getInitWriter();
-                if (valueWriter == null) {
+                if (fieldValue.getClass() != fieldWriter.fieldClass) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
```
wenshao: **[Critical]** This change introduces a `StackOverflowError` regression in `JSON.toJSON` for cyclic object graphs reached through a polymorphic field.

By resolving the writer from the runtime class, `toJSONObject` now traverses subtype-only fields (e.g. a back-reference) that the declared-type writer omitted, and then recurses at line 693 (`objectWriterAdapter.toJSONObject(fieldValue)`) with no reference/cycle guard. `JSON.toJSON` only takes this `toJSONObject` fast-path when `ReferenceDetection` is off (`JSON.java:4080-4084`), and the recursion never goes through a `JSONWriter`, so the `JSONWriter` level cap never engages. The result is an uncaught `StackOverflowError` (an `Error`, not the `JSONException` that `com.alibaba.fastjson.JSON.toJSON` catches).

Reproduced on this branch vs `main` — field declared as an interface `Tree`, value a `TreeNode` whose `parent` closes a cycle:
- `main`: `JSON.toJSON(holder)` → `{"node":{"id":"child"}}`
- this PR: `JSON.toJSON(holder)` → `StackOverflowError`

The declared-type writer previously did not include the cycle-forming subtype field, so serialization terminated; following the runtime subtype's fields now re-enters the cycle forever. (`toJSONObject` already lacked cycle protection for same-type nested beans — this change widens the exposure to polymorphic graphs.) Please add reference/cycle protection to the `toJSONObject` recursion, e.g. thread an `IdentityHashMap` of in-progress objects, which also closes the pre-existing gap.

_— claude-opus-4-8[1m] via /qreview_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -682,7 +682,9 @@ public JSONObject toJSONObject(T object, long features) {
             }
             if (fieldWriter instanceof FieldWriterObject && fieldValue != null && !(fieldValue instanceof Map)) {
                 ObjectWriter valueWriter = fieldWriter.getInitWriter();
-                if (valueWriter == null) {
+                if (fieldValue.getClass() != fieldWriter.fieldClass) {
```
wenshao: **[Suggestion]** Checking `fieldValue.getClass() != fieldWriter.fieldClass` before `valueWriter == null` discards the field's configured writer (`getInitWriter()`) for every subtype value. For a field with a custom `@JSONField(serializeUsing = ...)` writer (stored in `initObjectWriter` with `writeUsing = true`), `toJSON` now ignores that writer and falls back to the runtime class's default writer — so `toJSON` diverges from `toJSONString`, and a custom writer that intentionally narrows/redacts output is bypassed.

The normal serialization path handles this in `FieldWriterObject.getObjectWriter` via the `writeUsing && initValueClass.isAssignableFrom(valueClass)` type-match; consider mirroring that so a configured custom writer is preserved while still selecting the runtime writer for plain beans. Note that a naive "init-writer first" reorder is not enough — if `initObjectWriter` was already lazily cached as the declared-type writer, it would re-break #7655.

_— claude-opus-4-8[1m] via /qreview_

wenshao (changes requested):

wenshao (approved):
No issues found. LGTM! ✅ _— glm-5.2 via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -681,25 +691,42 @@ public JSONObject toJSONObject(T object, long features) {
...
             }
             if (fieldWriter instanceof FieldWriterObject && fieldValue != null && !(fieldValue instanceof Map)) {
-                ObjectWriter valueWriter = fieldWriter.getInitWriter();
-                if (valueWriter == null) {
+                FieldWriterObject fieldWriterObject = (FieldWriterObject) fieldWriter;
+                Class<?> fieldValueClass = fieldValue.getClass();
+                ObjectWriter valueWriter = fieldWriterObject.getInitWriter();
+                if (fieldValueClass != fieldWriter.fieldClass
+                        && !fieldWriterObjectWriteUsing(fieldWriterObject, valueWriter, fieldValueClass)) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValueClass, this.features | features);
```
wenshao, line 699: **[Critical]** Switching to the runtime writer here leaves two value shapes outside the new ref-aware projection. First, broad fields such as `Object value` holding an `ArrayList` now select `ObjectWriterImplList`; because that writer is not an `ObjectWriterAdapter`, the block below does nothing and `jsonObject.put(...)` stores the raw `java.util.List`. On `main`, the declared `Object` writer had no fields and fell back to `JSON.toJSON(fieldValue)`, producing a `JSONArray`.

Second, when this branch exposes a subtype-only `Collection` field, its elements are still converted by the existing top-level `JSON.toJSON(item)` call, which starts a fresh refs map. A graph like `Holder.base = child` and `child.owners = List.of(holder)` can therefore recurse as `Holder -> child -> owners -> JSON.toJSON(holder) -> ...` until `StackOverflowError`.

Suggested fix: only use the direct runtime projection for adapter writers that are safe to project, and route non-adapter values and collection elements through a conversion path that preserves JSON container types and the shared reference map.

_— GPT-5 via Qwen Code /review_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -681,25 +691,42 @@ public JSONObject toJSONObject(T object, long features) {
...
+                } else if (valueWriter == null) {
                     valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
                 }
                 if (valueWriter instanceof ObjectWriterAdapter) {
                     ObjectWriterAdapter objectWriterAdapter = (ObjectWriterAdapter) valueWriter;
                     if (!objectWriterAdapter.getFieldWriters().isEmpty()) {
-                        fieldValue = objectWriterAdapter.toJSONObject(fieldValue);
+                        fieldValue = objectWriterAdapter.toJSONObject(fieldValue, features, refs);
```
wenshao, line 706: **[Critical]** This direct private recursion bypasses writer-level shape metadata. Top-level `JSON.toJSON` only uses the `toJSONObject` shortcut when the selected writer does not have `WriteClassName`, and `ObjectWriterRootName` preserves `@JSONType(rootName=...)` by overriding the public `toJSONObject` method. Nested runtime subtype writers now go through this private helper unconditionally, so type metadata such as `@type`/custom type keys, and root-name wrappers, can be dropped from the projected `JSONObject` even though normal serialization preserves them.

Suggested fix: make the ref-aware projection honor the same writer safety checks and virtual overrides, or fall back to writer-based conversion for writers that need type/root metadata while still preserving the shared reference map.

_— GPT-5 via Qwen Code /review_

wenshao (changes requested):

hutiefang76:
Thanks for the review. The remaining feedback shows this runtime-writer approach can bypass writer-level shape metadata and affects more JSON.toJSON semantics than the original regression test covered. I'll close this PR rather than keep broadening the change. If there is a narrower accepted direction for #7655, I can rework it from that direction with targeted coverage.

PR #7679, opened 2026-07-17
fix contains method when emtpy for issue #7670

yanxutao89:

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
@@ -31,7 +31,8 @@ abstract class JSONPathSegment {
 
     public boolean contains(JSONPath.Context context) {
         eval(context);
-        return context.value != null;
+        Object value = context.value;
+        return !(value == null || (value instanceof Collection && ((Collection<?>) value).isEmpty()) || (value instanceof Map && ((Map<?, ?>) value).isEmpty()));
     }
 
     public boolean remove(JSONPath.Context context) {
@@ -1539,7 +1540,7 @@ abstract class JSONPathSegment {
 
             public void accept(Object object) {
                 if (object instanceof Map) {
-                    for (Iterator<Map.Entry> it = ((Map) object).entrySet().iterator(); it.hasNext();) {
+                    for (Iterator<Map.Entry> it = ((Map) object).entrySet().iterator(); it.hasNext(); ) {
                         Map.Entry entry = it.next();
                         if (name.equals(entry.getKey())) {
                             it.remove();
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
@@ -32,7 +32,7 @@ abstract class JSONPathSegment {
     public boolean contains(JSONPath.Context context) {
         eval(context);
         Object value = context.value;
-        return !(value == null || (value instanceof Collection && ((Collection<?>) value).isEmpty()) || (value instanceof Map && ((Map<?, ?>) value).isEmpty()));
+        return !(value == null || (value instanceof Collection && ((Collection<?>) value).isEmpty()));
     }
 
     public boolean remove(JSONPath.Context context) {
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
@@ -8,7 +8,6 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class Issue7670 {
-
     @Test
     public void test() {
         String json = "{\"classes\":[{\"className\":\"yuwen\"},{\"className\":\"shuxue\"}]}";
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
@@ -1,0 +1,20 @@
+package com.alibaba.fastjson2.issues_7000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONPath;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue7670 {
+
+    @Test
+    public void test() {
+        String json = "{\"classes\":[{\"className\":\"yuwen\"},{\"className\":\"shuxue\"}]}";
+        Object test = JSON.parse(json);
+        assertTrue(JSONPath.contains(test, "$.classes[?(@.className == 'yuwen')]"));
+        assertTrue(JSONPath.contains(test, "$.classes[?(@.className == 'shuxue')]"));
+        assertFalse(JSONPath.contains(test, "$.classes[?(@.className == 'yingyu')]"));
+    }
+}
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
@@ -611,7 +611,7 @@ abstract class JSONPathFilter
                         array.add(item);
                     }
                 }
-                context.value = array.isEmpty() ? null : array;
+                context.value = array;
                 context.eval = true;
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
```diff
@@ -31,7 +31,8 @@ abstract class JSONPathSegment {
 
     public boolean contains(JSONPath.Context context) {
         eval(context);
-        return context.value != null;
+        Object value = context.value;
+        return !(value == null || (value instanceof Collection && ((Collection<?>) value).isEmpty()) || (value instanceof Map && ((Map<?, ?>) value).isEmpty()));
```
wenshao: **[Critical]** The empty-Collection/empty-Map check in the base `contains()` regresses index-style segments. `JSONPathSegmentIndex.eval()` sets `context.value` to the selected element itself (not a match-list), so an existing element that happens to be an empty `{}` or `[]` is now wrongly reported as "not contained". — Failure scenario: `JSONPath.contains(JSON.parse("{\"items\":[{}]}"), "$.items[0]")` returns `false` (previously `true`), because the element at index 0 is an empty Map. The element **exists** at the specified index but the new code treats it as absent. Same for `$.items[0]` on `{"items":[[]]}` and `$[0]` on `[{}]`.

Consider keeping the base method as `return context.value != null;` and overriding `contains()` in the filter segment base (`JSONPathFilter`) to add the `isEmpty()` check there, so the fix applies only to filter expressions where an empty result means "no match":

```java
// In JSONPathSegment (base — revert to original):
public boolean contains(JSONPath.Context context) {
    eval(context);
    return context.value != null;
}

// In JSONPathFilter (new override):
@Override
public boolean contains(JSONPath.Context context) {
    eval(context);
    Object value = context.value;
    return value != null
        && !(value instanceof Collection && ((Collection<?>) value).isEmpty())
        && !(value instanceof Map && ((Map<?, ?>) value).isEmpty());
}
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue7670 {
+    @Test
+    public void test() {
+        String json = "{\"classes\":[{\"className\":\"yuwen\"},{\"className\":\"shuxue\"}]}";
+        Object test = JSON.parse(json);
+        assertTrue(JSONPath.contains(test, "$.classes[?(@.className == 'yuwen')]"));
+        assertTrue(JSONPath.contains(test, "$.classes[?(@.className == 'shuxue')]"));
+        assertFalse(JSONPath.contains(test, "$.classes[?(@.className == 'yingyu')]"));
```
wenshao, line 17: **[Suggestion]** The new `empty Map` branch in `contains()` has no test coverage — only the `empty Collection` branch is exercised by these filter assertions. — Concrete cost: if the Map branch were incorrect (e.g. a future refactor accidentally inverts the check), no test would catch the regression. Consider adding a test case that exercises the base `contains()` with an empty Map result.

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
wenshao: **[Critical]** The empty-Collection/Map check in the base `contains()` is too broad — it regresses `contains` for non-filter segments (e.g., `JSONPathSegmentIndex`) where an empty collection/map is a legitimate value the path points to. `JSONPathSegmentIndex` does not override `contains()`, so it inherits this base-class check designed for filter semantics. — Failure scenario: `JSONPath.contains(JSON.parse("{\"a\": [[]]}"), "$.a[0]")` now returns `false` instead of `true`. The path `$.a[0]` exists and points to a valid empty array, but the new check treats the empty Collection as "not contained". Same for `{"a": [{}]}` with `$.a[0]`.

```suggestion
        Object value = context.value;
        return value != null;
```

Consider moving the empty-collection/map check into filter segment classes only (e.g., override `contains` in `JSONPathFilter`), restoring the base class to `return context.value != null;`.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
wenshao, line 17: **[Suggestion]** Test covers only the filter-expression case. Two gaps: (1) the empty-Map branch (`value instanceof Map && ((Map<?, ?>) value).isEmpty()`) has no test coverage — the filter path only produces an empty Collection, never an empty Map; (2) the base-class change also alters `contains()` semantics for wildcard (`[*]`), recursive descent (`..name`), range-index, and multi-index segments, none of which are exercised. — Concrete cost: if the Map branch had a logic error, or a future refactor broke non-filter segment behavior, no test would catch it.

Consider adding assertions for the empty-Map branch and at least one non-filter segment type (e.g., wildcard on empty/non-empty arrays).

wenshao (commented):
⚠️ This run could not certify that any of this diff was reviewed. Suggestions are inline. Not reviewed: the entire diff — no agent reported covering it; nobody read it. Not reviewed: every dimension — none of the 12 required agents is on record as launched with a prompt this skill built, so this diff was reviewed, if at all, from prompts the run wrote for itself: no record shows the severity bar, the finding format or this project's own rules reaching an agent. Not reviewed: verification and reverse audit — neither the verifier nor the reverse auditor was launched with a prompt this skill builds — the posted findings were ruled on, and the misses the rest of the review left were hunted, if at all, without the briefs this skill certifies against.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7670.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue7670 {
+    @Test
+    public void test() {
+        String json = "{\"classes\":[{\"className\":\"yuwen\"},{\"className\":\"shuxue\"}]}";
+        Object test = JSON.parse(json);
```
wenshao, line 14: **[Suggestion]** The new empty-Map branch in `contains()` (`value instanceof Map && ((Map<?, ?>) value).isEmpty()`) has no test coverage — only the empty-Collection branch is exercised by these filter assertions. A mutation deleting the Map condition would survive undetected. — Concrete cost: if a future refactor accidentally drops or breaks the empty-Map clause, no test in the suite would fail. Note: to exercise the base `contains()`, the test must use a non-name segment (e.g. index or filter), since `$.a` routes through `JSONPathSegmentName.contains()` which overrides the base. The right test case depends on how the Critical above (base-class check too broad) is resolved.

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
```diff
@@ -31,7 +31,8 @@ abstract class JSONPathSegment {
 
     public boolean contains(JSONPath.Context context) {
         eval(context);
-        return context.value != null;
+        Object value = context.value;
+        return !(value == null || (value instanceof Collection && ((Collection<?>) value).isEmpty()));
```
wenshao, line 35: **[Critical]** The empty-Collection check was added to the base `JSONPathSegment.contains()`, but only `JSONPathSegmentName` overrides `contains()` — index-style segments (`JSONPathSegmentIndex`, `RangeIndexSegment`, `MultiIndexSegment`, `RandomIndexSegment`) inherit this base method. For an index segment, `eval()` sets `context.value` to the selected element itself (`list.get(index)`), which can legitimately be an empty Collection, so an existing element is wrongly reported as "not contained". This is the same defect as the two prior Criticals — the "fix remove map" commit removed the empty-Map half but kept the empty-Collection half. — Failure scenario: `JSONPath.contains(JSON.parse("{\"a\": [[]]}"), "$.a[0]")` returns `false` (previously `true`): the element at index 0 exists, its value is the empty array `[]`. Same for `$[0]` on `[[]]` and `$.items[0]` on `{"items":[[]]}`.

```suggestion
        return context.value != null;
```

Then move the empty-Collection guard into a `contains()` override on the filter segment (`JSONPathFilter`/`GroupFilter`), where an empty result genuinely means "no match":

```java
@Override
public boolean contains(JSONPath.Context context) {
    eval(context);
    Object value = context.value;
    return value != null
        && !(value instanceof Collection && ((Collection<?>) value).isEmpty());
}
```

wenshao (changes requested):

PR #7698, opened 2026-07-26
fix: add serialVersionUID to JSONArray in fastjson1-compatible for Java serialization compatibility

### LINKED ISSUE #7687: [BUG] 升级fastjson 2.0.62 兼容版本后需要清理相关缓存
### 使用版本 2.0.62

Image

### 升级原因：1.2.83 漏洞

Image

[【已复现】无需 gadget、黑名单失效--Fastjson 1.2.83 远程代码执行漏洞](https://mp.weixin.qq.com/s/cfz3mrZdnobLEjQ9f6UWjA?scene=1&click_id=62963893)

### 问题描述

Image

由于序列化ID不一致，反序列化会出错，因此需要清理缓存，保证版本一致。

这个应该是兼容版本漏掉ID=1L导致的。

报错信息：
```
java.io.InvalidClassException: com.alibaba.fastjson.JSONArray; local class incompatible: stream classdesc serialVersionUID = 3642856726642946224, local class serialVersionUID = 1
	at java.io.ObjectStreamClass.initNonProxy(ObjectStreamClass.java:699)
	at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:2005)
	at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1852)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2186)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1669)
```

wenshao:
## Issue

Fixes #7687

## Problem

`com.alibaba.fastjson.JSONArray` in the fastjson1-compatible module implements `Serializable` but was missing an explicit `serialVersionUID`. Java auto-computes a UID from the class structure, which doesn't match the original fastjson 1.x value (`1L`). This causes `InvalidClassException` when deserializing objects previously serialized with fastjson 1.x:

```
java.io.InvalidClassException: com.alibaba.fastjson.JSONArray; local class incompatible:
  stream classdesc serialVersionUID = 3642856726642946224,
  local class serialVersionUID = 1
```

## Fix

Added `private static final long serialVersionUID = 1L;` to `JSONArray`, matching:
- The original fastjson 1.x `JSONArray` value
- The sibling `JSONObject` class in the same module (which already declares `serialVersionUID = 1L`)

## Test

Added `JSONArrayTest_readObject` verifying:
- `serialVersionUID` equals `1L` (via `ObjectStreamClass.lookup`)
- Round-trip Java serialization/deserialization works correctly

### DIFF
--- a/fastjson1-compatible/src/main/java/com/alibaba/fastjson/JSONArray.java
+++ b/fastjson1-compatible/src/main/java/com/alibaba/fastjson/JSONArray.java
@@ -27,6 +27,8 @@ public class JSONArray
     static ObjectReader<JSONArray> arrayReader;
     static ObjectReader<JSONObject> objectReader;
 
+    private static final long serialVersionUID = 1L;
+
     private List list = new com.alibaba.fastjson2.JSONArray();
 
     protected transient Object relatedArray;
--- a/fastjson1-compatible/src/test/java/com/alibaba/fastjson/JSONArrayTest_readObject.java
+++ b/fastjson1-compatible/src/test/java/com/alibaba/fastjson/JSONArrayTest_readObject.java
@@ -1,0 +1,62 @@
+package com.alibaba.fastjson;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JSONArrayTest_readObject {
+    @Test
+    public void test_serialVersionUID() {
+        ObjectStreamClass desc = ObjectStreamClass.lookup(JSONArray.class);
+        assertEquals(1L, desc.getSerialVersionUID());
+    }
+
+    @Test
+    public void test_0() throws Exception {
+        JSONArray jsonArray = new JSONArray();
+        jsonArray.add(123);
+        jsonArray.add("hello");
+        jsonArray.add(new JSONObject());
+
+        ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+        ObjectOutputStream objOut = new ObjectOutputStream(bytesOut);
+        objOut.writeObject(jsonArray);
+        objOut.flush();
+
+        byte[] bytes = bytesOut.toByteArray();
+
+        ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytes);
+        ObjectInputStream objIn = new ObjectInputStream(bytesIn);
+
+        Object obj = objIn.readObject();
+
+        assertEquals(JSONArray.class, obj.getClass());
+        assertEquals(jsonArray, obj);
+    }
+
+    @Test
+    public void test_1() throws Exception {
+        JSONArray jsonArray = JSON.parseArray("[1,2,3,{\"id\":123}]");
+
+        ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+        ObjectOutputStream objOut = new ObjectOutputStream(bytesOut);
+        objOut.writeObject(jsonArray);
+        objOut.flush();
+
+        byte[] bytes = bytesOut.toByteArray();
+
+        ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytes);
+        ObjectInputStream objIn = new ObjectInputStream(bytesIn);
+
+        Object obj = objIn.readObject();
+
+        assertEquals(JSONArray.class, obj.getClass());
+        assertEquals(jsonArray, obj);
+    }
+}


core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -580,6 +580,9 @@ public Object readAny() {
             }
             case BC_BIGINT: {
                 int len = readInt32Value();
+                if (len < 0 || len > end - offset) {
+                    throw new JSONException("BC_BIGINT length out of range: " + len);
+                }
                 byte[] buf = new byte[len];
```
wenshao, line 586: **[Suggestion]** The same 3-line length guard is copy-pasted verbatim into four `BC_BIGINT` sites instead of being extracted into a check helper. This file already centralizes fixed-size bounds checks as static helpers (`check3`/`check7`/`outOfBoundsCheckFromToIndex`, lines 6291–6307), so the variable-length case diverges from the established convention. — Concrete cost: if the comparison or message ever needs correcting, the fix must be applied in four places and one can be missed (the asymmetric-sibling hazard).

```suggestion
            case BC_BIGINT: {
                int len = readInt32Value();
                checkBigintLen(len, offset, end);
                byte[] buf = new byte[len];
```

(with a new `static void checkBigintLen(int len, int offset, int end)` helper alongside `check3`/`check7`, called at all four sites)

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
```diff
@@ -0,0 +1,43 @@
...
+                (byte) 0x48,
+                (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
+        };
+        assertThrows(JSONException.class, () -> JSONB.parse(payload));
+    }
+
+    @Test
+    public void testBigIntValidPayloadStillWorks() {
+        // Verify legitimate BigInteger values still round-trip correctly
+        java.math.BigInteger value = java.math.BigInteger.valueOf(Long.MAX_VALUE).multiply(java.math.BigInteger.TEN);
```
wenshao: **[Suggestion]** Fully-qualified class names (`java.math.BigInteger`, `org.junit.jupiter.api.Assertions.assertEquals`) are used inline instead of imports, inconsistent with this file's own static import of `assertThrows` and with neighboring test files. — Concrete cost: readability; the mixed import/FQN style within a single file signals an unfinished edit.

```suggestion
        BigInteger value = BigInteger.valueOf(Long.MAX_VALUE).multiply(BigInteger.TEN);
```

(add `import java.math.BigInteger;` and `import static org.junit.jupiter.api.Assertions.assertEquals;` to the import block)

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
```diff
@@ -0,0 +1,43 @@
...
+public class Issue7669 {
+    @Test
+    public void testBigIntDeclaredLengthOOM() {
+        // 0xBB = BC_BIGINT, 0x48 = BC_INT32, 0x7FFFFFFF = Integer.MAX_VALUE
+        byte[] payload = {
+                (byte) 0xBB,
+                (byte) 0x48,
+                (byte) 0x7F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
+        };
+        assertThrows(JSONException.class, () -> JSONB.parse(payload));
```
wenshao: **[Suggestion]** All three `BC_BIGINT` tests exercise only `readAny()` (via `JSONB.parse`); the identical guards in `readBigInteger()`, `readNumber()`, and `readString()` have no test that would catch their removal. — Failure scenario: a crafted JSONB payload with a `BigInteger` field whose `BC_BIGINT` declared length is `Integer.MAX_VALUE`, deserialized via `JSONB.parseObject(payload, Bean.class)`, routes through `readBigInteger()`. If a future refactor drops the guard there, the OOM reopens and no test fails.

Consider adding a typed-path test, e.g.:
```java
@Test
public void testBigIntDeclaredLengthOOM_typedPath() {
    // Craft a JSONB object with a BigInteger field encoding BC_BIGINT(len=MAX_VALUE)
    assertThrows(JSONException.class, () -> JSONB.parseObject(payload, Bean.class));
}
```

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -3253,6 +3256,9 @@ private String readStringTypeNotMatch() {
                 return Long.toString(int64Value);
             case BC_BIGINT: {
                 int len = readInt32Value();
+                if (len < 0 || len > end - offset) {
+                    throw new JSONException("BC_BIGINT length out of range: " + len);
+                }
```
wenshao: **[Suggestion]** Exception message omits the available buffer size, making corrupted-payload triage harder. — Failure scenario: at 3 AM the oncall sees `BC_BIGINT length out of range: 2147483647` but cannot tell whether the buffer had 0 bytes remaining (truncation) or 2 billion (malicious craft). The existing `outOfBoundsCheckFromToIndex` helper already includes both offset and end in its message, so including diagnostic context is this file's established practice.

```suggestion
                if (len < 0 || len > end - offset) {
                    throw new JSONException("BC_BIGINT length out of range: " + len + ", available: " + (end - offset));
                }
```

(applied consistently at all four sites)

wenshao (commented):

mujPhoto:
awesome

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
wenshao: Done. Extracted into `checkBigintLen(int len, int offset, int end)` alongside `check3`/`check7`, called at all four sites.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
wenshao: Done. Replaced FQN with proper imports for `BigInteger` and `assertEquals`.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
wenshao: Done. Added typed-path tests exercising `readBigInteger()` (`BigInteger.class`), `readString()` (`String.class`), and `readNumber()` (`Number.class`) with the same malicious payload.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
wenshao: Done. Message now includes available buffer size: `"BC_BIGINT length out of range: " + len + ", available: " + (end - offset)`, applied consistently via the shared helper.

wenshao (commented):
Reviewed. 1 Suggestion-level finding(s) could not be anchored to a changed line and were dropped; nothing further to act on here.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR. Reviewed.

wenshao:
## Code Review

### Overview

Adds `private static final long serialVersionUID = 1L;` to `com.alibaba.fastjson.JSONArray` in `fastjson1-compatible`, so that Java-serialized payloads produced by fastjson 1.x can be read back by the compat module (#7687). Plus a new test class with 3 tests.

**Verdict: correct and minimal — LGTM.** Notes below are mostly about test strength and a residual (pre-existing) interop gap.

### Correctness — verified end-to-end

I checked the actual interop scenario rather than relying on the round-trip tests:

- Serialized `JSON.parseArray("[1,\"hello\",{\"id\":123}]")` with **real fastjson 1.2.83**, then deserialized with this PR's `fastjson1-compatible` classes → **succeeds** (`[1,"hello",{"id":123}]`, size 3). Without the patch this is the reported `InvalidClassException` (auto-computed UID `3642856726642946224`).
- The serialized field shape matches 1.x exactly: field name `list`, erased type `Ljava/util/List;`, and `relatedArray` / `componentType` are `transient` in both versions. So field-level compatibility is genuinely restored, not just the UID check.
- `JSON` (the superclass) is not `Serializable` and has only static fields, so nothing is silently dropped, and its implicit no-arg constructor satisfies the deserialization requirement.
- `JSONArray` was the **only** `Serializable` class in `fastjson1-compatible/src/main/java` still missing an explicit UID — this closes the gap. `1L` matches both fastjson 1.x and the sibling `JSONObject`.

### Test coverage — the main suggestion

`test_0` and `test_1` are same-JVM round trips, so **they pass with or without the fix** (I confirmed: reverting the one-line change leaves only `test_serialVersionUID` failing). They're fine as smoke tests, but the actual regression is guarded by a single `ObjectStreamClass.lookup` assertion.

A stronger guard is a golden byte stream captured from real fastjson 1.x — it pins the whole serialized form (field names/types), not just the UID, so a future refactor that renames `list` or changes its declared type is caught too. Bytes below were produced by fastjson 1.2.83 for `[1,"hello",{"id":123}]`:

```java
@Test
public void test_readFastjson1Bytes() throws Exception {
    // produced by fastjson 1.2.83: JSON.parseArray("[1,\"hello\",{\"id\":123}]")
    String base64 = "rO0ABXNyAB5jb20uYWxpYmFiYS5mYXN0anNvbi5KU09OQXJyYXkAAAAAAAAAAQIAAUwABGxpc3R0"
            + "ABBMamF2YS91dGlsL0xpc3Q7eHBzcgATamF2YS51dGlsLkFycmF5TGlzdHiB0h2Zx2GdAwABSQAE"
            + "c2l6ZXhwAAAAA3cEAAAAA3NyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4"
            + "cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAABdAAFaGVsbG9zcgAfY29tLmFsaWJh"
            + "YmEuZmFzdGpzb24uSlNPTk9iamVjdAAAAAAAAAABAgABTAADbWFwdAAPTGphdmEvdXRpbC9NYXA7"
            + "eHBzcgARamF2YS51dGlsLkhhc2hNYXAFB9rBwxZg0QMAAkYACmxvYWRGYWN0b3JJAAl0aHJlc2hv"
            + "bGR4cD9AAAAAAAAMdwgAAAAQAAAAAXQAAmlkc3EAfgAFAAAAe3h4";
    byte[] bytes = java.util.Base64.getDecoder().decode(base64);
    ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(bytes));
    JSONArray array = (JSONArray) in.readObject();
    assertEquals(JSON.parseArray("[1,\"hello\",{\"id\":123}]"), array);
}
```

I ran this against the PR branch — it passes, and fails with `InvalidClassException` without the fix.

### Residual limitation (pre-existing, out of scope but worth knowing)

The **reverse** direction is still broken for arrays built via the default constructor:

```java
JSONArray a = new JSONArray();   // list = com.alibaba.fastjson2.JSONArray
a.add(123);
// deserialize in real fastjson 1.2.83:
// java.lang.ClassNotFoundException: com.alibaba.fastjson2.JSONArray
```

Because `list` is initialized to `new com.alibaba.fastjson2.JSONArray()`, that fastjson2 class name lands in the stream and 1.x can't resolve it. (Arrays from `JSON.parseArray(...)` happen to be fine, and `JSONObject`'s `map` defaults to `HashMap`, so it's unaffected.) If bidirectional compat is a goal, a `writeObject` that normalizes `list` to an `ArrayList` would close it — but that's a separate change; this PR fixes the direction the issue reports.

### Style / minor

- Placement after the two static `ObjectReader` fields mirrors `JSONObject` exactly — consistent with the module. (Conventionally `serialVersionUID` goes first, but matching the sibling class is the better call here.)
- Test class name `JSONArrayTest_readObject` follows the module's existing `JSONArrayTest_hashCode` convention. Good.
- Optional: the three tests contain duplicated serialize/deserialize boilerplate — a small private helper would tighten it.

### Performance / security

- Performance: zero impact — a compile-time constant.
- Security: no new attack surface. The class was already `Serializable`; a fixed UID doesn't create a gadget that didn't exist. Worth noting that `JSONArray` has no `readObject` validation, so a hostile stream can install an arbitrary `List` into `list` — but that's inherent to the class being `Serializable` and unchanged by this PR.

PR #7699, opened 2026-07-26
fix: use Unsafe.allocateInstance for non-static inner class to avoid JDK 25 NPE (#7654)

### LINKED ISSUE #7654: [BUG] JDK25字符串转对象包含非静态内部类时空指针, JDK21正常
### 问题描述
JDK25字符串转对象包含非静态内部类时空指针, JDK21正常

### 环境信息
*请填写以下信息：*

 - JDK信息： [e.g.：openjdk version "25.0.3" 2026-04-21 LTS]
 - 版本信息：[e.g.：Fastjson2 2.0.62]
 

### 重现步骤
*如何操作可以重现该问题：*

1. 执行 `main` 方法
2. 出现 `java.lang.NullPointerException` 错误, Summary内部类改成静态内部类或JDK25切回JDK21运行正常

```java

public class AmzListingJsonFeedResult {

        String json = "{\"issues\":[],\"summary\":{\"errors\":0,\"warnings\":0,\"messagesProcessed\":7,\"messagesAccepted\":7,\"messagesInvalid\":0}}";
        AmzListingJsonFeedResult result = JSON.parseObject(json, AmzListingJsonFeedResult.class);
        System.out.println(result.getSummary().getMessagesAccepted());
    }

    private Summary summary;

    public Summary getSummary() {
        return summary;
    }

    public void setSummary(Summary summary) {
        this.summary = summary;
    }

    public class Summary {
        private Integer errors;
        private Integer warnings;
        private Integer messagesProcessed;
        private Integer messagesAccepted;
        private Integer messagesInvalid;

        public Integer getErrors() {
            return errors;
        }

        public void setErrors(Integer errors) {
            this.errors = errors;
        }

        public Integer getWarnings() {
            return warnings;
        }

        public void setWarnings(Integer warnings) {
            this.warnings = warnings;
        }

        public Integer getMessagesProcessed() {
            return messagesProcessed;
        }

        public void setMessagesProcessed(Integer messagesProcessed) {
            this.messagesProcessed = messagesProcessed;
        }

        public Integer getMessagesAccepted() {
            return messagesAccepted;
        }

        public void setMessagesAccepted(Integer messagesAccepted) {
            this.messagesAccepted = messagesAccepted;
        }

        public Integer getMessagesInvalid() {
            return messagesInvalid;
        }

        public void setMessagesInvalid(Integer messagesInvalid) {
            this.messagesInvalid = messagesInvalid;
        }
    }
}
```

### 期待的正确结果
*无报错*

### 相关日志输出
Exception in thread "main" java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:220)
	at AmzListingJsonFeedResult$Summary.<init>(AmzListingJsonFeedResult.java:21)
	at com.alibaba.fastjson2.reader.ORG_2_5_Summary.readObject(Unknown Source)
	at com.alibaba.fastjson2.reader.ORG_1_1_AmzListingJsonFeedResult.readObject(Unknown Source)
	at com.alibaba.fastjson2.JSON.parseObject(JSON.java:992)
	at AmzListingJsonFeedResult.main(AmzListingJsonFeedResult.java:7)

wenshao:
## Problem

On JDK 25, deserializing JSON into a class containing a non-static inner class throws `NullPointerException`:

```
java.lang.NullPointerException
  at java.base/java.util.Objects.requireNonNull
  at AmzListingJsonFeedResult$Summary.<init>
  at com.alibaba.fastjson2.reader.ORG_2_5_Summary.readObject
```

JDK 25 adds `Objects.requireNonNull(this$0)` in inner class constructors. The previous approach of passing `null` for the enclosing instance and patching `this$0` afterwards via `Unsafe.putObject` now fails because the NPE fires during construction.

## Fix

Use `Unsafe.allocateInstance` to bypass the constructor entirely for non-static inner classes. The existing `this$0` fixup mechanism (via `Unsafe.putObject` in ASM path / reflection in `FieldReaderObject`) continues to set the enclosing instance correctly after creation.

### Changes

- **ObjectReaderCreatorASM**: Route inner class constructors (parameterCount != 0) to `Unsafe.allocateInstance` in both `createInstance` generation and `genCreateObject`
- **ConstructorSupplier**: Use `Unsafe.allocateInstance` instead of `constructor.newInstance(new Object[1])`
- **ObjectReaderAdapter**: Fallback path uses `Unsafe.allocateInstance` when parameterCount > 0

Closes #7654

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
@@ -5,6 +5,7 @@ import com.alibaba.fastjson2.JSONFactory;
 import com.alibaba.fastjson2.codec.FieldInfo;
 import com.alibaba.fastjson2.internal.asm.ASMUtils;
 import com.alibaba.fastjson2.util.Fnv;
+import com.alibaba.fastjson2.util.JDKUtils;
 import com.alibaba.fastjson2.util.TypeUtils;
 
 import java.lang.reflect.*;
@@ -23,6 +24,12 @@ final class ConstructorFunction<T>
     final String[] paramNames;
     final boolean marker;
     final long[] hashCodes;
+    /**
+     * Non-null when parameter 0 is the synthetic enclosing instance of a non-static inner class.
+     * Such a parameter never appears in the JSON, so it must not be left null — the constructor
+     * may dereference it (JDK 25 rejects a null enclosing instance outright).
+     */
+    final Class enclosingParamType;
     final List<Constructor> alternateConstructors;
     Map<Set<Long>, Constructor> alternateConstructorMap;
     Map<Set<Long>, String[]> alternateConstructorNames;
@@ -56,6 +63,15 @@ final class ConstructorFunction<T>
             }
             hashCodes[i] = Fnv.hashCode64(name);
         }
+
+        Class declaringClass = constructor.getDeclaringClass();
+        Class enclosingClass = declaringClass.getDeclaringClass();
+        this.enclosingParamType = enclosingClass != null
+                && !Modifier.isStatic(declaringClass.getModifiers())
+                && parameters.length > 0
+                && parameters[0].getType() == enclosingClass
+                ? enclosingClass
+                : null;
 
         this.alternateConstructors = alternateConstructors;
         if (alternateConstructors != null) {
@@ -99,6 +115,22 @@ final class ConstructorFunction<T>
         }
     }
 
+    /**
+     * Default for a parameter absent from the JSON. Parameter 0 of a non-static inner class
+     * constructor is the enclosing instance, which is never present in the JSON, so allocate a
+     * bare one rather than passing null.
+     */
+    private Object defaultArg(int index, Class<?> paramClass) {
+        if (index == 0 && enclosingParamType != null) {
+            try {
+                return JDKUtils.UNSAFE.allocateInstance(enclosingParamType);
+            } catch (Throwable ignored) {
+                // fall back to null below
+            }
+        }
+        return TypeUtils.getDefaultValue(paramClass);
+    }
+
     @Override
     public T apply(Map<Long, Object> values) {
         boolean containsAll = true;
@@ -139,7 +171,7 @@ final class ConstructorFunction<T>
             Object arg = values.get(hashCodes[0]);
             Class<?> paramType = param.getType();
             if (arg == null) {
-                arg = TypeUtils.getDefaultValue(paramType);
+                arg = defaultArg(0, paramType);
             } else {
                 if (!paramType.isInstance(arg)) {
                     arg = TypeUtils.cast(arg, paramType);
@@ -153,7 +185,7 @@ final class ConstructorFunction<T>
             Parameter param0 = parameters[0];
             Class<?> param0Type = param0.getType();
             if (arg0 == null) {
-                arg0 = TypeUtils.getDefaultValue(param0Type);
+                arg0 = defaultArg(0, param0Type);
             } else {
                 if (!param0Type.isInstance(arg0)) {
                     arg0 = TypeUtils.cast(arg0, param0Type);
@@ -193,8 +225,8 @@ final class ConstructorFunction<T>
                     args[i] = arg;
                 } else {
                     flag |= (1 << i);
-                    if (paramClass.isPrimitive()) {
-                        args[i] = TypeUtils.getDefaultValue(paramClass);
+                    if (paramClass.isPrimitive() || (i == 0 && enclosingParamType != null)) {
+                        args[i] = defaultArg(i, paramClass);
                     }
                 }
                 n = i + 1;
@@ -210,7 +242,7 @@ final class ConstructorFunction<T>
                 Type paramType = parameter.getParameterizedType();
                 Object arg = values.get(hashCodes[i]);
                 if (arg == null) {
-                    arg = TypeUtils.getDefaultValue(paramClass);
+                    arg = defaultArg(i, paramClass);
                 } else {
                     if (!paramClass.isInstance(arg)) {
                         arg = TypeUtils.cast(arg, paramClass);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
@@ -124,8 +124,8 @@ final class ConstructorFunction<T>
         if (index == 0 && enclosingParamType != null) {
             try {
                 return JDKUtils.UNSAFE.allocateInstance(enclosingParamType);
-            } catch (Throwable ignored) {
-                // fall back to null below
+            } catch (InstantiationException ignored) {
+                // the enclosing type cannot be allocated, fall back to the default value below
             }
         }
         return TypeUtils.getDefaultValue(paramClass);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
@@ -4,6 +4,7 @@ import com.alibaba.fastjson2.JSONException;
 import com.alibaba.fastjson2.JSONFactory;
 import com.alibaba.fastjson2.codec.FieldInfo;
 import com.alibaba.fastjson2.internal.asm.ASMUtils;
+import com.alibaba.fastjson2.util.BeanUtils;
 import com.alibaba.fastjson2.util.Fnv;
 import com.alibaba.fastjson2.util.JDKUtils;
 import com.alibaba.fastjson2.util.TypeUtils;
@@ -64,14 +65,7 @@ final class ConstructorFunction<T>
             hashCodes[i] = Fnv.hashCode64(name);
         }
 
-        Class declaringClass = constructor.getDeclaringClass();
-        Class enclosingClass = declaringClass.getDeclaringClass();
-        this.enclosingParamType = enclosingClass != null
-                && !Modifier.isStatic(declaringClass.getModifiers())
-                && parameters.length > 0
-                && parameters[0].getType() == enclosingClass
-                ? enclosingClass
-                : null;
+        this.enclosingParamType = BeanUtils.getEnclosingInstanceParamType(constructor);
 
         this.alternateConstructors = alternateConstructors;
         if (alternateConstructors != null) {
@@ -118,7 +112,10 @@ final class ConstructorFunction<T>
     /**
      * Default for a parameter absent from the JSON. Parameter 0 of a non-static inner class
      * constructor is the enclosing instance, which is never present in the JSON, so allocate a
-     * bare one rather than passing null.
+     * bare one rather than passing null. When the enclosing class cannot be allocated (for
+     * example an abstract class), fall back to null as before JDK 25; inner classes compiled
+     * by JDK 25+ then reject the null enclosing instance, which is unavoidable because an
+     * abstract enclosing class cannot be instantiated.
      */
     private Object defaultArg(int index, Class<?> paramClass) {
         if (index == 0 && enclosingParamType != null) {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
@@ -12,6 +12,7 @@ final class ConstructorSupplier
     final Constructor constructor;
     final Class objectClass;
     final boolean useClassNewInstance;
+    final Class paramType;
 
     public ConstructorSupplier(Constructor constructor) {
         constructor.setAccessible(true);
@@ -20,6 +21,9 @@ final class ConstructorSupplier
         this.useClassNewInstance = constructor.getParameterCount() == 0
                 && Modifier.isPublic(constructor.getModifiers())
                 && Modifier.isPublic(objectClass.getModifiers());
+        this.paramType = constructor.getParameterCount() == 1
+                ? constructor.getParameterTypes()[0]
+                : null;
     }
 
     @Override
@@ -28,8 +32,7 @@ final class ConstructorSupplier
             if (useClassNewInstance) {
                 return objectClass.newInstance();
             } else {
-                if (constructor.getParameterCount() == 1) {
-                    Class<?> paramType = constructor.getParameterTypes()[0];
+                if (paramType != null) {
                     Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
                     return constructor.newInstance(dummy);
                 } else {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
@@ -29,7 +29,9 @@ final class ConstructorSupplier
                 return objectClass.newInstance();
             } else {
                 if (constructor.getParameterCount() == 1) {
-                    return JDKUtils.UNSAFE.allocateInstance(objectClass);
+                    Class<?> paramType = constructor.getParameterTypes()[0];
+                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    return constructor.newInstance(dummy);
                 } else {
                     return constructor.newInstance();
                 }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.reader;
 
 import com.alibaba.fastjson2.JSONException;
+import com.alibaba.fastjson2.util.JDKUtils;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
@@ -28,7 +29,7 @@ final class ConstructorSupplier
                 return objectClass.newInstance();
             } else {
                 if (constructor.getParameterCount() == 1) {
-                    return constructor.newInstance(new Object[1]);
+                    return JDKUtils.UNSAFE.allocateInstance(objectClass);
                 } else {
                     return constructor.newInstance();
                 }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.reader;
 
 import com.alibaba.fastjson2.JSONException;
+import com.alibaba.fastjson2.util.BeanUtils;
 import com.alibaba.fastjson2.util.JDKUtils;
 
 import java.lang.reflect.Constructor;
@@ -21,9 +22,7 @@ final class ConstructorSupplier
         this.useClassNewInstance = constructor.getParameterCount() == 0
                 && Modifier.isPublic(constructor.getModifiers())
                 && Modifier.isPublic(objectClass.getModifiers());
-        this.paramType = constructor.getParameterCount() == 1
-                ? constructor.getParameterTypes()[0]
-                : null;
+        this.paramType = BeanUtils.getEnclosingInstanceParamType(constructor);
     }
 
     @Override
@@ -31,14 +30,24 @@ final class ConstructorSupplier
         try {
             if (useClassNewInstance) {
                 return objectClass.newInstance();
-            } else {
-                if (paramType != null) {
-                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
-                    return constructor.newInstance(dummy);
-                } else {
-                    return constructor.newInstance();
-                }
             }
+
+            if (paramType != null) {
+                Object dummy;
+                try {
+                    dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                } catch (InstantiationException ignored) {
+                    // the enclosing class cannot be allocated (for example an abstract class),
+                    // pass null as before JDK 25
+                    dummy = null;
+                }
+                return constructor.newInstance(dummy);
+            }
+
+            if (constructor.getParameterCount() == 1) {
+                return constructor.newInstance(new Object[1]);
+            }
+            return constructor.newInstance();
         } catch (Throwable e) {
             throw new JSONException("create instance error", e);
         }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
@@ -4,6 +4,7 @@ import com.alibaba.fastjson2.*;
 import com.alibaba.fastjson2.schema.JSONSchema;
 import com.alibaba.fastjson2.util.BeanUtils;
 import com.alibaba.fastjson2.util.Fnv;
+import com.alibaba.fastjson2.util.JDKUtils;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
@@ -443,7 +444,12 @@ public class ObjectReaderAdapter<T>
 
         if (constructor != null) {
             try {
-                T object = (T) constructor.newInstance(new Object[parameterCount]);
+                T object;
+                if (parameterCount == 0) {
+                    object = (T) constructor.newInstance();
+                } else {
+                    object = (T) JDKUtils.UNSAFE.allocateInstance(objectClass);
+                }
                 if (hasDefaultValue) {
                     initDefaultValue(object);
                 }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
@@ -448,7 +448,9 @@ public class ObjectReaderAdapter<T>
                 if (parameterCount == 0) {
                     object = (T) constructor.newInstance();
                 } else {
-                    object = (T) JDKUtils.UNSAFE.allocateInstance(objectClass);
+                    Class<?> paramType = constructor.getParameterTypes()[0];
+                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    object = (T) constructor.newInstance(dummy);
                 }
                 if (hasDefaultValue) {
                     initDefaultValue(object);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
@@ -447,10 +447,12 @@ public class ObjectReaderAdapter<T>
                 T object;
                 if (parameterCount == 0) {
                     object = (T) constructor.newInstance();
-                } else {
+                } else if (parameterCount == 1) {
                     Class<?> paramType = constructor.getParameterTypes()[0];
                     Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
                     object = (T) constructor.newInstance(dummy);
+                } else {
+                    object = (T) constructor.newInstance(new Object[parameterCount]);
                 }
                 if (hasDefaultValue) {
                     initDefaultValue(object);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
@@ -447,9 +447,16 @@ public class ObjectReaderAdapter<T>
                 T object;
                 if (parameterCount == 0) {
                     object = (T) constructor.newInstance();
-                } else if (parameterCount == 1) {
+                } else if (parameterCount == 1 && BeanUtils.getEnclosingInstanceParamType(constructor) != null) {
                     Class<?> paramType = constructor.getParameterTypes()[0];
-                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    Object dummy;
+                    try {
+                        dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    } catch (InstantiationException ignored) {
+                        // the enclosing class cannot be allocated (for example an abstract class),
+                        // pass null as before JDK 25
+                        dummy = null;
+                    }
                     object = (T) constructor.newInstance(dummy);
                 } else {
                     object = (T) constructor.newInstance(new Object[parameterCount]);
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
@@ -706,10 +706,17 @@ public class ObjectReaderCreatorASM
             mw.invokespecial(TYPE_OBJECT, "<init>", "()V");
         } else {
             Class paramType = defaultConstructor.getParameterTypes()[0];
-            mw.getstatic(TYPE_UNSAFE_UTILS, "UNSAFE", "Lsun/misc/Unsafe;");
-            mw.visitLdcInsn(paramType);
-            mw.invokevirtual("sun/misc/Unsafe", "allocateInstance", "(Ljava/lang/Class;)Ljava/lang/Object;");
-            mw.checkcast(ASMUtils.type(paramType));
+            // ldc/checkcast on the enclosing type are access-checked against the generated class,
+            // which lives in DynamicClassLoader; a non-public enclosing type fails with
+            // IllegalAccessError there, so keep passing null for it as before.
+            if (Modifier.isPublic(paramType.getModifiers())) {
+                mw.getstatic(TYPE_UNSAFE_UTILS, "UNSAFE", "Lsun/misc/Unsafe;");
+                mw.visitLdcInsn(paramType);
+                mw.invokevirtual("sun/misc/Unsafe", "allocateInstance", "(Ljava/lang/Class;)Ljava/lang/Object;");
+                mw.checkcast(ASMUtils.type(paramType));
+            } else {
+                mw.aconst_null();
+            }
             mw.invokespecial(TYPE_OBJECT, "<init>", "(" + ASMUtils.desc(paramType) + ")V");
         }
     }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
@@ -620,7 +620,8 @@ public class ObjectReaderCreatorASM
             String methodName = fieldBased && defaultConstructor == null ? "createInstance0" : "createInstance";
 
             if ((externalClass && defaultConstructor != null)
-                    || fieldBased && (defaultConstructor == null || !Modifier.isPublic(defaultConstructor.getModifiers()) || !Modifier.isPublic(objectClass.getModifiers()))) {
+                    || fieldBased && (defaultConstructor == null || !Modifier.isPublic(defaultConstructor.getModifiers()) || !Modifier.isPublic(objectClass.getModifiers()))
+                    || (defaultConstructor != null && defaultConstructor.getParameterCount() != 0)) {
                 MethodWriter mw = cw.visitMethod(
                         Opcodes.ACC_PUBLIC,
                         methodName,
@@ -2953,7 +2954,8 @@ public class ObjectReaderCreatorASM
         int objectModifiers = objectClass == null ? Modifier.PUBLIC : objectClass.getModifiers();
         boolean publicObject = Modifier.isPublic(objectModifiers) && (objectClass == null || !classLoader.isExternalClass(objectClass));
 
-        if (defaultConstructor == null || !publicObject || !Modifier.isPublic(defaultConstructor.getModifiers())) {
+        if (defaultConstructor == null || !publicObject || !Modifier.isPublic(defaultConstructor.getModifiers())
+                || defaultConstructor.getParameterCount() != 0) {
             if (creator != null) {
                 mw.aload(THIS);
                 mw.getfield(classNameType, "creator", "Ljava/util/function/Supplier;");
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
@@ -620,8 +620,7 @@ public class ObjectReaderCreatorASM
             String methodName = fieldBased && defaultConstructor == null ? "createInstance0" : "createInstance";
 
             if ((externalClass && defaultConstructor != null)
-                    || fieldBased && (defaultConstructor == null || !Modifier.isPublic(defaultConstructor.getModifiers()) || !Modifier.isPublic(objectClass.getModifiers()))
-                    || (defaultConstructor != null && defaultConstructor.getParameterCount() != 0)) {
+                    || fieldBased && (defaultConstructor == null || !Modifier.isPublic(defaultConstructor.getModifiers()) || !Modifier.isPublic(objectClass.getModifiers()))) {
                 MethodWriter mw = cw.visitMethod(
                         Opcodes.ACC_PUBLIC,
                         methodName,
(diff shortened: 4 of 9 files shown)


core/src/test/java/com/alibaba/fastjson2/v1issues/issue_1000/Issue1082.java
```diff
@@ -15,15 +14,8 @@
...
-        try {
-            Model_1082 m = (Model_1082) JSON.parseObject("{}", Model_1082.class);
-        } catch (JSONException ex) {
-            error = ex;
-        } catch (NullPointerException ex) {
-            error = ex;
-        }
-        assertNotNull(error);
+        Model_1082 m = (Model_1082) JSON.parseObject("{}", Model_1082.class);
+        assertNotNull(m);
```
wenshao, line 18: **[Critical]** The companion test in `fastjson1-compatible/src/test/java/com/alibaba/fastjson/issue_1000/Issue1082.java` still asserts the old behaviour (expects an exception). After this fix, `JSON.parseObject("{}", Model_1082.class)` succeeds instead of throwing, so `assertNotNull(error)` at line 21 of that test fails. — Failure scenario: `mvn test -pl fastjson1-compatible` → `Issue1082.test_for_issue:21 expected: not <null>` (verified by running the test). This is also the cause of the 29 CI check failures on this PR.

```suggestion
        Model_1082 m = (Model_1082) JSON.parseObject("{}", Model_1082.class);
        assertNotNull(m);
```

Update the `fastjson1-compatible` module's `Issue1082` test to match — remove the try/catch, assert success.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
```diff
@@ -28,7 +29,9 @@ public Object get() {
                 return objectClass.newInstance();
             } else {
                 if (constructor.getParameterCount() == 1) {
-                    return constructor.newInstance(new Object[1]);
+                    Class<?> paramType = constructor.getParameterTypes()[0];
+                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
```
wenshao: **[Suggestion]** `constructor.getParameterTypes()` clones the internal `Class[]` array on every `get()` call. In this hot path (called once per deserialized object on the reflection fallback), that's one unnecessary allocation per instance. — Concrete cost: GC pressure under high-throughput inner-class deserialization; the parameter type is immutable for the lifetime of the `Constructor`.

```suggestion
                Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
```

Cache `paramType` as a field in the constructor (like `ObjectReaderImplEnum.createMethodParamType`), then reference the field here.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
```diff
@@ -443,7 +444,14 @@ public T createInstance(long features) {
...
             try {
-                T object = (T) constructor.newInstance(new Object[parameterCount]);
+                T object;
+                if (parameterCount == 0) {
+                    object = (T) constructor.newInstance();
+                } else {
+                    Class<?> paramType = constructor.getParameterTypes()[0];
+                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    object = (T) constructor.newInstance(dummy);
+                }
```
wenshao, line 463: **[Suggestion]** The `else` branch implicitly assumes `parameterCount == 1` but reads as a catch-all for any non-zero count. `ConstructorSupplier` uses an explicit `if (constructor.getParameterCount() == 1)` guard for the same logic. — Concrete cost: if `getDefaultConstructor` ever returns a 2+ parameter constructor, `constructor.newInstance(dummy)` throws `IllegalArgumentException` with no indication of why; the sibling's explicit guard would make the assumption visible.

```suggestion
                if (parameterCount == 0) {
                    object = (T) constructor.newInstance();
                } else if (parameterCount == 1) {
                    Class<?> paramType = constructor.getParameterTypes()[0];
                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
                    object = (T) constructor.newInstance(dummy);
                } else {
                    object = (T) constructor.newInstance(new Object[parameterCount]);
                }
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class Issue7654 {
+    @Test
+    public void testNonStaticInnerClass() {
+        String json = "{\"issues\":[],\"summary\":{\"errors\":0,\"warnings\":0,\"messagesProcessed\":7,\"messagesAccepted\":7,\"messagesInvalid\":0}}";
+        AmzListingJsonFeedResult result = JSON.parseObject(json, AmzListingJsonFeedResult.class);
```
wenshao: **[Suggestion]** This test is inert on JDK < 25 — verified by revert-probe: it passes whether or not the `allocateInstance` fix is applied, because `Summary`'s constructor never dereferences the outer instance, so a `null` `this$0` is silently stored. — Concrete cost: if a future change regresses the `allocateInstance` path back to passing `null`, this test will not catch it on any JDK below 25. The modified `Issue1082` test does gate the fix (its constructor calls `Issue1082.this.f()`), but only for the path it exercises.

Consider adding a variant whose inner class constructor explicitly dereferences the outer instance (like `Model_1082` does), so the fix is regression-gated on all JDK versions.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI failing: Test JDK 25 on windows-latest, Test JDK 21 on windows-latest, Test Reflect JDK 25 on ubuntu-24.04, Test JDK 8 on ubuntu-24.04, Test JDK 17 on windows-latest, Test JDK 21 on ubuntu-24.04, Test JDK 25 on ubuntu-24.04, Test JDK 11 on macos-latest, Test JDK 17 on ubuntu-24.04, Test JDK 8 on macos-latest, Test JDK 11 on ubuntu-24.04, Test Reflect JDK 25 on windows-latest, Test Reflect JDK 17 on macos-latest, Test JDK 11 on windows-latest, Test Reflect JDK 21 on windows-latest, Test Reflect JDK 17 on windows-latest, Test Reflect JDK 21 on ubuntu-24.04, Test Reflect JDK 11 on windows-latest, Test Reflect JDK 11 on ubuntu-24.04, Test Reflect JDK 8 on ubuntu-24.04, Test Reflect JDK 21 on macos-latest, Test Reflect JDK 8 on windows-latest, Test JDK 17 on macos-latest, Test Reflect JDK 17 on ubuntu-24.04, Test Reflect JDK 11 on macos-latest, coverage, Test Reflect JDK 8 on macos-latest, Test JDK 21 on macos-latest, Test JDK 8 on windows-latest. Reviewed. Suggestions are inline.

core/src/test/java/com/alibaba/fastjson2/v1issues/issue_1000/Issue1082.java
wenshao: Done — the `fastjson1-compatible` `Issue1082` now asserts a successful parse (try/catch removed, `assertNotNull(m)`).

One correction on the diagnosis though: this was **not** the cause of the 29 CI failures. Every failing check reported the same 4 errors, all in the `core` module — the build failed there, so `fastjson1-compatible` never ran:

```
java.lang.IllegalAccessError: failed to access class com.alibaba.fastjson2.v1issues.issue_1200.Issue1202$Model2Test
  from class com.alibaba.fastjson2.reader.ORG_3_1_Model2
  (... Issue1202$Model2Test is in unnamed module of loader 'app';
   ORG_3_1_Model2 is in unnamed module of loader com.alibaba.fastjson2.util.DynamicClassLoader)
```

Root cause is in this PR's ASM change: `ldc` and `checkcast` on the enclosing type are access-checked against the *generated* reader, which is defined in `DynamicClassLoader`. `Issue1202$Model1Test` / `Model2Test` are package-private, so the generated bytecode cannot reference them at all. The previous `aconst_null` never triggered this because a parameter type appearing only in an `invokespecial` descriptor is not access-checked.

Fixed centrally in `ObjectReaderCreatorASM.newObject` (all three call sites have different guards, so guarding there is the safe place): emit the `UNSAFE.allocateInstance` sequence only when the enclosing type is public, otherwise keep `aconst_null` as before. The reflection paths (`ConstructorSupplier`, `ObjectReaderAdapter`) are unaffected and still allocate a real enclosing instance.

Verified on JDK 17: `Issue1202`, `Issue7654`, both `Issue1082` tests pass, and the full `core` + `fastjson1-compatible` suites are green locally.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
wenshao: Done. `paramType` is now a final field on `ConstructorSupplier`, resolved once in the constructor (`getParameterCount() == 1 ? getParameterTypes()[0] : null`). `get()` reads the field and no longer calls `getParameterTypes()`, so the per-instance `Class[]` clone is gone; the null check also replaces the repeated `getParameterCount()` call.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
wenshao: Done. Applied as suggested — explicit `parameterCount == 0` / `== 1` / `else` branches, with the `else` restoring the pre-PR `constructor.newInstance(new Object[parameterCount])` behaviour for constructors with 2+ parameters instead of failing with an unexplained `IllegalArgumentException`.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
wenshao: Done. Added `testInnerClassConstructorDereferencingOuter`: `Outer` is a public static class, `Outer.Inner`'s constructor calls `Outer.this.name()`, so a null `this$0` throws on every JDK.

Revert-probe on JDK 17, with all three production files reverted to `main`:

```
[ERROR] Issue7654.testInnerClassConstructorDereferencingOuter -- NullPointerException:
        Cannot invoke "...Issue7654$Outer.name()" because "this$0" is null
```

while `testNonStaticInnerClass` passes either way — exactly the gap you described. Both pass with the fix applied.

wenshao: Follow-up on your finding about this test being inert: adding the gating test immediately turned up that the fix itself was incomplete. The `Test Reflect` CI jobs (JDK 8/17/21/25, both Windows and ubuntu) failed on `testInnerClassConstructorDereferencingOuter`:

```
java.lang.NullPointerException: Cannot invoke "...Issue7654$Outer.name()" because "this$0" is null
  at com.alibaba.fastjson2.reader.ConstructorFunction.apply(ConstructorFunction.java:148)
  at com.alibaba.fastjson2.reader.ObjectReaderNoneDefaultConstructor.readObject(...)
```

Reproducible locally on any OS with `-Dfastjson2.creator=reflect`, so it is not a Windows issue. Under the reflect creator a non-static inner class is routed through `beanInfo.creatorConstructor` → `ObjectReaderNoneDefaultConstructor` → `ConstructorFunction`, a fourth instantiation path the earlier commits did not touch. The synthetic enclosing-instance parameter never appears in the JSON, so it fell through to `TypeUtils.getDefaultValue()` and the constructor ran with a null `this$0`.

`ConstructorFunction` now identifies that parameter (declaring class is a non-static member class and `parameters[0].getType() == declaringClass.getDeclaringClass()`) and allocates a bare enclosing instance for it, consistent with `ConstructorSupplier`, `ObjectReaderAdapter` and the ASM path.

Full suite green locally on JDK 17 in both modes — default: core 7935 + fastjson1-compatible 1346; `-Dfastjson2.creator=reflect`: same, `BUILD SUCCESS`.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
```diff
@@ -99,6 +115,22 @@ final class ConstructorFunction<T>
...
+     * constructor is the enclosing instance, which is never present in the JSON, so allocate a
+     * bare one rather than passing null.
+     */
+    private Object defaultArg(int index, Class<?> paramClass) {
+        if (index == 0 && enclosingParamType != null) {
+            try {
+                return JDKUtils.UNSAFE.allocateInstance(enclosingParamType);
+            } catch (Throwable ignored) {
+                // fall back to null below
+            }
```
wenshao, line 126: **[Suggestion]** `catch (Throwable ignored)` swallows every failure from `allocateInstance` — including `OutOfMemoryError` — and falls back to null. On JDK 25, passing null as the enclosing instance is the exact NPE this PR fixes, so under memory pressure the fix silently stops working with no diagnostic trace. — Concrete cost: an OOM in production surfaces as an NPE in the inner-class constructor, with zero indication of the real cause; the actionable signal is lost.

```suggestion
            try {
                return JDKUtils.UNSAFE.allocateInstance(enclosingParamType);
            } catch (InstantiationException ignored) {
                // fall back to null below
            }
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
```diff
@@ -0,0 +1,118 @@
...
```
wenshao: **[Suggestion]** Both tests call `JSON.parseObject(...)` directly, which routes through the ASM creator only. The `ConstructorFunction` reflection path (`enclosingParamType` detection, `defaultArg`, and the changed marker-branch condition) has no test coverage. The project's convention (`TestUtils.readerCreators()`, used by hundreds of existing tests) iterates both creators. — Concrete cost: a regression in `defaultArg` (e.g., an inverted `enclosingParamType` check) would not be caught by any test in this diff; users who disable ASM or whose classes fail ASM generation would hit the broken reflection path and get NPE on JDK 25.

```suggestion
        for (ObjectReaderCreator creator : TestUtils.readerCreators()) {
            ObjectReader<AmzListingJsonFeedResult> reader = creator.createObjectReader(AmzListingJsonFeedResult.class);
            AmzListingJsonFeedResult result = reader.readObject(new JSONReader.StringReader(json), 0);
            assertNotNull(result.getSummary());
            assertEquals(7, result.getSummary().getMessagesAccepted());
        }
```

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
wenshao: Done — narrowed to `catch (InstantiationException ignored)`, so an `OutOfMemoryError` (or any other `Error`) from `allocateInstance` now propagates instead of degrading into a null enclosing instance.

Checked the sibling call sites while I was here: `ConstructorSupplier` and `ObjectReaderAdapter` both let the failure escape into a `JSONException`, so neither has the silent-null problem this one had — `defaultArg` was the only place that fell back to null.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
wenshao: Done — both tests now iterate `TestUtils.readerCreators()`.

`testInnerClassConstructorDereferencingOuter` builds the reader for `Outer.Inner` directly from each creator, which is what actually reaches `ConstructorFunction`. `testNonStaticInnerClass` does the same against `AmzListingJsonFeedResult.Summary` — reading the outer type through a creator would not have covered it, since the nested reader is resolved lazily from the provider rather than from the creator under test.

Revert-probe on the rebased branch, with the `enclosingParamType` branch in `defaultArg` disabled:

```
[ERROR] Issue7654.testInnerClassConstructorDereferencingOuter:49 » NullPointer
        Cannot invoke "...Issue7654$Outer.name()" because "this$0" is null
  at com.alibaba.fastjson2.reader.ConstructorFunction.apply(ConstructorFunction.java:180)
  at com.alibaba.fastjson2.reader.ObjectReaderNoneDefaultConstructor.readObject(...)
```

Line 49 is the new creator loop — the preceding `JSON.parseObject` call still passes, which is exactly the gap you described.

Also rebased onto current `main`. Green locally on JDK 26: core 7959 + extension 56 + fastjson1-compatible 1355 + kotlin 36 + safemode-test 2, and core again with `-Dfastjson2.creator=reflect`.

On the three red checks from the previous run — they are not related to this PR: all three are 20-minute job timeouts inside `dokka:2.1.0-Beta:dokka @ fastjson2-kotlin`, well after the `core` tests had finished.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
```diff
@@ -27,8 +32,9 @@ public Object get() {
             if (useClassNewInstance) {
                 return objectClass.newInstance();
             } else {
-                if (constructor.getParameterCount() == 1) {
-                    return constructor.newInstance(new Object[1]);
+                if (paramType != null) {
+                    Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
+                    return constructor.newInstance(dummy);
```
wenshao: **[Critical]** `Unsafe.allocateInstance` throws `InstantiationException` for abstract enclosing classes. `ConstructorSupplier.get()`, `ObjectReaderAdapter.createInstance()` fallback, and ASM `newObject` do not catch it, while `ConstructorFunction.defaultArg` does. — Failure scenario: `JSON.parseObject("{\"value\":1}", Outer.Inner.class)` where `Outer` is abstract → `InstantiationException` from `allocateInstance` on JDK ≤ 24 (pre-PR: succeeded with null `this$0`). ASM path has no handler so the checked exception escapes raw from `JSON.parseObject`; `ConstructorSupplier` wraps in `JSONException`. Reflective creator succeeds (`defaultArg` catches and falls back to null). Probe-confirmed: the ASM creator throws, the reflective creator succeeds.

```suggestion
            if (paramType != null && !Modifier.isAbstract(paramType.getModifiers())) {
                Object dummy = JDKUtils.UNSAFE.allocateInstance(paramType);
                return constructor.newInstance(dummy);
            }
```

Mirror `defaultArg` at all three sites: catch `InstantiationException` and fall back to null. For ASM `newObject`, add `&& !Modifier.isAbstract(paramType.getModifiers())` to the condition.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
```diff
@@ -19,6 +21,9 @@ public ConstructorSupplier(Constructor constructor) {
         this.useClassNewInstance = constructor.getParameterCount() == 0
                 && Modifier.isPublic(constructor.getModifiers())
                 && Modifier.isPublic(objectClass.getModifiers());
+        this.paramType = constructor.getParameterCount() == 1
+                ? constructor.getParameterTypes()[0]
+                : null;
```
wenshao: **[Suggestion]** `paramType` is set for every single-parameter constructor, not just non-static inner class constructors. The upstream invariant (`BeanUtils.getDefaultConstructor` only returns 1-param constructors for inner classes) protects internal callers, but the code reads broader than its actual scope. `ConstructorFunction` correctly gates on `enclosingParamType != null` with an explicit inner-class check. — Concrete cost: if a future code path creates `ConstructorSupplier` with a non-inner-class 1-param constructor, it would silently pass an uninitialized object instead of null.

```suggestion
            this.paramType = constructor.getParameterCount() == 1
                    && constructor.getDeclaringClass().getDeclaringClass() != null
                    && !Modifier.isStatic(constructor.getDeclaringClass().getModifiers())
                    && constructor.getParameterTypes()[0] == constructor.getDeclaringClass().getDeclaringClass()
                    ? constructor.getParameterTypes()[0]
                    : null;
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * The inner class constructor dereferences the enclosing instance, so a {@code null}
+     * {@code this$0} throws NPE on every JDK version. Unlike {@link #testNonStaticInnerClass()},
+     * this gates the fix on JDK &lt; 25 as well.
+     */
+    @Test
+    public void testInnerClassConstructorDereferencingOuter() {
+        Outer.Inner inner = JSON.parseObject("{\"value\":123}", Outer.Inner.class);
+        assertNotNull(inner);
```
wenshao, line 44: **[Suggestion]** No test exercises the `biFunction` (2-param) or `marker`/multi-param paths in `ConstructorFunction.apply` that this diff modifies. Both test inner classes have implicit constructors with only the synthetic `this$0` parameter (1 param total). — Concrete cost: if `defaultArg(0, param0Type)` at line 188 were reverted to `TypeUtils.getDefaultValue`, no test would fail. Add a test with a non-static inner class that has an explicit constructor parameter, e.g. `public Inner2(int x) { Outer2.this.toString(); this.x = x; }`, via all `readerCreators()`.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
```diff
@@ -706,7 +706,17 @@ private static void newObject(MethodWriter mw, String TYPE_OBJECT, Constructor d
...
+            // which lives in DynamicClassLoader; a non-public enclosing type fails with
+            // IllegalAccessError there, so keep passing null for it as before.
+            if (Modifier.isPublic(paramType.getModifiers())) {
+                mw.getstatic(TYPE_UNSAFE_UTILS, "UNSAFE", "Lsun/misc/Unsafe;");
+                mw.visitLdcInsn(paramType);
+                mw.invokevirtual("sun/misc/Unsafe", "allocateInstance", "(Ljava/lang/Class;)Ljava/lang/Object;");
+                mw.checkcast(ASMUtils.type(paramType));
+            } else {
+                mw.aconst_null();
+            }
```
wenshao, line 743: **[Suggestion]** The non-public enclosing class fallback (`aconst_null`) is untested and on JDK 25 still produces the exact NPE this PR fixes. Both test enclosing classes are `public`, so this `else` branch is never taken. — Concrete cost: package-private enclosing class + non-static inner class via ASM reader on JDK 25 → NPE; same input with reflective creator succeeds. Creator-divergent behavior, untested. Consider adding a test with a package-private enclosing class.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
```diff
@@ -57,6 +64,15 @@ final class ConstructorFunction<T>
...
         }
 
+        Class declaringClass = constructor.getDeclaringClass();
+        Class enclosingClass = declaringClass.getDeclaringClass();
+        this.enclosingParamType = enclosingClass != null
+                && !Modifier.isStatic(declaringClass.getModifiers())
+                && parameters.length > 0
+                && parameters[0].getType() == enclosingClass
+                ? enclosingClass
+                : null;
```
wenshao: **[Suggestion]** This inner-class detection logic duplicates `BeanUtils.isNoneStaticMemberClass` (already used by `FieldReader` in the same package). — Concrete cost: if a future edge case requires refining the heuristic (e.g. local classes where `getDeclaringClass()` returns null but `getEnclosingClass()` does not), the fix must be applied independently to both sites, risking inconsistent detection. Consider extracting a shared helper.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR. ⚠️ This run could not certify that any of this diff was reviewed. Suggestions are inline. Not reviewed: the entire diff — no agent reported covering it; nobody read it. Not reviewed: the linked-issue fidelity pass, the line-by-line correctness pass, the security pass, the reuse and duplication pass, the altitude and abstraction pass, the consistency and clarity pass, the performance pass, the test-coverage pass, the open-ended audit (attacker mindset), the open-ended audit (oncall mindset), the open-ended audit (maintainer mindset), the removed-behavior audit, the cross-file consistency pass, the build-and-test check — its prompt was built, but no agent on record was launched with it. Not reviewed: verification and reverse audit — both prompts were built, but no agent was launched with either — the posted findings cannot be counted as verified, and the pass that hunts what the rest of the review missed cannot be certified.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
```diff
@@ -99,6 +115,22 @@ final class ConstructorFunction<T>
...
+     * constructor is the enclosing instance, which is never present in the JSON, so allocate a
+     * bare one rather than passing null.
+     */
+    private Object defaultArg(int index, Class<?> paramClass) {
+        if (index == 0 && enclosingParamType != null) {
+            try {
+                return JDKUtils.UNSAFE.allocateInstance(enclosingParamType);
+            } catch (InstantiationException ignored) {
+                // the enclosing type cannot be allocated, fall back to the default value below
+            }
```
wenshao, line 126: **[Suggestion]** The `defaultArg` fallback is self-defeating for the JDK 25 goal: it fires exactly when the enclosing type cannot be allocated (an **abstract** outer class — `Unsafe.allocateInstance` throws `InstantiationException`), which is precisely the case that needs a non-null `this$0`. It then returns `TypeUtils.getDefaultValue(...)` = null, so JDK 25's `requireNonNull(this$0)` still throws. — Concrete cost: for `public abstract class Outer { public class Inner {...} }`, `JSON.parseObject(json, Outer.Inner.class)` still fails on JDK 25 exactly as before this PR; the fix is incomplete for abstract outers, and the new tests cover only concrete outers.

Consider throwing a descriptive `JSONException` (e.g. `"cannot instantiate enclosing class X of inner class Y"`) instead of silently falling back to null, or documenting the limitation here and adding a test that pins the current behaviour for abstract outers.

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
```diff
@@ -193,8 +225,8 @@ public T apply(Map<Long, Object> values) {
                     args[i] = arg;
                 } else {
                     flag |= (1 << i);
-                    if (paramClass.isPrimitive()) {
-                        args[i] = TypeUtils.getDefaultValue(paramClass);
+                    if (paramClass.isPrimitive() || (i == 0 && enclosingParamType != null)) {
+                        args[i] = defaultArg(i, paramClass);
```
wenshao, line 226: **[Suggestion]** The new `|| (i == 0 && enclosingParamType != null)` guard in this general-loop branch is not exercised by any test. Both inner classes in `Issue7654` (`Summary`, `Inner`) have no-arg constructors that compile to 1-parameter constructors (just `this$0`), so they always hit the `parameters.length == 1` fast path and never reach this loop. — Concrete cost: deleting `|| (i == 0 && enclosingParamType != null)` here would leave `args[0]` null for a 3+ parameter inner-class constructor (the enclosing param is always absent from the JSON) and reintroduce the NPE on JDK 25; no existing test distinguishes that mutant.

Add a test with an inner class that has two or more explicit constructor parameters, e.g. `public class Outer2 { public class Inner2 { public Inner2(int a, String b) { Outer2.this.toString(); } } }`, parsed across `TestUtils.readerCreators()`.

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
wenshao: Fixed in 326b5104e. All three non-reflective paths now treat a non-allocatable (e.g. abstract) enclosing class the same way `ConstructorFunction.defaultArg` already did — catch `InstantiationException` and pass `null` (the pre-JDK-25 behavior):

- `ConstructorSupplier.get()` and `ObjectReaderAdapter.createInstance()`: wrapped `UNSAFE.allocateInstance(paramType)` in try/catch, falling back to `null` on `InstantiationException`.
- `ObjectReaderCreatorASM.newObject`: added `!Modifier.isAbstract(paramType.getModifiers())` to the inline-allocation condition, so an abstract enclosing type emits `aconst_null` instead of an `allocateInstance` call that would throw `InstantiationException` at runtime (the raw checked-exception leak you described).

This restores the pre-PR behavior for inner classes of abstract outers on every JDK and removes the creator divergence. Verified: `testAbstractEnclosingClass` (added, iterating `TestUtils.readerCreators()`) now succeeds on both creators — before this change the ASM creator threw a raw `InstantiationException` while the reflective creator succeeded. Full core suite green on JDK 8/17/25 (default + `-Dfastjson2.creator=reflect`).

<details><summary>中文</summary>

已在 326b5104e 修复。三个非反射路径现在对无法分配（如 abstract）的外部类的处理与 `ConstructorFunction.defaultArg` 一致——捕获 `InstantiationException` 并回退为传 `null`（JDK 25 之前的行为）：

- `ConstructorSupplier.get()` 与 `ObjectReaderAdapter.createInstance()`：对 `UNSAFE.allocateInstance(paramType)` 加了 try/catch，遇到 `InstantiationException` 回退为 `null`。
- `ObjectReaderCreatorASM.newObject`：在内联分配条件中加了 `!Modifier.isAbstract(paramType.getModifiers())`，abstract 外部类改为生成 `aconst_null`，不再在运行时抛出 `InstantiationException`（即你指出的裸 checked 异常泄漏）。

这恢复了 JDK 25 之前 abstract 外部类内部类的行为，并消除了 creator 间的不一致。已验证：新增的 `testAbstractEnclosingClass`（遍历 `TestUtils.readerCreators()`）在两种 creator 下都成功——改动前 ASM creator 会抛裸 `InstantiationException` 而反射 creator 成功。core 全量测试在 JDK 8/17/25（默认 + reflect 模式）全绿。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorSupplier.java
wenshao: Done in 326b5104e. `paramType` is now gated through the new shared helper `BeanUtils.getEnclosingInstanceParamType(constructor)`, which returns non-null only when the constructor belongs to a non-static member class and its first parameter is exactly the enclosing class (`enclosingClass != null && !Modifier.isStatic(declaringClass) && parameterTypes[0] == enclosingClass`). For any other single-parameter constructor `paramType` is `null`, and `get()` falls back to the pre-PR `constructor.newInstance(new Object[1])`.

The `!Modifier.isStatic` check also guards a latent `getDefaultConstructor` quirk: a *static* nested class with a single-parameter constructor that happens to take the outer type would otherwise be misdetected as an inner class and get an `allocateInstance`'d dummy.

<details><summary>中文</summary>

已在 326b5104e 处理。`paramType` 现在通过新的共享辅助方法 `BeanUtils.getEnclosingInstanceParamType(constructor)` 门控：仅当构造器属于非静态成员类、且其第一个参数恰好是外部类时（`enclosingClass != null && !Modifier.isStatic(declaringClass) && parameterTypes[0] == enclosingClass`）才返回非 null。其余单参构造器的 `paramType` 为 `null`，`get()` 回退到 PR 之前的 `constructor.newInstance(new Object[1])`。

`!Modifier.isStatic` 检查还规避了 `getDefaultConstructor` 的一个潜在问题：*静态* 嵌套类若有一个恰好接收外部类类型的单参构造器，否则会被误判为内部类并得到一个 `allocateInstance` 出来的哑实例。

</details>

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7654.java
wenshao: Done in 3e4929dd0. Added two tests, both iterating `TestUtils.readerCreators()`:

- `testInnerClassWithExplicitConstructorParameter` — `InnerWithOneParam(int)` (2 total params incl. `this$0`) exercises the `BiFunction` fast path.
- `testInnerClassWithTwoExplicitConstructorParameters` — `InnerWithTwoParams(int, String)` (3 total params) exercises the general argument loop.

Both constructors dereference the enclosing instance (`Outer2.this.name()`), so a null `this$0` throws on every JDK regardless of the `requireNonNull` codegen. Revert-probe: reverting the production changes to the pre-PR base makes both fail (`NullPointerException` / `invoke constructor error`), confirming they gate the fix.

<details><summary>中文</summary>

已在 3e4929dd0 完成。新增两个测试，均遍历 `TestUtils.readerCreators()`：

- `testInnerClassWithExplicitConstructorParameter`——`InnerWithOneParam(int)`（含 `this$0` 共 2 参）覆盖 `BiFunction` 快路径。
- `testInnerClassWithTwoExplicitConstructorParameters`——`InnerWithTwoParams(int, String)`（共 3 参）覆盖通用参数循环。

两个构造器都解引用外部实例（`Outer2.this.name()`），因此 `this$0` 为 null 时在任何 JDK 上都会抛出，与 `requireNonNull` 代码生成无关。还原探针：把主代码还原到 PR 前基线，两个测试都失败（`NullPointerException` / `invoke constructor error`），确认它们能门控该修复。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderCreatorASM.java
wenshao: Addressed in 3e4929dd0 — and closed the JDK 25 hole rather than only pinning it. A generated reader can't `ldc`/`checkcast` a non-public enclosing type (it lives in `DynamicClassLoader`), so inline `allocateInstance` is impossible there. Instead of keeping `aconst_null` (which NPEs on JDK 25-compiled classes), the ASM creator now detects this via `enclosingTypeVisible(...)` and delegates instance creation to the reflective `ConstructorSupplier`/`ConstructorFunction`, which allocate the enclosing instance without referencing it from generated bytecode:

- `genCreateObject` and the generated `createInstance` override both skip the inline path when the enclosing type is non-public.
- This also fixes the creator-constructor case: an inner class with an explicit constructor parameter previously got an ASM-generated reader that passed `null` for `this$0`; such readers are now delegated to the reflective path too.

Added `testPackagePrivateEnclosingClass` (package-private outer, constructor dereferences `this$0`, iterates both creators). Revert-probe confirms it fails without the fix; `Issue1202` (package-private `@Nested` outers) still passes on JDK 25.

<details><summary>中文</summary>

已在 3e4929dd0 处理——不只是用测试固化，而是直接堵上了 JDK 25 的洞。生成的 reader 无法对非 public 外部类做 `ldc`/`checkcast`（它定义在 `DynamicClassLoader` 中），因此无法在那里内联 `allocateInstance`。ASM creator 不再保留 `aconst_null`（那在 JDK 25 编译的类上会 NPE），而是通过 `enclosingTypeVisible(...)` 检测到这种情况，把实例创建委托给反射的 `ConstructorSupplier`/`ConstructorFunction`——它们分配外部实例时无需在生成字节码中引用该类：

- 当外部类非 public 时，`genCreateObject` 与生成的 `createInstance` 覆写都跳过内联路径。
- 这同时修复了 creator-constructor 场景：带显式构造参数的内部类此前会得到一个对 `this$0` 传 `null` 的 ASM 生成 reader；现在这类 reader 也委托到反射路径。

新增 `testPackagePrivateEnclosingClass`（package-private 外部类、构造器解引用 `this$0`、遍历两种 creator）。还原探针确认没有该修复时测试失败；`Issue1202`（package-private 的 `@Nested` 外部类）在 JDK 25 上仍然通过。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
wenshao: Done in 326b5104e / 3e4929dd0. Extracted `BeanUtils.getEnclosingInstanceParamType(Constructor)` and replaced all four copies of the detection (`ConstructorFunction`, `ConstructorSupplier`, `ObjectReaderAdapter`, and the ASM creator's delegation guard) with it.

I kept it separate from `isNoneStaticMemberClass` deliberately: that one takes an `(objectClass, memberClass)` pair and inspects the member's *first declared* constructor (with primitive/String/List filtering and its own cache), whereas the readers here already hold the *specific* constructor to test and need its first parameter checked. Reusing it would have required re-deriving the constructor and coupling the two heuristics. The shared rule now lives in exactly one place, so a future refinement (e.g. local classes where `getDeclaringClass()` is null but `getEnclosingClass()` is not) only needs one edit.

<details><summary>中文</summary>

已在 326b5104e / 3e4929dd0 完成。抽取了 `BeanUtils.getEnclosingInstanceParamType(Constructor)`，并把四处重复的检测逻辑（`ConstructorFunction`、`ConstructorSupplier`、`ObjectReaderAdapter`，以及 ASM creator 的委托守卫）统一替换为它。

我有意没有复用 `isNoneStaticMemberClass`：那个方法接收 `(objectClass, memberClass)` 对，并检查成员类的*第一个声明的*构造器（还带 primitive/String/List 过滤和自己的缓存）；而这里的 reader 已经持有*具体的*构造器，只需检查它的第一个参数。复用它需要重新推导构造器、并把两种启发式耦合在一起。现在这条共享规则只存在于一处，未来若要细化（例如 `getDeclaringClass()` 为 null 但 `getEnclosingClass()` 非 null 的局部类）只需改一处。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
wenshao: Kept the null fallback deliberately (documented + pinned rather than throwing) — reasoning:

- An abstract enclosing class cannot be fabricated at all (`allocateInstance` throws, and there is no concrete instance to use). So on JDK 25+ an inner class of an abstract outer is inherently undeserializable; no code path can fix it.
- On JDK ≤ 24 the pre-PR behavior was "pass `null`, succeed as long as the constructor doesn't dereference it". Throwing a `JSONException` instead would be a hard regression for code that works today, in exchange for no capability on 25+.

So I chose the backward-compatible option you offered: document + pin. The limitation is now documented in the `defaultArg` javadoc and the `newObject` comment, and `testAbstractEnclosingClass` pins the pre-JDK-25 success across both creators (this module compiles at release 8, so the constructor accepts the null `this$0`). Inner classes of abstract outers compiled by JDK 25+ will still fail — that's unavoidable and now explicit rather than a silent `InstantiationException` leak.

<details><summary>中文</summary>

有意保留 null 回退（采用你给的"文档 + 测试固化"方案，而不是抛异常），理由：

- abstract 外部类根本无法被构造（`allocateInstance` 会抛异常，也没有可用的具体实例）。因此 JDK 25+ 上 abstract 外部类的内部类本质上无法反序列化，任何代码路径都修不了。
- JDK ≤ 24 上 PR 之前的行为是"传 `null`，只要构造器不解引用它就成功"。改成抛 `JSONException` 对现在能跑的代码是硬回归，而在 25+ 上又换不来任何能力。

所以我选择了你提出的向后兼容方案：文档 + 测试固化。该限制已写入 `defaultArg` 的 javadoc 和 `newObject` 的注释，`testAbstractEnclosingClass` 在两种 creator 下固化 JDK 25 之前的成功行为（本模块按 release 8 编译，构造器接受 null 的 `this$0`）。JDK 25+ 编译的 abstract 外部类内部类仍会失败——这不可避免，但现在是显式的，而不是裸的 `InstantiationException` 泄漏。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ConstructorFunction.java
wenshao: Added `testInnerClassWithTwoExplicitConstructorParameters` (3e4929dd0), which is exactly the suggested shape — an inner class with two explicit constructor parameters that dereferences the enclosing instance, parsed across `TestUtils.readerCreators()`. It exercises the general (non-fast-path) argument loop with the enclosing instance allocated via `defaultArg`. Revert-probe confirms it fails without the fix.

One clarification on the branch itself: the `marker` loop is only reachable for Kotlin classes (`beanInfo.markerConstructor` is set solely by `KotlinUtils`), so no pure-Java test can reach line 226's guard. While attempting a Kotlin `inner class` test I hit a *pre-existing, separate* bug: Kotlin inner classes with default parameters get their default-mask bits shifted by one because the synthetic `this$0` occupies parameter index 0 — fields come back as their defaults regardless of the JSON. I reproduced this identically on `main` in a probe worktree, so it predates this PR and is orthogonal to the enclosing-instance allocation (which the guard does handle consistently with the other paths). I've left that mask-alignment issue out of this PR; happy to file a follow-up.

<details><summary>中文</summary>

已新增 `testInnerClassWithTwoExplicitConstructorParameters`（3e4929dd0），正是建议的形态——带两个显式构造参数、且解引用外部实例的内部类，遍历 `TestUtils.readerCreators()` 解析。它覆盖了通用（非快路径）参数循环，其中外部实例通过 `defaultArg` 分配。还原探针确认没有修复时测试失败。

关于该分支本身的一点澄清：`marker` 循环只有 Kotlin 类才能到达（`beanInfo.markerConstructor` 仅由 `KotlinUtils` 设置），因此纯 Java 测试无法触及 line 226 的守卫。在尝试写 Kotlin `inner class` 测试时，我碰到一个*既有的、独立的* bug：带默认参数的 Kotlin 内部类，其默认掩码位会整体偏移一位（因为合成的 `this$0` 占据了参数索引 0），导致字段无论 JSON 如何都取默认值。我已在探针 worktree 中于 `main` 上原样复现，确认它早于本 PR，且与外部实例分配无关（该守卫对其他路径的处理是一致的）。我把这个掩码对齐问题留在本 PR 之外；如需要可以建一个 follow-up issue。

</details>

PR #7700, opened 2026-07-27
test: cover both strBuf sizing paths in readString (#7685)

### LINKED ISSUE #7685: [BUG]fastjson2.0.61版本,将字符串转成对象，如果数组型的字符串中含有\n回车，且字符串比较长，会出现数据越界异常。
### 问题描述
fastjson2,将字符串转成对象，如果数组型的字符串中含有\n回车，且字符串比较长，会出现数据越界异常。

### 环境信息

 - OS信息： windows和linux
 - JDK信息：jdk-21.0.7
 - 版本信息：Fastjson2 2.0.61，其他版本没有这个问题，比如Fastjson2 2.0.60，Fastjson2 2.0.53，Fastjson2 2.0.54
 

### 重现步骤
1. fastjson2,将字符串转成对象，如果数组型的字符串中含有\n回车，且字符串比较长，会出现数据越界异常。
2. 可以执行运行如下代码
```java
String str= "{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"s Howard), dedicated to dinosaur conservation, was invited by Benjamin Lockwood. Benjamin and his subordinate Eli Mills (played by Rafi Spo) hope that she and relevant technicians will go to Isla Nublar to save the dinosaurs there and relocate them to a new habitat. Claire, who is infected, manages to bring Owen (played by Chris Pratt) along to go, while Owen hopes to rescue the velociraptor Blue that he has personally tamed. When they set foot on this isolated island, they found that things were not that simple.\\nVolcanoes are restless and ready to erupt, while ugly conspiracies are quietly brewing amidst chaos.\"}]}";
JSON.parseObject(str, Object.class);
```

### 期待的正确结果
*对您期望发生的结果进行清晰简洁的描述。*

### 相关日志输出
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 516 out of bounds for char[512]
	at java.base/java.lang.System.arraycopy(Native Method)
	at com.alibaba.fastjson2.JSONReaderUTF16.readString(JSONReaderUTF16.java:3259)
	at com.alibaba.fastjson2.JSONReader.readObject(JSONReader.java:3777)
	at com.alibaba.fastjson2.JSONReader.readArray(JSONReader.java:4118)
	at com.alibaba.fastjson2.reader.ObjectReaderImplObject.readObject(ObjectReaderImplObject.java:209)
	at com.alibaba.fastjson2.JSON.parseObject(JSON.java:992)
	at Test.main(Test.java:2222)

#### 附加信息

wenshao:
## Summary

Regression tests for #7685 — `ArrayIndexOutOfBoundsException` in `readString` when a string containing a `\n` escape is long enough to outgrow `strBuf`.

**Test-only, no production change.** The bug itself was already fixed on `main`: `bdfe92f76` (#3989) added the `stroff > strBuf.length` guard before the pre-copy, but that fix did not make it into the 2.0.61 release, which is what the reporter hit. What was missing was coverage — the tests here pin the behaviour of both `strBuf` sizing paths in `JSONReaderUTF16` / `JSONReaderUTF8`:

```java
char[] strBuf = this.strBuf;
if (strBuf == null) {
    strBuf = new char[stroff + 512];      // initial allocation
    this.strBuf = strBuf;
} else if (stroff > strBuf.length) {      // growth on reuse
    ...
}
System.arraycopy(chars, start, strBuf, 0, stroff);
```

## Tests

| Test | Covers |
|---|---|
| `test` | The exact payload from the issue report |
| `testVariousLengths` | Growth path via `JSONReaderUTF16`, sweeping lengths 1–600 around the 512 boundary |
| `testVariousLengthsUTF8` | Same payloads via `JSONReaderUTF8`, plus one at 100 000 to clear any buffer a preceding test left in the process-wide cache |
| `testLongPrefixBeforeFirstEscape` | Initial allocation — the escape sits in the first string value, behind 500–4096 already-scanned characters |
| `testVariousLengthsJSONB` | Round-trip sanity check only; documented as *not* covering `strBuf` |

Assertions compare full string content, not just length, so a substitution in the escape table cannot slip through.

## Verification

Each claim was revert-probed rather than assumed.

**Growth path** — restoring `new char[stroff + 512]` to `new char[512]` in both readers:

```
[ERROR] Issue7685.testLongPrefixBeforeFirstEscape:69
        java.lang.ArrayIndexOutOfBoundsException:
        arraycopy: last destination index 600 out of bounds for char[512]
  at com.alibaba.fastjson2.JSONReaderUTF16.readString(JSONReaderUTF16.java:3261)
```

**Content assertions** — mutating `'n', '\n'` to `'n', '\r'` in the `char1_escaped` table (`JSONReader.java:102`) fails 3 of the 5 tests. With length-only assertions the same mutation passed all of them.

## Notes

Two things worth recording, since they are what makes the tests work at all:

- A plain-ASCII payload never reaches `JSONReaderUTF16`. `JSON.parseObject(String)` routes a LATIN1-coded string to the ASCII/UTF-8 reader, so the value has to be prefixed with a multi-byte character to exercise the UTF-16 reader. `JSONReaderUTF16.strBuf` is a per-instance field and therefore null on every parse, which is what makes the initial-allocation path reachable from a test.
- `JSONReaderUTF8` takes `strBuf` from a process-wide cache, so its initial allocation cannot be forced from a test in a warm JVM. That branch stays uncovered; it is not claimed otherwise.

Full suites green locally on JDK 26 — core 7963, extension 56, fastjson1-compatible 1355, kotlin 36, safemode-test 2 — and core again with `-Dfastjson2.creator=reflect`.

Closes #7685

<details>
<summary>中文说明</summary>

## 概述

#7685 的回归测试 —— 含 `\n` 转义的字符串长到超出 `strBuf` 时，`readString` 抛 `ArrayIndexOutOfBoundsException`。

**纯测试，不改主代码。** 缺陷本身在 `main` 上已经修好：`bdfe92f76`（#3989）在 pre-copy 之前加了 `stroff > strBuf.length` 判断，但该修复没进 2.0.61 发布版，正是报告者踩到的。缺的是覆盖 —— 这里的测试固定住 `JSONReaderUTF16` / `JSONReaderUTF8` 中 `strBuf` 两条尺寸路径的行为：

```java
char[] strBuf = this.strBuf;
if (strBuf == null) {
    strBuf = new char[stroff + 512];      // 首次分配
    this.strBuf = strBuf;
} else if (stroff > strBuf.length) {      // 复用时扩容
    ...
}
System.arraycopy(chars, start, strBuf, 0, stroff);
```

## 测试

| 测试 | 覆盖 |
|---|---|
| `test` | issue 中原样的 payload |
| `testVariousLengths` | 经 `JSONReaderUTF16` 的扩容路径，长度 1–600 扫过 512 边界 |
| `testVariousLengthsUTF8` | 同样的 payload 经 `JSONReaderUTF8`，另加一个 100 000 的用例，越过前序测试可能留在进程级缓存里的 buffer |
| `testLongPrefixBeforeFirstEscape` | 首次分配 —— 转义位于第一个字符串值中，前面有 500–4096 个已扫描字符 |
| `testVariousLengthsJSONB` | 仅 round-trip sanity check；已注明**不**覆盖 `strBuf` |

断言比较字符串完整内容而非仅长度，转义表里的字符替换无法蒙混过关。

## 验证

每一条结论都做了 revert-probe，不是推断。

**扩容路径** —— 把两个 reader 的 `new char[stroff + 512]` 还原为 `new char[512]`：

```
[ERROR] Issue7685.testLongPrefixBeforeFirstEscape:69
        java.lang.ArrayIndexOutOfBoundsException:
        arraycopy: last destination index 600 out of bounds for char[512]
  at com.alibaba.fastjson2.JSONReaderUTF16.readString(JSONReaderUTF16.java:3261)
```

**内容断言** —— 把 `char1_escaped` 表（`JSONReader.java:102`）中的 `'n', '\n'` 改成 `'n', '\r'`，5 个测试挂 3 个。若只断言长度，同样的变异全部通过。

## 备注

两点值得记录，因为它们决定了测试能否成立：

- 纯 ASCII 的 payload 根本到不了 `JSONReaderUTF16`。`JSON.parseObject(String)` 对 LATIN1 编码的字符串走 ASCII/UTF-8 reader，因此值前面必须加多字节字符才能触发 UTF-16 reader。`JSONReaderUTF16.strBuf` 是实例字段，每次解析都为 null，这正是首次分配路径可被测试触达的原因。
- `JSONReaderUTF8` 的 `strBuf` 取自进程级缓存，在热 JVM 中无法从测试里强制其首次分配。该分支保持未覆盖，不作相反声明。

本地 JDK 26 全量通过 —— core 7963、extension 56、fastjson1-compatible 1355、kotlin 36、safemode-test 2；`-Dfastjson2.creator=reflect` 下 core 同样通过。

</details>

### DIFF
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
@@ -1,0 +1,41 @@
+package com.alibaba.fastjson2.issues_7000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class Issue7685 {
+    @Test
+    public void test() {
+        String str = "{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"s Howard), dedicated to dinosaur conservation, was invited by Benjamin Lockwood. Benjamin and his subordinate Eli Mills (played by Rafi Spo) hope that she and relevant technicians will go to Isla Nublar to save the dinosaurs there and relocate them to a new habitat. Claire, who is infected, manages to bring Owen (played by Chris Pratt) along to go, while Owen hopes to rescue the velociraptor Blue that he has personally tamed. When they set foot on this isolated island, they found that things were not that simple.\\nVolcanoes are restless and ready to erupt, while ugly conspiracies are quietly brewing amidst chaos.\"}]}";
+        JSONObject result = JSON.parseObject(str);
+        assertNotNull(result);
+        assertEquals(2, result.getJSONArray("Test").size());
+    }
+
+    @Test
+    public void testVariousLengths() {
+        for (int len = 1; len <= 600; len++) {
+            StringBuilder sb = new StringBuilder();
+            sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
+            for (int i = 0; i < len; i++) {
+                sb.append('a');
+            }
+            sb.append("\\n");
+            for (int i = 0; i < len; i++) {
+                sb.append('b');
+            }
+            sb.append("\"}]}");
+
+            JSONObject result = JSON.parseObject(sb.toString());
+            assertNotNull(result);
+            String message = result.getJSONArray("Test")
+                    .getJSONObject(1)
+                    .getString("message");
+            assertEquals(len * 2 + 1, message.length());
+        }
+    }
+}
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
@@ -1,8 +1,11 @@
 package com.alibaba.fastjson2.issues_7000;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONB;
 import com.alibaba.fastjson2.JSONObject;
 import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -36,6 +39,64 @@ public class Issue7685 {
                     .getJSONObject(1)
                     .getString("message");
             assertEquals(len * 2 + 1, message.length());
+        }
+    }
+
+    /**
+     * Same payloads as {@link #testVariousLengths()}, but parsed from UTF-8 bytes so the
+     * escape handling in JSONReaderUTF8.readString is exercised as well — that reader got
+     * the same strBuf growth fix and is not reached through the String entry point, which
+     * routes multi-byte input to JSONReaderUTF16.
+     */
+    @Test
+    public void testVariousLengthsUTF8() {
+        for (int len = 1; len <= 600; len++) {
+            assertMessageRoundTripUTF8(len);
+        }
+        // strBuf is pooled process-wide, so a preceding test may leave a buffer that already
+        // covers the lengths above; this one is past any capacity another test can have grown it to.
+        assertMessageRoundTripUTF8(100_000);
+    }
+
+    private void assertMessageRoundTripUTF8(int len) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
+        for (int i = 0; i < len; i++) {
+            sb.append('a');
+        }
+        sb.append("\\n");
+        for (int i = 0; i < len; i++) {
+            sb.append('b');
+        }
+        sb.append("\"}]}");
+
+        byte[] utf8 = sb.toString().getBytes(StandardCharsets.UTF_8);
+        JSONObject result = JSON.parseObject(utf8);
+        assertNotNull(result);
+        String message = result.getJSONArray("Test")
+                .getJSONObject(1)
+                .getString("message");
+        assertEquals(len * 2 + 1, message.length());
+    }
+
+    @Test
+    public void testVariousLengthsJSONB() {
+        for (int len = 1; len <= 600; len++) {
+            StringBuilder sb = new StringBuilder();
+            sb.append("中\n");
+            for (int i = 0; i < len; i++) {
+                sb.append('a');
+            }
+            sb.append('\n');
+            for (int i = 0; i < len; i++) {
+                sb.append('b');
+            }
+            String message = sb.toString();
+
+            JSONObject object = new JSONObject().fluentPut("message", message);
+            JSONObject parsed = (JSONObject) JSONB.parse(JSONB.toBytes(object));
+            assertNotNull(parsed);
+            assertEquals(message, parsed.getString("message"));
         }
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
@@ -22,23 +22,9 @@ public class Issue7685 {
     @Test
     public void testVariousLengths() {
         for (int len = 1; len <= 600; len++) {
-            StringBuilder sb = new StringBuilder();
-            sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
-            for (int i = 0; i < len; i++) {
-                sb.append('a');
-            }
-            sb.append("\\n");
-            for (int i = 0; i < len; i++) {
-                sb.append('b');
-            }
-            sb.append("\"}]}");
-
-            JSONObject result = JSON.parseObject(sb.toString());
+            JSONObject result = JSON.parseObject(payload(len));
             assertNotNull(result);
-            String message = result.getJSONArray("Test")
-                    .getJSONObject(1)
-                    .getString("message");
-            assertEquals(len * 2 + 1, message.length());
+            assertEquals(expectedMessage(len), messageOf(result));
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag("regression")
+@Tag("jsonb")
+public class Issue7669 {
+    @Test
+    public void testBigIntDeclaredLengthOOM() {
+        // 0xBB = BC_BIGINT, 0x48 = BC_INT32, 0x7FFFFFFF = Integer.MAX_VALUE
```
wenshao: **[Suggestion]** Test coverage gaps — only `readAny()` path exercised — Failure scenario: a future change weakens the bounds check in `readStringTypeNotMatch()`, `readNumber()`, or `readBigInteger()` and the existing test stays green because it goes through `JSONB.parse()` → `readAny()`, not those methods.

Consider adding malicious-input tests targeting `readString()`, `readNumber()`, and `readBigInteger()` specifically, using the same crafted payload pattern (BC_BIGINT + INT32(MAX_VALUE) and negative length) routed through each method.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class Issue7685 {
+    @Test
+    public void test() {
+        String str = "{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"s Howard), dedicated to dinosaur conservation, was invited by Benjamin Lockwood. Benjamin and his subordinate Eli Mills (played by Rafi Spo) hope that she and relevant technicians will go to Isla Nublar to save the dinosaurs there and relocate them to a new habitat. Claire, who is infected, manages to bring Owen (played by Chris Pratt) along to go, while Owen hopes to rescue the velociraptor Blue that he has personally tamed. When they set foot on this isolated island, they found that things were not that simple.\\nVolcanoes are restless and ready to erupt, while ugly conspiracies are quietly brewing amidst chaos.\"}]}";
+        JSONObject result = JSON.parseObject(str);
```
wenshao, line 17: **[Suggestion]** Test does not exercise JSONB code paths — Failure scenario: a future regression in `JSONReaderJSONB.readString()` (the JSONB binary path modified by this PR) would not be caught by this test, since it uses `JSON.parseObject()` which routes through the JSON text parser. The efficacy probe confirms this test passes with the source changes reverted.

Consider adding a JSONB-specific variant using `JSONB.parse()` / `JSONB.toBytes()` with multi-byte characters and `\n` escapes at various lengths to cover the JSONB `readString` path.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR. Reviewed. Suggestions are inline. 2 Suggestion-level finding(s) could not be anchored to a changed line and were dropped; nothing further to act on here. **[Critical]** JSONReaderJSONB.java:4485 — BC_BINARY case in readBigInteger0() has the same unguarded readInt32Value() → new byte[len] → System.arraycopy pattern that this PR fixes in four BC_BIGINT cases. No bounds check present. A crafted JSONB payload reaching readBigInteger() (called from 13+ sites) via BC_BINARY type triggers Integer.MAX_VALUE allocation → OOM DoS, the same attack vector as Issue7669.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
wenshao: Not applicable on this branch: `JSONReaderJSONB.java` here is byte-identical to `main` (`git diff main HEAD -- core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java` is empty). The hunk only shows up in the PR diff because the branch predates the merge of #7696, which landed these same `BC_BIGINT` guards on `main`; the merge-base diff therefore still attributes them to this PR.

The follow-ups belong to the PRs that actually own the JSONB work: the `checkBigintLen` helper extraction and the typed-path `Issue7669` tests are on #7698 (already applied there), and the missing `BC_BINARY` sibling is on #7705. Nothing to change here — I'll rebase this branch onto `main` so the diff stops showing unrelated files.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
wenshao: Partly done, with a correction on the premise.

This PR does not modify the JSONB `readString` path — `JSONReaderJSONB.java` on this branch is byte-identical to `main` (the hunk in the diff comes from the branch predating the #7696 merge). And the #7685 failure mode does not exist on the JSONB path at all: JSONB strings are length-prefixed and `JSONReaderJSONB.readString()` does no escape processing, so there is no `strBuf` accumulation to overflow. The fix for this bug (bdfe92f76) touched `JSONReaderUTF16` and `JSONReaderUTF8` only.

I added `testVariousLengthsJSONB` anyway as round-trip coverage for the same multi-byte + newline content, but the real gap your finding points at is the **UTF-8 text** reader: `JSON.parseObject(String)` routes multi-byte input to `JSONReaderUTF16`, so `JSONReaderUTF8.readString` — which received the same `strBuf` growth fix — was never exercised. Added `testVariousLengthsUTF8`, parsing the same payloads from `getBytes(UTF_8)`.

First attempt at that test passed even with the fix reverted: `strBuf` is pooled process-wide via `CACHE_ITEMS`, and the preceding UTF-16 test leaves a buffer already large enough. The test now ends with a length past any capacity another test can have grown it to. Revert-probe with the `JSONReaderUTF8` hunk restored to pre-fix:

```
[ERROR] Issue7685.testVariousLengthsUTF8:58->assertMessageRoundTripUTF8:74
        ArrayIndexOutOfBoundsException: Index 1728 out of bounds for length 1728
```

All 4 tests pass with the fix in place.

wenshao (commented):
Reviewed. Unresolved, please confirm: **[Critical]** JSONReaderJSONB.java readBigInteger0() BC_BINARY case (~line 4483): prior review's Critical re-checked — the unguarded readInt32Value() -> new byte[len] -> System.arraycopy OOM-DoS pattern IS present at the reviewed commit, but it lives in code this test-only PR does not modify (unrelated to issue #7685) and was independently fixed in main by 373cbe728 (#7705); whether it should gate this regression-test PR is undecidable from this diff.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
```diff
@@ -0,0 +1,102 @@
...
+        StringBuilder sb = new StringBuilder();
+        sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
+        for (int i = 0; i < len; i++) {
+            sb.append('a');
+        }
+        sb.append("\\n");
+        for (int i = 0; i < len; i++) {
+            sb.append('b');
+        }
+        sb.append("\"}]}");
```
wenshao: **[Suggestion]** The 10-line JSON payload builder is duplicated verbatim in `testVariousLengths()` (lines 25–34) and `assertMessageRoundTripUTF8()` (lines 62–71). The Javadoc documents the sync invariant ("Same payloads as {@link #testVariousLengths()}") but nothing enforces it. — Concrete cost: if the payload shape is later changed (e.g. adding another escape sequence), editing only one copy leaves JSONReaderUTF16 and JSONReaderUTF8 tested against *different inputs*, silently masking the reader-specific divergence this test exists to catch.

```suggestion
    private static String buildJsonPayload(int len) {
        StringBuilder sb = new StringBuilder();
        sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
        for (int i = 0; i < len; i++) {
            sb.append('a');
        }
        sb.append("\\n");
        for (int i = 0; i < len; i++) {
            sb.append('b');
        }
        sb.append("\"}]}");
        return sb.toString();
    }
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
```diff
@@ -0,0 +1,102 @@
...
+            sb.append("\"}]}");
+
+            JSONObject result = JSON.parseObject(sb.toString());
+            assertNotNull(result);
+            String message = result.getJSONArray("Test")
+                    .getJSONObject(1)
+                    .getString("message");
+            assertEquals(len * 2 + 1, message.length());
+        }
+    }
```
wenshao: **[Suggestion]** `testVariousLengths` and `assertMessageRoundTripUTF8` assert only the parsed string's *length*, not its *content*. A character-substitution mutation in the escape handler (e.g. `\n` → `\r`) preserves length and passes all four tests. The JSONB test in the same file already demonstrates the full-content assertion pattern. — Failure scenario: a one-byte mutation to `CHAR1_ESCAPED` changing `'n'` from `'\n'` (0x0A) to `'\r'` (0x0D) makes every `\n` JSON escape produce a carriage return; all four tests pass because `message.length()` is unchanged.

```suggestion
            StringBuilder expected = new StringBuilder(len * 2 + 1);
            for (int i = 0; i < len; i++) {
                expected.append('a');
            }
            expected.append('\n');
            for (int i = 0; i < len; i++) {
                expected.append('b');
            }
            assertEquals(expected.toString(), message);
        }
    }
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
```diff
@@ -0,0 +1,102 @@
...
+            assertMessageRoundTripUTF8(len);
+        }
+        // strBuf is pooled process-wide, so a preceding test may leave a buffer that already
+        // covers the lengths above; this one is past any capacity another test can have grown it to.
+        assertMessageRoundTripUTF8(100_000);
+    }
+
+    private void assertMessageRoundTripUTF8(int len) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("{\"Test\":[{\"message\":\"中\\n\"},{\"message\":\"");
```
wenshao: **[Suggestion]** No test exercises the initial `strBuf` allocation path with non-zero `stroff`. Every payload places multi-byte char `中` first, forcing `stroff = 0` at allocation time, so `new char[stroff + 512]` is indistinguishable from the pre-fix `new char[512]`. — Failure scenario: reverting `JSONReaderUTF16.java:3256` or `JSONReaderUTF8.java:5251` from `new char[stroff + 512]` to `new char[512]` passes all four tests. A user whose first parsed string has >512 ASCII chars before an escape hits `ArrayIndexOutOfBoundsException` at the `System.arraycopy` pre-copy. Consider adding a test where the first string value has a long ASCII run before the escape, e.g. `{"message":"aaa...(600 a's)...\n"}`.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
```diff
@@ -0,0 +1,102 @@
...
+        assertNotNull(result);
+        String message = result.getJSONArray("Test")
+                .getJSONObject(1)
+                .getString("message");
+        assertEquals(len * 2 + 1, message.length());
+    }
+
+    @Test
+    public void testVariousLengthsJSONB() {
+        for (int len = 1; len <= 600; len++) {
```
wenshao, line 80: **[Suggestion]** `testVariousLengthsJSONB` exercises `JSONReaderJSONB`, which has no `strBuf` field and does not share the escape-handling code path fixed by Issue 7685 (`grep strBuf JSONReaderJSONB.java` → zero matches). The three parallel method names create the impression of redundant coverage across all three reader implementations, but the JSONB test always passes regardless of `strBuf` regressions. — Concrete cost: a maintainer may conclude the fix is verified across all three readers when only UTF16 and UTF8 are actually covered. Consider adding a Javadoc comment clarifying this is a general JSONB round-trip sanity check.

wenshao (commented):
Reviewed. Suggestions are inline. Unresolved, please confirm: **[Critical]** JSONReaderJSONB.java readBigInteger0() BC_BINARY case (~line 4483): unguarded readInt32Value() → new byte[len] → System.arraycopy OOM-DoS pattern is present at the reviewed commit, but lives in code this test-only PR does not modify; independently fixed in main by 373cbe728 (#7705) which is not yet in this branch's merge base; whether it should gate this regression-test PR is undecidable from this diff

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
wenshao: Done — one `payload(int len)` builder now feeds both `testVariousLengths()` and `assertMessageRoundTripUTF8()`, alongside `expectedMessage(int len)` and a small `repeat(...)` helper, so the two readers cannot end up tested against different inputs. The Javadoc that used to state the invariant is gone, since the shared builder now enforces it.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
wenshao: Done — the length assertions are replaced by full-content `assertEquals(expectedMessage(len), ...)` in the UTF-16, UTF-8 and new prefix tests.

Confirmed with the mutation you described. Changing `'n', '\n'` to `'n', '\r'` in the `char1_escaped` table (`JSONReader.java:102`):

```
Tests run: 5, Failures: 3, Errors: 0
  Issue7685.testLongPrefixBeforeFirstEscape:69 expected: <中aaa...\n> but was: <中aaa...\r>
```

3 of the 5 tests fail. Before this change the same mutation passed everything, exactly as you said.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
wenshao: Confirmed and fixed — this was a real gap, and the first version of the test I wrote did not close it either.

Added `testLongPrefixBeforeFirstEscape`, sweeping `500, 511, 512, 513, 600, 4096` characters before the escape, with the escape in the first string value of the document.

One detail worth recording, because it is what makes the test work: a payload of plain ASCII does **not** reach `JSONReaderUTF16` at all. `JSON.parseObject(String)` routes a LATIN1-coded string to the ASCII/UTF-8 reader, and `JSONReaderUTF8` takes `strBuf` from a process-wide cache, which any earlier test may have already grown. So my first attempt (`{"message":"aaa...\n"}`, both as `String` and as `byte[]`, as suggested) still passed with the fix reverted. Prefixing the value with a multi-byte character forces the UTF-16 reader, whose `strBuf` is a per-instance field and therefore null on every parse.

Revert-probe with `new char[stroff + 512]` restored to `new char[512]` in both readers:

```
[ERROR] Issue7685.testLongPrefixBeforeFirstEscape:69
        java.lang.ArrayIndexOutOfBoundsException:
        arraycopy: last destination index 600 out of bounds for char[512]
  at com.alibaba.fastjson2.JSONReaderUTF16.readString(JSONReaderUTF16.java:3261)
```

The UTF-8 side of the same allocation stays uncovered — with a shared cached buffer there is no reliable way to force `strBuf == null` from a test in a warm JVM. Flagging that rather than pretending it is covered.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7685.java
wenshao: Done — Javadoc added, stating that it is a round-trip sanity check and does not cover the strBuf fix: `JSONReaderJSONB` has no strBuf and reads string values by length prefix instead of scanning for escapes, so it cannot reach the path #7685 is about.

Also rebased onto current `main` (the branch was 15 commits behind). Green locally on JDK 26: core 7963 + extension 56 + fastjson1-compatible 1355 + kotlin 36 + safemode-test 2, and core again with `-Dfastjson2.creator=reflect`.

PR #7701, opened 2026-07-27
fix: detect references for Collection elements (e.g. HashSet) (#7678)

wenshao:
## Problem

When `JSONWriter.Feature.ReferenceDetection` is enabled, `ObjectWriterImplCollection` did not detect references among collection elements. A `HashSet` (or any `Collection` that is not a `List`) containing a duplicate object would serialize the object in full both times instead of emitting a `$ref` for the second occurrence. fastjson1 handled this correctly.

Two gaps existed in `ObjectWriterImplCollection`:

1. **`write` (JSON path):** no reference detection at all — elements were written directly without `setPath`/`writeReference`/`popPath`.
2. **`writeJSONB` (JSONB path):** reference detection was explicitly disabled for any non-`SortedSet`, non-`LinkedHashSet` collection with size > 1.

## Fix

- **JSON path:** added per-element reference detection following the same pattern as `ObjectWriterImplList.write` — checks `jsonWriter.isRefDetect()`, filters via `ObjectWriterProvider.isNotReferenceDetect(itemClass)`, then calls `setPath`/`writeReference`/`popPath`.
- **JSONB path:** removed the restriction that disabled ref detection for unordered sets.

## Test

`Issue7678.java` — 5 test cases:
- HashSet ref detection in JSON (exact scenario from the issue)
- HashSet ref detection in JSONB (round-trip)
- Multi-element HashSet
- LinkedHashSet
- TreeSet

## Verification

- 5/5 new tests pass
- 431 regression tests pass (features, jsonb, issues_7000, Set-specific) — 0 failures
- `mvn validate` (checkstyle) clean

### DIFF
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7678.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7678.java
@@ -1,0 +1,132 @@
+package com.alibaba.fastjson2.issues_7000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONB;
+import com.alibaba.fastjson2.JSONWriter;
+import org.junit.jupiter.api.Test;
+
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class Issue7678 {
+    public static class Person {
+        public Set<Bean> testSet;
+    }
+
+    public static class Bean {
+        public String xxx1;
+        public String xxx2;
+        public String xxx3;
+    }
+
+    public static class Root {
+        public List<Object> data;
+        public Person person;
+    }
+
+    @Test
+    public void testHashSetRefDetectJSON() {
+        Bean bean = new Bean();
+        bean.xxx1 = "123";
+        bean.xxx2 = "123";
+        bean.xxx3 = "123";
+
+        Person person = new Person();
+        person.testSet = new HashSet<>();
+        person.testSet.add(bean);
+
+        Map<String, Object> item = new LinkedHashMap<>();
+        item.put("Person", person);
+
+        Root root = new Root();
+        root.data = new ArrayList<>();
+        root.data.add(item);
+        root.person = person;
+
+        String json = JSON.toJSONString(root, JSONWriter.Feature.ReferenceDetection);
+        assertTrue(json.contains("$ref"), "Expected $ref in output: " + json);
+
+        Object parsed = JSON.parse(json);
+        assertNotNull(parsed);
+    }
+
+    @Test
+    public void testHashSetRefDetectJSONB() {
+        Bean bean = new Bean();
+        bean.xxx1 = "123";
+        bean.xxx2 = "123";
+        bean.xxx3 = "123";
+
+        Person person = new Person();
+        person.testSet = new HashSet<>();
+        person.testSet.add(bean);
+
+        Map<String, Object> item = new LinkedHashMap<>();
+        item.put("Person", person);
+
+        Root root = new Root();
+        root.data = new ArrayList<>();
+        root.data.add(item);
+        root.person = person;
+
+        byte[] bytes = JSONB.toBytes(root, JSONWriter.Feature.ReferenceDetection);
+        Root parsed = JSONB.parseObject(bytes, Root.class);
+        assertNotNull(parsed);
+        assertNotNull(parsed.person);
+        assertEquals(1, parsed.person.testSet.size());
+        Bean parsedBean = parsed.person.testSet.iterator().next();
+        assertEquals("123", parsedBean.xxx1);
+    }
+
+    @Test
+    public void testHashSetMultiElementRefDetect() {
+        Bean bean1 = new Bean();
+        bean1.xxx1 = "a";
+
+        Bean bean2 = new Bean();
+        bean2.xxx1 = "b";
+
+        Set<Bean> set = new HashSet<>();
+        set.add(bean1);
+        set.add(bean2);
+
+        Map<String, Object> map = new LinkedHashMap<>();
+        map.put("set1", set);
+        map.put("set2", set);
+
+        String json = JSON.toJSONString(map, JSONWriter.Feature.ReferenceDetection);
+        assertTrue(json.contains("$ref"), "Expected $ref in output: " + json);
+    }
+
+    @Test
+    public void testLinkedHashSetRefDetect() {
+        Bean bean = new Bean();
+        bean.xxx1 = "123";
+
+        Person person = new Person();
+        person.testSet = new LinkedHashSet<>();
+        person.testSet.add(bean);
+
+        Map<String, Object> map = new LinkedHashMap<>();
+        map.put("p1", person);
+        map.put("p2", person);
+
+        String json = JSON.toJSONString(map, JSONWriter.Feature.ReferenceDetection);
+        assertTrue(json.contains("$ref"), "Expected $ref in output: " + json);
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -580,6 +580,9 @@ public Object readAny() {
             }
             case BC_BIGINT: {
                 int len = readInt32Value();
+                if (len < 0 || len > end - offset) {
+                    throw new JSONException("BC_BIGINT length out of range: " + len);
+                }
```
wenshao, line 585: **[Suggestion]** The same 3-line bounds check is duplicated across four `BC_BIGINT` locations (lines 583, 3259, 4189, 4421). This duplication directly caused the `BC_BINARY` sibling in `readBigInteger0()` to be missed — the same `readInt32Value()` → `new byte[len]` → `System.arraycopy` pattern exists there without a guard.

Consider extracting a helper:

```suggestion
                checkByteArrayLen(len);
```

```java
private void checkByteArrayLen(int len) {
        throw new JSONException("byte array length out of range: " + len);
    }
}
```

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
```diff
@@ -0,0 +1,43 @@
...
+public class Issue7669 {
+    @Test
+    public void testBigIntDeclaredLengthOOM() {
+        // 0xBB = BC_BIGINT, 0x48 = BC_INT32, 0x7FFFFFFF = Integer.MAX_VALUE
+        byte[] payload = {
+                (byte) 0xBB,
+                (byte) 0x48,
+                (byte) 0x7F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
+        };
+        assertThrows(JSONException.class, () -> JSONB.parse(payload));
```
wenshao, line 21: **[Suggestion]** The bounds guard is added to four methods — `readAny()`, `readString()`, `readDoubleValue()`, and `readBigInteger()` — but these tests only exercise `readAny()` via `JSONB.parse(payload)`. The other three methods are untested.

Add tests through typed parse entry points to catch regressions in each guarded method:

```java
assertThrows(JSONException.class, () -> JSONB.parseObject(payload, java.math.BigInteger.class));
assertThrows(JSONException.class, () -> JSONB.parseObject(payload, String.class));
```

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7678.java
```diff
@@ -0,0 +1,132 @@
...
+        Map<String, Object> item = new LinkedHashMap<>();
+        item.put("Person", person);
+
+        Root root = new Root();
+        root.data = new ArrayList<>();
+        root.data.add(item);
+        root.person = person;
+
+        String json = JSON.toJSONString(root, JSONWriter.Feature.ReferenceDetection);
+        assertTrue(json.contains("$ref"), "Expected $ref in output: " + json);
```
wenshao, line 49: **[Suggestion]** The `$ref` asserted here is generated by parent-level writers (Map/Root), not by the new per-item `setPath`/`popPath` logic in `ObjectWriterImplCollection.write()`. These tests would likely pass even without the fix.

Add a test where the same object appears as multiple items within a single collection to directly exercise the per-item ref detection:

```java
@Test
public void testDuplicateItemWithinCollection() {
    Bean bean = new Bean();
    bean.xxx1 = "shared";
    List<Bean> list = new ArrayList<>();
    list.add(bean);
    list.add(bean); // same instance → should produce $ref
    String json = JSON.toJSONString(list, JSONWriter.Feature.ReferenceDetection);
    assertTrue(json.contains("$ref"), "Expected $ref for duplicate item within list: " + json);
}
```

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running. Reviewed. Suggestions are inline. **[Critical]** BC_BINARY case in readBigInteger0() (JSONReaderJSONB.java:4484-4489) — same OOM vulnerability pattern as BC_BIGINT but unguarded. A crafted JSONB payload with BC_BINARY type byte + Integer.MAX_VALUE length triggers a ~2GB heap allocation via new byte[len] without bounds validation. This is the same vulnerability class (readInt32Value() → new byte[len] → System.arraycopy) the PR fixes for BC_BIGINT, missed in a sibling code path.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
wenshao: Not applicable on this branch: `JSONReaderJSONB.java` here is byte-identical to `main` (`git diff main HEAD -- core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java` is empty). The hunk only shows up in the PR diff because the branch predates the merge of #7696, which landed these same `BC_BIGINT` guards on `main`; the merge-base diff therefore still attributes them to this PR.

The follow-ups belong to the PRs that actually own the JSONB work: the `checkBigintLen` helper extraction and the typed-path `Issue7669` tests are on #7698 (already applied there), and the missing `BC_BINARY` sibling is on #7705. Nothing to change here — I'll rebase this branch onto `main` so the diff stops showing unrelated files.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7669.java
wenshao: Not applicable on this branch: `JSONReaderJSONB.java` here is byte-identical to `main` (`git diff main HEAD -- core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java` is empty). The hunk only shows up in the PR diff because the branch predates the merge of #7696, which landed these same `BC_BIGINT` guards on `main`; the merge-base diff therefore still attributes them to this PR.

The follow-ups belong to the PRs that actually own the JSONB work: the `checkBigintLen` helper extraction and the typed-path `Issue7669` tests are on #7698 (already applied there), and the missing `BC_BINARY` sibling is on #7705. Nothing to change here — I'll rebase this branch onto `main` so the diff stops showing unrelated files.

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7678.java
wenshao: Done — added `testDuplicateItemWithinCollection` and `testDuplicateItemWithinCollectionJSONB`.

One adjustment to the suggested snippet: a `List` routes to `ObjectWriterImplList`, not `ObjectWriterImplCollection`, and a `Set` cannot hold the same instance twice, so neither shape reaches the new per-item `setPath`/`popPath`. I used `ArrayDeque` — it is a `Collection` (so `ObjectWriterBaseModule` returns `ObjectWriterImplCollection.INSTANCE`) and it keeps duplicates.

Revert-probe with `ObjectWriterImplCollection` restored to `main`:

```
[ERROR] Issue7678.testDuplicateItemWithinCollection Expected $ref for duplicate item within collection:
        [{"xxx1":"shared"},{"xxx1":"shared"}] ==> expected: <true> but was: <false>
[ERROR] Issue7678.testDuplicateItemWithinCollectionJSONB ...
[ERROR] Tests run: 7, Failures: 2
```

The 5 pre-existing tests pass in that run too, which confirms your point that they were not gating the fix. All 7 pass with the change applied. The JSONB variant also covers the removed `collection.size() > 1 && !SortedSet && !LinkedHashSet` blanket disable in `writeJSONB`, and asserts the `$ref` resolves back to the same instance on parse.

PR #7703, opened 2026-07-27
fix: strengthen autoType type name validation and whitelist verification

wenshao:
## Summary

Hardens the autoType deserialization path with three layers of defense:

1. **Type name format validation** — Type names containing `:` or `!` are reported as unresolvable in `checkAutoType()`, `ContextAutoTypeBeforeHandler.apply()`, and `TypeUtils.loadClass()`. These characters never appear in valid Java binary class names, and a nested jar URL (`jar:http://host/x.jar!/`) is resolvable by some class loaders, so a type name carrying them must never reach one. Shared as `TypeUtils.hasIllegalTypeNameChars` — validating in only some entry points is the asymmetry this PR exists to fix.

   Unresolvable, not an error: fastjson2 falls back to a map for an `@type` it cannot resolve, and `@type` is also a JSON-LD keyword whose value is an IRI, so throwing here would break every JSON-LD document. `ErrorOnNotSupportAutoType` still turns it into an exception for callers that want one.

2. **Whitelist text verification** — After a rolling-hash match against `acceptHashCodes`, verify the actual prefix text against a parallel `acceptNameSet` rather than trusting the hash value alone.

3. **Consistent blacklist enforcement** — Apply the `ClassLoader` and `DataSource/RowSet` blacklist in *both* whitelist branches of `checkAutoType` and in `ContextAutoTypeBeforeHandler`. Previously only the general `SupportAutoType` fallback had it, so a whitelisted prefix gave *weaker* checking than no whitelist entry at all: `{"@type":"com.example.EvilClassLoader"}` was rejected without an accept entry and accepted with `addAutoTypeAccept("com.example.")`.

   Matching an accept entry that names the type **in full** stays an explicit opt-in and skips the blacklist, so `addAutoTypeAccept("com.example.MyDataSource")` keeps working. A prefix match on a blacklisted type continues scanning rather than failing outright, so a shorter prefix entry cannot preempt a full-name entry appearing later in the type name.

## Two things text verification surfaced

- The hardcoded accept hash `-6293031534589903644L` is `Fnv.hashCode64("com.alibaba.fastjson.util.AntiCollisionHashMap")`. It has no accept name behind it, so text verification would have silently killed that always-accepted entry. The name is now registered alongside the hash, and a test pins the two together.

- `ObjectReaderProvider` hashed accept names with a literal `$` while the rolling hash in `checkAutoType` normalizes `$` to `.`, so an accept entry written with the binary name of a nested class (`a.b.Outer$Inner`) never matched. Accept names are now normalized through a shared `TypeUtils.normalizeAcceptName`, matching what `ContextAutoTypeBeforeHandler` already did.

## Breaking change

Registering a `ClassLoader` or `DataSource`/`RowSet` type **by prefix** no longer allows it — the type must be named in full. This breaks the fastjson 1.x style `ParserConfig.getGlobalInstance().addAccept("com.alibaba.druid.pool.")`, and the equivalent with `addAutoTypeAccept`, `-Dfastjson2.autoTypeAccept` and `JSONReader.autoTypeFilter`. Documented in `docs/autotype_{cn,en}.md` §2.1; worth a line in the 2.0.62 release notes.

When a rejection is caused by exactly this, the message says so (`an accept prefix does not cover it, add the type name in full to accept`) rather than the generic text — the hint is raised only when an accept prefix actually matched, so an error triggered by untrusted input naming a gadget does not carry instructions for whitelisting it.

## Also

- `addAutoTypeAccept` publishes `acceptNameSet` **before** `acceptHashCodes`. Readers consult the hash array first, so volatile happens-before transitivity guarantees a reader that sees a new hash also sees the name it verifies against, instead of transiently rejecting an accepted type.
- `ContextAutoTypeBeforeHandler` checks the blacklist after its class cache is consulted, so a cached class is checked too rather than the check resting on a deny class never being cached.

## Changes

| File | Change |
|------|--------|
| `ObjectReaderProvider.java` | `acceptNameSet`, format validation, text verification, blacklist in both whitelist branches, `$` normalization of accept names, publication order |
| `ContextAutoTypeBeforeHandler.java` | `acceptNameSet`, format validation, text verification, blacklist on prefix matches |
| `TypeUtils.java` | `hasIllegalTypeNameChars`, `normalizeAcceptName`, format validation in `loadClass()` |
| `JDKUtils.java` | `isAutoTypeDenyClass()` — single definition of the blacklist |
| `AutoTypeValidationTest.java` | Regression tests covering all three layers |
| `docs/autotype_{cn,en}.md` | Document the accept-prefix behaviour change |

## Testing

- `AutoTypeValidationTest` — 27 cases, mutation-checked. Removing any of these fails at least one test:
  - either character in `hasIllegalTypeNameChars`, and its call in `TypeUtils.loadClass` and in `checkAutoType`
  - text verification in both `checkAutoType` branches and in `ContextAutoTypeBeforeHandler`
  - the blacklist in both `checkAutoType` branches, the fallback, and `ContextAutoTypeBeforeHandler`, and **either arm** of `isAutoTypeDenyClass`
  - the full-name-accept exemption, the `AntiCollisionHashMap` accept name, `normalizeAcceptName`, and the `denyPrefixOnly` message
  - The one guard not independently pinned is the `hasIllegalTypeNameChars` call in `ContextAutoTypeBeforeHandler.apply` — it is redundant with the identical check in `TypeUtils.loadClass`, kept as defense in depth so the public entry point does not depend on `loadClass` internals.
- `core` — 7933 tests pass, 413 of them `@Tag("autotype")`
- `fastjson1-compatible` (1346) and `safemode-test` (2) pass
- checkstyle passes

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
+++ b/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
@@ -276,12 +276,6 @@ public class ContextAutoTypeBeforeHandler
                 if (clazz == null) {
                     clazz = loadClass(typeName);
                     if (clazz != null) {
-                        // matching an accept prefix is not an opt-in for gadget base types, only an
-                        // accept entry naming the type in full is; keep scanning for such an entry
-                        if (i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) {
-                            continue;
-                        }
-
                         Class origin = putCacheIfAbsent(typeNameHash, clazz);
                         if (origin != null) {
                             clazz = origin;
@@ -290,6 +284,13 @@ public class ContextAutoTypeBeforeHandler
                 }
 
                 if (clazz != null) {
+                    // matching an accept prefix is not an opt-in for gadget base types, only an
+                    // accept entry naming the type in full is; keep scanning for such an entry.
+                    // checked after the cache read so that a cached class is checked too
+                    if (i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) {
+                        continue;
+                    }
+
                     return clazz;
                 }
             }
--- a/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
+++ b/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
@@ -3,6 +3,7 @@ package com.alibaba.fastjson2.filter;
 import com.alibaba.fastjson2.JSON;
 import com.alibaba.fastjson2.JSONReader;
 import com.alibaba.fastjson2.util.Fnv;
+import com.alibaba.fastjson2.util.JDKUtils;
 import com.alibaba.fastjson2.util.TypeUtils;
 
 import java.math.BigDecimal;
@@ -276,6 +277,12 @@ public class ContextAutoTypeBeforeHandler
                 if (clazz == null) {
                     clazz = loadClass(typeName);
                     if (clazz != null) {
+                        // matching an accept prefix is not an opt-in for gadget base types, only an
+                        // accept entry naming the type in full is; keep scanning for such an entry
+                        if (i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) {
+                            continue;
+                        }
+
                         Class origin = putCacheIfAbsent(typeNameHash, clazz);
                         if (origin != null) {
                             clazz = origin;
--- a/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
+++ b/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
@@ -223,7 +223,7 @@ public class ContextAutoTypeBeforeHandler
             }
 
             array[index++] = hashCode;
-            normalizedNames.add(name.replace('$', '.'));
+            normalizedNames.add(normalizeAcceptName(name));
         }
 
         if (index != array.length) {
@@ -253,7 +253,7 @@ public class ContextAutoTypeBeforeHandler
             typeName = "Object";
         }
 
-        if (typeName.indexOf(':') >= 0 || typeName.indexOf('!') >= 0) {
+        if (hasIllegalTypeNameChars(typeName)) {
             return null;
         }
 
@@ -267,8 +267,7 @@ public class ContextAutoTypeBeforeHandler
             hash *= MAGIC_PRIME;
 
             if (Arrays.binarySearch(acceptHashCodes, hash) >= 0) {
-                String prefix = typeName.substring(0, i + 1).replace('$', '.');
-                if (!acceptNameSet.contains(prefix)) {
+                if (!acceptNameSet.contains(normalizeAcceptName(typeName.substring(0, i + 1)))) {
                     continue;
                 }
                 long typeNameHash = Fnv.hashCode64(typeName);
--- a/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
+++ b/core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
@@ -24,6 +24,7 @@ import static com.alibaba.fastjson2.util.TypeUtils.*;
 public class ContextAutoTypeBeforeHandler
         implements JSONReader.AutoTypeBeforeHandler {
     final long[] acceptHashCodes;
+    final Set<String> acceptNameSet;
     final ConcurrentMap<Integer, ConcurrentHashMap<Long, Class>> tclHashCaches = new ConcurrentHashMap<>();
     final Map<Long, Class> classCache = new ConcurrentHashMap<>(16, 0.75f, 1);
 
@@ -206,6 +207,7 @@ public class ContextAutoTypeBeforeHandler
         }
 
         long[] array = new long[nameSet.size()];
+        Set<String> normalizedNames = new HashSet<>(nameSet.size());
 
         int index = 0;
         for (String name : nameSet) {
@@ -220,6 +222,7 @@ public class ContextAutoTypeBeforeHandler
             }
 
             array[index++] = hashCode;
+            normalizedNames.add(name.replace('$', '.'));
         }
 
         if (index != array.length) {
@@ -227,6 +230,7 @@ public class ContextAutoTypeBeforeHandler
         }
         Arrays.sort(array);
         this.acceptHashCodes = array;
+        this.acceptNameSet = Collections.unmodifiableSet(normalizedNames);
     }
 
     public Class<?> apply(long typeNameHash, Class<?> expectClass, long features) {
@@ -248,6 +252,10 @@ public class ContextAutoTypeBeforeHandler
             typeName = "Object";
         }
 
+        if (typeName.indexOf(':') >= 0 || typeName.indexOf('!') >= 0) {
+            return null;
+        }
+
         long hash = MAGIC_HASH_CODE;
         for (int i = 0, typeNameLength = typeName.length(); i < typeNameLength; ++i) {
             char ch = typeName.charAt(i);
@@ -258,6 +266,10 @@ public class ContextAutoTypeBeforeHandler
             hash *= MAGIC_PRIME;
 
             if (Arrays.binarySearch(acceptHashCodes, hash) >= 0) {
+                String prefix = typeName.substring(0, i + 1).replace('$', '.');
+                if (!acceptNameSet.contains(prefix)) {
+                    continue;
+                }
                 long typeNameHash = Fnv.hashCode64(typeName);
                 Class clazz = apply(typeNameHash, expectClass, features);
 
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -196,7 +196,20 @@ public class ObjectReaderProvider
     boolean disableAutoType = JSONFactory.isDisableAutoType();
     boolean disableSmartMatch = JSONFactory.isDisableSmartMatch();
 
+    /**
+     * Always accepted, it is the map implementation fastjson 1.x substitutes for {@code HashMap}
+     * when hash collision protection is enabled.
+     */
+    static final String ANTI_COLLISION_HASH_MAP = "com.alibaba.fastjson.util.AntiCollisionHashMap";
+    static final long ANTI_COLLISION_HASH_MAP_HASH = -6293031534589903644L; // Fnv.hashCode64(ANTI_COLLISION_HASH_MAP)
+
     private volatile long[] acceptHashCodes;
+
+    /**
+     * The accept names behind {@link #acceptHashCodes}, normalized the same way the rolling hash in
+     * {@link #checkAutoType} normalizes a type name. A hash match is only honored when the matched
+     * prefix text is in this set, so a hash collision alone cannot whitelist a type name.
+     */
     private volatile Set<String> acceptNameSet = Collections.emptySet();
 
     private AutoTypeBeforeHandler autoTypeBeforeHandler = DEFAULT_AUTO_TYPE_BEFORE_HANDLER;
@@ -211,13 +224,14 @@ public class ObjectReaderProvider
         } else {
             hashCodes = new long[AUTO_TYPE_ACCEPT_LIST.length + 1];
             for (int i = 0; i < AUTO_TYPE_ACCEPT_LIST.length; i++) {
-                String name = AUTO_TYPE_ACCEPT_LIST[i].replace('$', '.');
-                hashCodes[i] = Fnv.hashCode64(AUTO_TYPE_ACCEPT_LIST[i]);
+                String name = normalizeAcceptName(AUTO_TYPE_ACCEPT_LIST[i]);
+                hashCodes[i] = Fnv.hashCode64(name);
                 names.add(name);
             }
         }
 
-        hashCodes[hashCodes.length - 1] = -6293031534589903644L;
+        hashCodes[hashCodes.length - 1] = ANTI_COLLISION_HASH_MAP_HASH;
+        names.add(ANTI_COLLISION_HASH_MAP);
 
         Arrays.sort(hashCodes);
         acceptHashCodes = hashCodes;
@@ -260,7 +274,8 @@ public class ObjectReaderProvider
      */
     public synchronized void addAutoTypeAccept(String name) {
         if (name != null && name.length() != 0) {
-            long hash = Fnv.hashCode64(name);
+            String acceptName = normalizeAcceptName(name);
+            long hash = Fnv.hashCode64(acceptName);
             long[] current = this.acceptHashCodes;
             if (Arrays.binarySearch(current, hash) < 0) {
                 long[] hashCodes = new long[current.length + 1];
@@ -269,10 +284,25 @@ public class ObjectReaderProvider
                 Arrays.sort(hashCodes);
                 this.acceptHashCodes = hashCodes;
             }
-            Set<String> names = new HashSet<>(this.acceptNameSet);
-            names.add(name.replace('$', '.'));
-            this.acceptNameSet = Collections.unmodifiableSet(names);
+            if (!this.acceptNameSet.contains(acceptName)) {
+                Set<String> names = new HashSet<>(this.acceptNameSet);
+                names.add(acceptName);
+                this.acceptNameSet = Collections.unmodifiableSet(names);
+            }
         }
+    }
+
+    /**
+     * Normalizes an accept name the same way {@link #checkAutoType} normalizes the type name it
+     * hashes, so that an accept entry written with the binary name of a nested class
+     * ({@code a.b.Outer$Inner}) matches as well as one written with its canonical name
+     * ({@code a.b.Outer.Inner}).
+     *
+     * @param name the accept name to normalize
+     * @return the normalized accept name
+     */
+    private static String normalizeAcceptName(String name) {
+        return name.indexOf('$') >= 0 ? name.replace('$', '.') : name;
     }
 
     @Deprecated
@@ -844,12 +874,17 @@ public class ObjectReaderProvider
                 hash ^= ch;
                 hash *= MAGIC_PRIME;
                 if (Arrays.binarySearch(acceptHashCodes, hash) >= 0) {
-                    String prefix = typeName.substring(0, i + 1).replace('$', '.');
-                    if (!acceptNameSet.contains(prefix)) {
+                    if (!acceptNameSet.contains(normalizeAcceptName(typeName.substring(0, i + 1)))) {
                         continue;
                     }
                     clazz = loadClass(typeName);
                     if (clazz != null) {
+                        // matching an accept prefix is not an opt-in for gadget base types, only an
+                        // accept entry naming the type in full is; keep scanning for such an entry
+                        if (i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) {
+                            continue;
+                        }
+
                         if (expectClass != null && !expectClass.isAssignableFrom(clazz)) {
                             throw new JSONException("type not match. " + typeName + " -> " + expectClass.getName());
                         }
@@ -873,16 +908,14 @@ public class ObjectReaderProvider
 
                 // white list
                 if (Arrays.binarySearch(acceptHashCodes, hash) >= 0) {
-                    String prefix = typeName.substring(0, i + 1).replace('$', '.');
-                    if (!acceptNameSet.contains(prefix)) {
+                    if (!acceptNameSet.contains(normalizeAcceptName(typeName.substring(0, i + 1)))) {
                         continue;
                     }
                     clazz = loadClass(typeName);
 
-                    if (clazz != null) {
-                        if (ClassLoader.class.isAssignableFrom(clazz) || JDKUtils.isSQLDataSourceOrRowSet(clazz)) {
-                            throw new JSONException("autoType is not support. " + typeName);
-                        }
+                    // see the SupportAutoType branch above
+                    if (clazz != null && i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) {
+                        continue;
                     }
 
                     if (clazz != null && expectClass != null && !expectClass.isAssignableFrom(clazz)) {
@@ -919,7 +952,7 @@ public class ObjectReaderProvider
         clazz = loadClass(typeName);
 
         if (clazz != null) {
-            if (ClassLoader.class.isAssignableFrom(clazz) || JDKUtils.isSQLDataSourceOrRowSet(clazz)) {
+            if (JDKUtils.isAutoTypeDenyClass(clazz)) {
                 throw new JSONException("autoType is not support. " + typeName);
             }
 
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -28,7 +28,9 @@ import java.util.function.Supplier;
 import static com.alibaba.fastjson2.JSONFactory.*;
 import static com.alibaba.fastjson2.util.Fnv.MAGIC_HASH_CODE;
 import static com.alibaba.fastjson2.util.Fnv.MAGIC_PRIME;
+import static com.alibaba.fastjson2.util.TypeUtils.hasIllegalTypeNameChars;
 import static com.alibaba.fastjson2.util.TypeUtils.loadClass;
+import static com.alibaba.fastjson2.util.TypeUtils.normalizeAcceptName;
 
 /**
  * ObjectReaderProvider is responsible for providing and managing ObjectReader instances
@@ -234,8 +236,9 @@ public class ObjectReaderProvider
         names.add(ANTI_COLLISION_HASH_MAP);
 
         Arrays.sort(hashCodes);
-        acceptHashCodes = hashCodes;
+        // see addAutoTypeAccept, the name set is published before the hash array
         acceptNameSet = Collections.unmodifiableSet(names);
+        acceptHashCodes = hashCodes;
 
         hashCache.put(ObjectArrayReader.TYPE_HASH_CODE, ObjectArrayReader.INSTANCE);
         final long STRING_CLASS_NAME_HASH = -4834614249632438472L; // Fnv.hashCode64(String.class.getName());
@@ -275,6 +278,15 @@ public class ObjectReaderProvider
     public synchronized void addAutoTypeAccept(String name) {
         if (name != null && name.length() != 0) {
             String acceptName = normalizeAcceptName(name);
+
+            // publish the name before the hash, so that a reader seeing the new hash array is
+            // guaranteed to see the name it verifies against rather than transiently rejecting
+            if (!this.acceptNameSet.contains(acceptName)) {
+                Set<String> names = new HashSet<>(this.acceptNameSet);
+                names.add(acceptName);
+                this.acceptNameSet = Collections.unmodifiableSet(names);
+            }
+
             long hash = Fnv.hashCode64(acceptName);
             long[] current = this.acceptHashCodes;
             if (Arrays.binarySearch(current, hash) < 0) {
@@ -284,25 +296,7 @@ public class ObjectReaderProvider
                 Arrays.sort(hashCodes);
                 this.acceptHashCodes = hashCodes;
             }
-            if (!this.acceptNameSet.contains(acceptName)) {
-                Set<String> names = new HashSet<>(this.acceptNameSet);
-                names.add(acceptName);
-                this.acceptNameSet = Collections.unmodifiableSet(names);
-            }
         }
-    }
-
-    /**
-     * Normalizes an accept name the same way {@link #checkAutoType} normalizes the type name it
-     * hashes, so that an accept entry written with the binary name of a nested class
-     * ({@code a.b.Outer$Inner}) matches as well as one written with its canonical name
-     * ({@code a.b.Outer.Inner}).
-     *
-     * @param name the accept name to normalize
-     * @return the normalized accept name
-     */
-    private static String normalizeAcceptName(String name) {
-        return name.indexOf('$') >= 0 ? name.replace('$', '.') : name;
     }
 
     @Deprecated
@@ -847,8 +841,8 @@ public class ObjectReaderProvider
             throw new JSONException("autoType is not support. " + typeName);
         }
 
-        if (typeName.indexOf(':') >= 0 || typeName.indexOf('!') >= 0) {
-            throw new JSONException("autoType is not support. " + typeName);
+        if (hasIllegalTypeNameChars(typeName)) {
+            throw new JSONException("autoType is not support, illegal type name. " + typeName);
         }
 
         if (typeName.charAt(0) == '[') {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -197,6 +197,7 @@ public class ObjectReaderProvider
     boolean disableSmartMatch = JSONFactory.isDisableSmartMatch();
 
     private volatile long[] acceptHashCodes;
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -832,6 +844,10 @@ public Class<?> checkAutoType(String typeName, Class<?> expectClass, long featur
                 hash ^= ch;
                 hash *= MAGIC_PRIME;
                 if (Arrays.binarySearch(acceptHashCodes, hash) >= 0) {
+                    String prefix = typeName.substring(0, i + 1).replace('$', '.');
+                    if (!acceptNameSet.contains(prefix)) {
+                        continue;
+                    }
                     clazz = loadClass(typeName);
                     if (clazz != null) {
                         if (expectClass != null && !expectClass.isAssignableFrom(clazz)) {
```
wenshao, line 890: **[Critical]** The `ClassLoader`/`DataSource` guard added by this PR is applied only in the `!autoTypeSupport` whitelist branch (line ~883), not in this `autoTypeSupport` whitelist branch. With `SupportAutoType` enabled and a whitelisted prefix (e.g. `"com.example."`), an attacker can load a `ClassLoader` subclass or SQL `DataSource` through this path without the guard — Failure scenario: `{"@type":"com.example.EvilClassLoader"}` matches the whitelist prefix, `loadClass` succeeds, and the class is returned at line 858 without the `ClassLoader.class.isAssignableFrom` check. The same request without `SupportAutoType` is correctly rejected.

```suggestion
                    clazz = loadClass(typeName);
                    if (clazz != null) {
                        if (ClassLoader.class.isAssignableFrom(clazz) || JDKUtils.isSQLDataSourceOrRowSet(clazz)) {
                            throw new JSONException("autoType is not support. " + typeName);
                        }
                        if (expectClass != null && !expectClass.isAssignableFrom(clazz)) {
```

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Test
+    public void testWhitelistTextVerification() {
+        ObjectReaderProvider provider = new ObjectReaderProvider();
+        provider.addAutoTypeAccept("com.example.");
+
+        long features = JSONReader.Feature.SupportAutoType.mask;
+        Class<?> clazz = provider.checkAutoType("com.other.NotAccepted", null, features);
+        assertNull(clazz);
```
wenshao: **[Suggestion]** This test never reaches the `acceptNameSet.contains(prefix)` text-verification branch it claims to test. No prefix of `"com.other.NotAccepted"` produces an FNV hash matching `"com.example."`, so `binarySearch` never returns ≥ 0 and the new guard is never executed — Concrete cost: deleting the entire text-verification guard from both `ObjectReaderProvider` and `ContextAutoTypeBeforeHandler` leaves this test green. The hash-collision defense the PR exists to close would regress undetected. Consider constructing a test that forces a hash-match/text-mismatch path.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running. Reviewed. Suggestions are inline.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testContextHandlerRejectsNonWhitelisted() {
+        ContextAutoTypeBeforeHandler handler = new ContextAutoTypeBeforeHandler(String.class);
+        assertNull(handler.apply("com.example.NotWhitelisted", null, 0));
+    }
+
+    @Test
+    public void testWhitelistWithAddAccept() {
+        ObjectReaderProvider provider = new ObjectReaderProvider();
+        provider.addAutoTypeAccept("com.alibaba.fastjson2.autoType.AutoTypeValidationTest$");
```
wenshao: **[Suggestion]** This test passes via the `SupportAutoType` general `loadClass` fallback (line ~919 of `ObjectReaderProvider`), not through the whitelist rolling-hash path — Concrete cost: `addAutoTypeAccept` stores `Fnv.hashCode64(name)` with literal `$`, but the rolling hash normalizes `$`→`.`, so the hash never matches. Deleting the `addAutoTypeAccept` line would not change the test result. Consider using a prefix without `$` (e.g. `"com.alibaba.fastjson2.autoType."`) so the whitelist path is genuinely exercised.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
```diff
@@ -0,0 +1,118 @@
...
+        ObjectReaderProvider provider = new ObjectReaderProvider();
+        long features = JSONReader.Feature.SupportAutoType.mask;
+        assertThrows(JSONException.class, () ->
+                provider.checkAutoType("com.example.Bean:invalid", null, features));
+    }
+
+    @Test
+    public void testLoadClassRejectsColon() {
+        assertNull(TypeUtils.loadClass("com.example.Bean:invalid"));
+    }
```
wenshao: **[Suggestion]** `testLoadClassRejectsColon` and `testLoadClassRejectsExclamation` are vacuous — `"com.example.Bean:invalid"` is not a real class, so `loadClass` returns null regardless of the new guard — Concrete cost: removing the `:`/`!` check from `TypeUtils.loadClass` leaves both tests green. Consider testing with a real class name augmented with `:` (e.g. `"java.lang.String:invalid"`) so the assertion discriminates the guard from the pre-existing class-not-found path.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
wenshao: Fixed in a0b90b0. The blacklist now applies in both whitelist branches and in `ContextAutoTypeBeforeHandler`, extracted as `JDKUtils.isAutoTypeDenyClass` so there is one definition.

Two deliberate refinements over the suggested patch:

- The check is skipped when the accept entry names the type **in full**, so `addAutoTypeAccept("com.example.MyDataSource")` stays an explicit opt-in. Only prefix matches are guarded.
- A guarded prefix match `continue`s instead of throwing, because accept entries are scanned shortest-prefix-first — throwing would let `com.example.` preempt a full-name entry for the same type later in the scan.

Covered by `testAcceptPrefixDoesNotAllowClassLoader`, `testExactAcceptNameAllowsClassLoader`, `testContextHandlerAcceptPrefixDoesNotAllowClassLoader` and an end-to-end `testParseRejectsClassLoaderUnderAcceptPrefix`.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
wenshao: Correct, the guard was never executed. Replaced in a0b90b0 with `testWhitelistHashMatchWithoutTextRejected` and `testWhitelistFullNameHashMatchWithoutTextRejected`, which reflectively inject a hash into `acceptHashCodes` without registering the text — a real FNV-1a-64 collision is not constructible, so this is what a collision would look like at the point the guard runs. Same approach for `ContextAutoTypeBeforeHandler` in `testContextHandlerHashMatchWithoutTextRejected`.

Removing the text check from either `checkAutoType` branch or from the handler now fails a test.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
wenshao: Right on both counts, and the `$` mismatch was a real bug rather than just a test problem: `ObjectReaderProvider` hashed accept names with a literal `$` while the rolling hash normalizes `$` to `.`, so an accept entry written with the binary name of a nested class never matched. Accept names are now normalized the same way, matching what `ContextAutoTypeBeforeHandler` already did.

The test is split in a0b90b0: `testWhitelistPrefixAccept` uses a `$`-free prefix with `features = 0` so only the whitelist path can resolve the type, and `testWhitelistAcceptNestedClassBinaryName` covers both spellings of the accept name.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
wenshao: Agreed, both were vacuous. A real class name plus `:` does not discriminate either — it fails to load with or without the guard. Replaced in a0b90b0 with `testLoadClassDoesNotReachClassLoader`, which installs a context class loader that resolves *any* name and asserts that `jar:http://127.0.0.1/evil.jar!/Evil` never reaches it, with an ordinary name as the control. That is the property the guard actually provides, and removing it fails the test.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -809,6 +847,10 @@ public Class<?> checkAutoType(String typeName, Class<?> expectClass, long featur
             throw new JSONException("autoType is not support. " + typeName);
         }
 
+        if (typeName.indexOf(':') >= 0 || typeName.indexOf('!') >= 0) {
+            throw new JSONException("autoType is not support. " + typeName);
+        }
```
wenshao: **[Suggestion]** The type-name format predicate (`indexOf(':') >= 0 || indexOf('!') >= 0`) is copy-pasted into three files: here, `ContextAutoTypeBeforeHandler.apply` (line ~257), and `TypeUtils.loadClass` (line ~2836). No shared predicate exists. — Concrete cost: if the rejected character set is later extended (e.g. adding `'\0'` or other URL-special characters), the change must be applied in three packages (`reader`, `filter`, `util`); missing one site recreates the exact asymmetric-validation failure this PR fixes.

```suggestion
            if (TypeUtils.hasIllegalTypeNameChars(typeName)) {
                throw new JSONException("autoType is not support. " + typeName);
            }
```

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+    // =========================================================================
+    // blacklist enforcement on whitelist matches
+    // =========================================================================
+
+    @Test
+    public void testAcceptPrefixDoesNotAllowClassLoader() {
+        ObjectReaderProvider provider = new ObjectReaderProvider();
+        provider.addAutoTypeAccept(PACKAGE_PREFIX);
```
wenshao, line 263: **[Suggestion]** The `DataSource`/`RowSet` branch of `isAutoTypeDenyClass` is never exercised through the whitelist-prefix code paths — all blacklist-enforcement tests use only a `ClassLoader` subclass. — Concrete cost: if a future refactor inlines only the `ClassLoader` check (the pre-PR pattern), no test detects the regression for the `DataSource`/`RowSet` half; a user who adds `com.mysql.` to their accept list would silently allow `MysqlDataSource` through the prefix match.

Consider adding a `TestDataSource` inner class implementing `javax.sql.DataSource` with parallel assertions for both `ObjectReaderProvider` and `ContextAutoTypeBeforeHandler`.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -264,9 +284,27 @@ public synchronized void addAutoTypeAccept(String name) {
                 Arrays.sort(hashCodes);
                 this.acceptHashCodes = hashCodes;
             }
+            if (!this.acceptNameSet.contains(acceptName)) {
+                Set<String> names = new HashSet<>(this.acceptNameSet);
+                names.add(acceptName);
+                this.acceptNameSet = Collections.unmodifiableSet(names);
+            }
```
wenshao: **[Suggestion]** `addAutoTypeAccept` publishes `acceptHashCodes` (line ~286) before `acceptNameSet` (line ~291) non-atomically. A concurrent reader in `checkAutoType` can see the new hash array but the old name set, producing a transient false rejection. — Concrete cost: between the two volatile writes, a parsing thread finds the hash via `binarySearch` but `acceptNameSet.contains` fails against the stale set, so the type is spuriously rejected. Fails closed (never a false accept), but swapping the write order closes the window entirely.

Publish `acceptNameSet` **before** `acceptHashCodes`: volatile happens-before transitivity then guarantees a reader seeing the new hash also sees the new name set.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -893,7 +952,7 @@ public Class<?> checkAutoType(String typeName, Class<?> expectClass, long featur
         clazz = loadClass(typeName);
 
         if (clazz != null) {
-            if (ClassLoader.class.isAssignableFrom(clazz) || JDKUtils.isSQLDataSourceOrRowSet(clazz)) {
+            if (JDKUtils.isAutoTypeDenyClass(clazz)) {
                 throw new JSONException("autoType is not support. " + typeName);
             }
```
wenshao, line 962: **[Suggestion]** The deny-class rejection message (`"autoType is not support. " + typeName`) is identical to the format and length rejections at lines ~848 and ~852. — Concrete cost: a user upgrading with `addAutoTypeAccept("com.zaxxer.hikari.")` who deserializes `HikariDataSource` gets no signal that the fix is adding the exact class name to the accept list; the 3 AM debugger cannot distinguish a misconfigured accept list from a malformed type name.

```suggestion
        if (JDKUtils.isAutoTypeDenyClass(clazz)) {
            throw new JSONException("autoType deny class matched by prefix only, add the full type name to accept list. " + typeName);
        }
```

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
```diff
@@ -220,13 +223,15 @@ public ContextAutoTypeBeforeHandler(boolean includeBasic, String... acceptNames)
             }
 
             array[index++] = hashCode;
+            normalizedNames.add(name.replace('$', '.'));
```
wenshao: **[Suggestion]** The `$` → `.` normalization is inlined here (`name.replace('$', '.')`) and at line 270 (`typeName.substring(0, i + 1).replace('$', '.')`), duplicating the rule defined in `ObjectReaderProvider.normalizeAcceptName` — Concrete cost: if the normalization rule changes in `normalizeAcceptName` (e.g., an additional character mapping), a maintainer would need to find and replicate the change in these two inline sites in a different package (`filter` vs `reader`), with no compiler or test signal on a miss. Given this is security-critical autoType validation, a desynchronization between the two paths would silently accept different type names depending on which configuration path the application uses.

Consider extracting the normalization to a shared static method (e.g., in `TypeUtils` or `Fnv`) called from both classes:

```java
// in TypeUtils or Fnv:
public static String normalizeAcceptName(String name) {
    return name.indexOf('$') >= 0 ? name.replace('$', '.') : name;
}
```

wenshao (commented):
Reviewed. Suggestions are inline. 1 Suggestion-level finding(s) could not be anchored to a changed line and were dropped; nothing further to act on here.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
wenshao: Done in 0228a1d. Extracted as `TypeUtils.hasIllegalTypeNameChars`, called from all three sites. Agreed on the reasoning — validating in only some entry points is precisely the asymmetry this PR exists to fix, so the predicate should not be able to drift between them.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
wenshao: Confirmed and fixed in 0228a1d. I mutation-checked it before and after: replacing `isAutoTypeDenyClass` with `ClassLoader.class.isAssignableFrom(type)` alone left every test green, exactly as you predicted, and now fails.

Added `TestDataSource` plus `testAcceptPrefixDoesNotAllowDataSource` and `testContextHandlerAcceptPrefixDoesNotAllowDataSource`. The helper is `abstract` — `checkAutoType` resolves the class without instantiating it, so none of the `DataSource` method bodies are needed.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
wenshao: Correct, fixed in 0228a1d. Readers consult `acceptHashCodes` first and `acceptNameSet` second, so publishing the name set first makes volatile happens-before transitivity close the window: a reader that sees the new hash array is guaranteed to see the name it verifies against. Applied to the instance initializer as well for consistency.

core/src/main/java/com/alibaba/fastjson2/filter/ContextAutoTypeBeforeHandler.java
wenshao: Done in 0228a1d. Extracted as `TypeUtils.normalizeAcceptName` and called from all three sites (`ObjectReaderProvider` accept registration and both `checkAutoType` branches, `ContextAutoTypeBeforeHandler` constructor and `apply`). Same reasoning as the format predicate: the two paths must not be able to normalize differently.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
wenshao: Taking the diagnostic point but not this wording.

This message fires for **every** deny-class rejection, not only for the accept-prefix case. The common trigger is untrusted JSON naming a gadget with no accept list involved at all, and there the suggested text tells the operator to add a `ClassLoader` or `DataSource` to their accept list — that is the one thing they must not do. An error string that fires on attack traffic should not carry instructions for disabling the control that caught it.

Addressed the half that this PR is responsible for instead: the format rejection introduced here now has its own message (`autoType is not support, illegal type name.`), so the ambiguity this PR added is gone. The remaining overlap is the pre-existing length rejection vs the deny rejection; changing that one is a separate call with its own message-compatibility surface, so I left it alone rather than folding it into a security fix.

If the accept-list case is worth a tailored message, the right shape is to raise it only when a prefix match on a deny class was actually skipped, rather than on the shared path — happy to do that as a follow-up if you want it.

wenshao:
## 总体

三层加固的方向是对的，尤其是「白名单前缀匹配反而比没有白名单更宽松」这个不对称性，确实是真问题。我在本地按 PR HEAD (`0228a1d`) 单独编译了这 4 个 main 源文件，叠在 `core/target/classes` 上做了对照验证：

| 场景 | main | PR 7703 |
|---|---|---|
| `addAutoTypeAccept("java.net.")` → `java.net.URLClassLoader`（无 SupportAutoType） | `URLClassLoader` | `null` |
| 同上（开 SupportAutoType） | `URLClassLoader` | throw |
| `addAutoTypeAccept("javax.sql.")` → `javax.sql.DataSource` | `DataSource` | `null` |
| `addAutoTypeAccept("java.net.URLClassLoader")`（全名）→ 同类 | `URLClassLoader` | `URLClassLoader` |
| `addAutoTypeAccept("java.net.")` → `java.net.URI`（正常类型） | `URI` | `URI` |

漏洞确实存在、修复也确实生效，全名 accept 的豁免和正常类型的解析都没被误伤。新增的 `AutoTypeValidationTest` 24 个用例本地跑下来全绿。

下面是几个需要处理的点。

---

### 1.【建议改】`:` / `!` 直接抛异常，改变了未知 `@type` 的降级语义

`ObjectReaderProvider.java:845`

fastjson2 目前对解析不出来的 `@type` 是**降级成 Map**，而不是报错（除非开了 `ErrorOnNotSupportAutoType`）。这个 PR 让含 `:` / `!` 的类型名变成无条件抛异常，和 `com.foo.Unknown` 这类「加载不到」的情况行为不一致。

影响最大的是 JSON-LD —— `@type` 是 JSON-LD 的保留关键字，取值是 IRI，必然带 `:`：

```java
String ld = "{\"@type\":\"http://schema.org/Person\",\"name\":\"x\"}";
JSON.parseObject(ld, Object.class);
```

实测对照（`@type` 是第一个 key 时才会走 autoType 分支）：

| 调用 | main | PR 7703 |
|---|---|---|
| `parseObject(ld, Object.class)` | Map | **throw** `autoType is not support, illegal type name.` |
| `parseArray("[" + ld + "]", Object.class)` | List&lt;Map&gt; | **throw** |
| bean 里的 `Object` 字段 | Map | **throw** |
| `parseObject(ld)` / `parseObject(ld, Map.class)` | OK | OK（不变） |

安全上这个 throw 并不必要 —— `TypeUtils.loadClass` 里的同一个检查已经让类加载不到，白名单扫描也进不去，类不可能被解析出来。所以建议 845 行改成 `return null`：

```java
if (hasIllegalTypeNameChars(typeName)) {
    return null;   // 与「类加载不到」保持一致；要不要报错交给 ErrorOnNotSupportAutoType
}
```

这样开 `ErrorOnNotSupportAutoType` 时调用方照样报错，不开时保持原有降级行为。对应 `testRejectColonInTypeName` / `testRejectExclamationInTypeName` / `testRejectColonWithSupportAutoType` 要从 `assertThrows` 改成 `assertNull`。

如果确实想保留硬报错，那至少要在 release notes 里明确列为 breaking change。

---

### 2.【需要写进 release notes】前缀白名单不再放行 DataSource / ClassLoader

这是有意为之，但对用户是行为破坏。最典型的是 1.x 迁移过来的写法：

```java
ParserConfig.getGlobalInstance().addAccept("com.alibaba.druid.pool.");
```

`DruidDataSource implements DataSource`，改动后前缀 accept 不再放行，必须改成写全名。`JSONReader.autoTypeFilter("com.xxx.")` 同理。

建议 2.0.62 release notes 单列一条；另外现在 `SupportAutoType` 分支抛的还是笼统的 `autoType is not support. xxx`，用户很难猜到要改成全名，可以在消息里带一句提示。

---

### 3.【小】`ContextAutoTypeBeforeHandler` 缓存命中时跳过了黑名单检查

`ContextAutoTypeBeforeHandler.java:273-294`

```java
Class clazz = apply(typeNameHash, expectClass, features);   // classCache 命中
if (clazz == null) {
    clazz = loadClass(typeName);
    if (clazz != null) {
        if (i + 1 < typeNameLength && JDKUtils.isAutoTypeDenyClass(clazz)) { continue; }
        ...
    }
}
if (clazz != null) { return clazz; }
```

黑名单检查在 `clazz == null` 分支里面，缓存命中时不走。我把 `putCacheIfAbsent` 的三处调用（前缀分支 / 数组分支 / mapping 分支）都追了一遍，当前只有「全名 accept」才可能把 deny 类放进 classCache，所以**没有实际漏洞**。但这是靠调用方不变式撑着的，建议把检查提到 `if (clazz != null) return clazz;` 之前，跟 `checkAutoType` 的写法对齐，成本可以忽略。

---

### 4.【小】文本校验的分配开销

`ObjectReaderProvider.java:871` / `:906`、`ContextAutoTypeBeforeHandler.java:270`：

```java
if (!acceptNameSet.contains(normalizeAcceptName(typeName.substring(0, i + 1))))
```

只在 hash 命中时执行，而且 `getObjectReader(String, ...)` 成功后会 `registerIfAbsent(Fnv.hashCode64(typeName), reader)`，同一个类型名后续走 hash 缓存，所以不在真正的热路径上，可以接受。

要做到零分配的话：`Arrays.binarySearch` 已经返回了下标，维护一个和 `acceptHashCodes` 同序的 `String[] acceptNames`，用 `regionMatches` 式的逐字符比较（顺带做 `$` 归一化）就行，不需要 substring + Set 查找。副作用是两个 accept 名字 hash 冲突时只有先注册的能通过校验 —— FNV-64 下这个概率可以忽略。可选优化。

---

### 5. 其他

- `JDKUtils.isAutoTypeDenyClass`：`JDKUtils` 的定位是 JDK 内部 / Unsafe 相关，autoType 黑名单策略放这里略微跑偏。不过 `isSQLDataSourceOrRowSet` 本来就在这，就近放置也说得通，看取舍。
- `TypeUtils.hasIllegalTypeNameChars` / `normalizeAcceptName` 是新增的 public API，一旦发布就要长期兼容。因为 `ContextAutoTypeBeforeHandler` 在 `filter` 包，收不成包级可见，建议 javadoc 里注明 internal。
- `normalizeAcceptName` 把 `$` 归一成 `.` 之后，`addAutoTypeAccept("a.b.Outer$Inner")` 和 `addAutoTypeAccept("a.b.Outer.Inner")` 变成等价了。这是跟随 `checkAutoType` 滚动 hash 的既有语义，没问题，但值得在 javadoc 点一句。
- `testParseWithAutoTypeFilterNormal` 只断言了 `assertNotNull(obj)`，降级成 JSONObject 时也会通过，建议补一个 `assertInstanceOf(HashMap.class, obj)`。
- `testLoadClassDoesNotReachClassLoader` 改了线程 TCCL，finally 里恢复了、surefire 也没开并行，OK。

---

### 并发那段

`addAutoTypeAccept` 先发布 `acceptNameSet` 再发布 `acceptHashCodes`、读侧先读 hash 数组的顺序是对的：两个字段都是 volatile，写侧程序序 + volatile 写的 happens-before 传递性保证了读侧看到新 hash 数组就一定能看到对应的 name，不会出现「hash 命中但 name 还没到」的瞬时误拒。注释也留下了，很好。

---

**结论**：核心加固逻辑站得住，测试也扎实。建议先处理 1（改成 `return null` + 相应改测试），2 补 release notes，3 顺手对齐；4、5 可选。

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
```diff
@@ -0,0 +1,341 @@
...
+                JSONReader.autoTypeFilter(PACKAGE_PREFIX));
+        assertEquals(123, bean.id);
+    }
+
+    @Test
+    public void testParseWithAutoTypeFilterNormal() {
+        String json = "{\"@type\":\"java.util.HashMap\",\"value\":1}";
+        Object obj = JSON.parseObject(json, Object.class,
+                JSONReader.autoTypeFilter("java.util.HashMap"));
+        assertNotNull(obj);
```
wenshao: **[Suggestion]** `assertNotNull(obj)` cannot distinguish correct whitelist acceptance from silent rejection — when `acceptNameSet` is empty (or the text-verification check is removed), `JSON.parseObject` falls back to a non-null `JSONObject`, and this assertion still passes. — Concrete cost: mutation "remove `acceptNameSet.contains()` check" leaves this test green; the end-to-end acceptance path has no independent guard here.

```suggestion
        Object obj = JSON.parseObject(json, Object.class,
                JSONReader.autoTypeFilter("java.util.HashMap"));
        assertEquals(java.util.HashMap.class, obj.getClass());
```

Note: `assertTrue(obj instanceof java.util.HashMap)` would also pass in the fallback case, because `JSONObject extends LinkedHashMap extends HashMap`. Use `assertEquals(HashMap.class, obj.getClass())` instead.

wenshao (commented):
Reviewed. Suggestions are inline. Not reviewed: 6a — attacker persona agent failed twice (subagent execution error).

wenshao:
Thanks — the JSON-LD case is a real break and I had missed it. Reproduced it before changing anything:

```
parseObject(ld, Object.class)  THREW  autoType is not support, illegal type name. http://schema.org/Person
parseArray("[" + ld + "]", …)  THREW  same
parseObject(ld) / Map.class    OK
```

All four points are handled in `a80aa4e`.

### 1. `:` / `!` → `return null`

Done. Your reasoning is right on both halves: it is inconsistent with `com.foo.Unknown`, and the throw buys nothing — `TypeUtils.loadClass` rejects the same names, so the class cannot be resolved either way. Reporting it as unresolved keeps `ErrorOnNotSupportAutoType` in charge of whether the caller sees an exception.

`testRejectColonInTypeName` / `testRejectExclamationInTypeName` / `testRejectColonWithSupportAutoType` are now `assertNull`, plus a `testParseJsonLdFallsBackToMap` that pins the JSON-LD path end to end: map for `parseObject(Object.class)`, map for `parseArray`, and an exception under `ErrorOnNotSupportAutoType`.

No breaking-change note needed for this one any more — with `return null` the only behaviour that changes versus main is that a `:`/`!` name can no longer resolve a class, which was the point.

### 2. Accept-prefix change → documented

Added a section to `docs/autotype_cn.md` and `docs/autotype_en.md` (§2.1) with exactly your Druid example, covering `addAutoTypeAccept`, `-Dfastjson2.autoTypeAccept`, `autoTypeFilter` and 1.x `ParserConfig.addAccept`. Flagged as breaking in the PR description too — the GitHub release notes are yours to write, so please pull it from there.

On the message: I had declined the earlier inline suggestion because that string fires for **every** deny-class rejection, including untrusted input naming a gadget with no accept list in play, where "add the full type name to accept" is the one thing the operator must not do. Since you asked again I implemented the precise version instead — a flag set only when an accept prefix actually matched a deny class:

```
no accept entry:              autoType is not support. …$TestClassLoader
addAutoTypeAccept("com.…."):  autoType is not support, an accept prefix does not cover it,
                              add the type name in full to accept. …$TestClassLoader
```

So the actionable hint reaches exactly the misconfigured-accept-list case and nothing else. `testDenyPrefixOnlyRejectionIsDistinguishable` asserts both messages.

### 3. Cached class skipped the blacklist

Fixed — the check moved below the cache read, so it covers the cache-hit path. Your invariant analysis matches mine (only a full-name accept can put a deny class in `classCache`, and the array/mapping branches recurse through `apply`), so there was no live hole, but the check no longer depends on that. Caching a deny class is now possible and harmless, since every read is re-checked. `testContextHandlerChecksCachedClass` injects one into `classCache` and asserts the prefix match still rejects it.

### 4. Allocation — leaving as is

Agreed it is off the hot path, and the parallel `String[]` variant has the downside you noted: two accept names sharing a hash bucket would leave only the first verifiable. The `Set` has no such failure mode. Not worth trading that for an allocation that only happens on a hash hit.

### 5. Other

- `TypeUtils.hasIllegalTypeNameChars` / `normalizeAcceptName` — javadoc now says internal, public only because the entry points are in different packages. Also spelled out that `$` and `.` spellings are equivalent as accept entries.
- `testParseWithAutoTypeFilterNormal` — now `assertEquals(HashMap.class, obj.getClass())`.
- `JDKUtils.isAutoTypeDenyClass` placement — left next to `isSQLDataSourceOrRowSet`. Moving it means either a second util class in the chain or duplicating the SQL class lookup; happy to relocate if you would rather it live elsewhere.

### Verification

- `AutoTypeValidationTest` 27 cases. Re-ran the mutation check over everything touched here: format guard in `checkAutoType` and in `loadClass`, the `denyPrefixOnly` flag and its message, and the handler's blacklist check and full-name exemption — all six fail a test when removed.
- `core` 7933 pass, `fastjson1-compatible` 1346, `safemode-test` 2, checkstyle clean.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
wenshao: Reopening my own reply — implemented in `a80aa4e`, in the precise form rather than on the shared path.

A `denyPrefixOnly` flag is set only when an accept prefix actually matched a deny class, and the rejection reads differently in that case:

```
no accept entry:              autoType is not support. …$TestClassLoader
addAutoTypeAccept("com.…."):  autoType is not support, an accept prefix does not cover it,
                              add the type name in full to accept. …$TestClassLoader
```

That reaches exactly your `addAutoTypeAccept("com.zaxxer.hikari.")` case while keeping the generic wording for untrusted input naming a gadget with no accept list in play — where the remediation advice would have been actively wrong. `testDenyPrefixOnlyRejectionIsDistinguishable` asserts both.

Also note the other half of my earlier reply no longer applies: the format check now returns null instead of throwing (it was breaking JSON-LD `@type` IRIs), so it has no message at all.

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeValidationTest.java
wenshao: Already fixed in `a80aa4e`, and in exactly the form you recommend:

```java
// not merely non-null, falling back to a JSONObject would satisfy that too
assertEquals(HashMap.class, obj.getClass());
```

Good catch on the `instanceof` trap — `JSONObject extends LinkedHashMap extends HashMap`, so `instanceof HashMap` would have passed in the fallback case and left the test just as blind. Identity on `getClass()` is what discriminates.

The one place I do use `assertInstanceOf(Map.class, …)` is `testParseJsonLdFallsBackToMap`, where asserting *any* map is the point — the fallback is the expected outcome there, and the test discriminates because the pre-fix code threw.

PR #7718, opened 2026-07-29
fix: initialize MethodHandles.Lookup before reading IMPL_LOOKUP (#7691)

### LINKED ISSUE #7691: [BUG] Jdk1.8 JSON.toJSONString(obj) 抛出异常 LambdaConversionException: Invalid caller
### 问题描述
*简要描述您碰到的问题。*
//JSON.toJSONString(people, "yyyy-MM-dd")// 不报错
JSON.toJSONString(people)// 报错，Caused by: java.lang.invoke.LambdaConversionException: Invalid caller: com.example.fastjson2.entity.People
JSONObject.toJSONString(people)// 不报错

调换上面两行代码的顺序，就都不报错了

### 环境信息
*请填写以下信息：*

 - OS信息：  [e.g.：Windows 11 专业版 21H2
 - JDK信息： [e.g.：Oracle JDK 1.8.0_131]
 - 版本信息：[e.g.：Fastjson2 2.0.61]
 

### 重现步骤
1. 使用 `JSON.toJSONString(obj)` 方法
2. 出现 `LambdaConversionException: Invalid caller` 错误

4. 代码示例
```java
import com.alibaba.fastjson2.JSONObject;
import lombok.Data;
import java.util.Date;

@Data
public class People {
	private String name;
	private Integer age;
	private Date birthday;

		People people = new People();
		people.setName("lis");
		people.setBirthday(new Date());
		System.out.println(JSON.toJSONString(people));
		System.out.println(JSONObject.toJSONString(people));
	}
}
```

### 期待的正确结果
希望JSON.toJSONString(people)方法不要报错，或者有什么开关绕过Lambda优化

### 相关日志输出
Exception in thread "main" java.lang.RuntimeException: Failed to create lambda for method: public java.lang.String com.example.fastjson2.entity.People.getName()
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactoryLambda.getObject(PropertyAccessorFactoryLambda.java:513)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactoryLambda.create(PropertyAccessorFactoryLambda.java:310)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactory.create(PropertyAccessorFactory.java:2806)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactory.create(PropertyAccessorFactory.java:2791)
	at com.alibaba.fastjson2.writer.FieldWriter.createPropertyAccessor(FieldWriter.java:222)
	at com.alibaba.fastjson2.writer.FieldWriter.<init>(FieldWriter.java:194)
	at com.alibaba.fastjson2.writer.FieldWriterString.<init>(FieldWriterString.java:25)
	at com.alibaba.fastjson2.writer.ObjectWriterCreator.createFieldWriter(ObjectWriterCreator.java:1428)
	at com.alibaba.fastjson2.writer.ObjectWriterCreatorASM.lambda$createObjectWriter$1(ObjectWriterCreatorASM.java:338)
	at com.alibaba.fastjson2.util.BeanUtils.getters(BeanUtils.java:1152)
	at com.alibaba.fastjson2.writer.ObjectWriterCreatorASM.createObjectWriter(ObjectWriterCreatorASM.java:244)
	at com.alibaba.fastjson2.writer.ObjectWriterProvider.getObjectWriterInternal(ObjectWriterProvider.java:720)
	at com.alibaba.fastjson2.writer.ObjectWriterProvider.getObjectWriter(ObjectWriterProvider.java:582)
	at com.alibaba.fastjson2.JSON.toJSONString(JSON.java:3228)
	at com.example.fastjson2.entity.People.main(People.java:24)
Caused by: java.lang.invoke.LambdaConversionException: Invalid caller: com.example.fastjson2.entity.People
	at java.lang.invoke.AbstractValidatingLambdaMetafactory.<init>(AbstractValidatingLambdaMetafactory.java:118)
	at java.lang.invoke.InnerClassLambdaMetafactory.<init>(InnerClassLambdaMetafactory.java:155)
	at java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:299)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactoryLambda.getObject(PropertyAccessorFactoryLambda.java:504)

wenshao:
## Problem

Since 2.0.61, `JSON.toJSONString(bean)` throws `LambdaConversionException: Invalid caller` on JDK 8 for any Bean with getters, making Bean serialization unusable.

## Root cause

`JDKUtils` reads `MethodHandles.Lookup.IMPL_LOOKUP` straight from memory with `Unsafe`:

```java
Class lookupClass = MethodHandles.Lookup.class;
Field implLookup = lookupClass.getDeclaredField("IMPL_LOOKUP");
long fieldOffset = UNSAFE.staticFieldOffset(implLookup);
trustedLookup = (MethodHandles.Lookup) UNSAFE.getObject(lookupClass, fieldOffset);
```

Neither `.class` nor `getDeclaredField` triggers class initialization, so if nothing has used `MethodHandles` before fastjson2 is loaded, `Lookup.<clinit>` has not run yet and the field is still **null**. Verified on JDK 8:

```
IMPL_LOOKUP before Lookup.<clinit>:        null
IMPL_LOOKUP after  MethodHandles.lookup(): /trusted   lookupModes = 15
```

From there the whole chain degrades:

1. `IMPL_LOOKUP` is null, so `JDKUtils` falls back to `MethodHandles.lookup()`, whose `lookupClass` is `JDKUtils` — not trusted.
2. `trustedLookup(Class)` therefore cannot find the private `Lookup(Class, int)` constructor, and sets the sticky `CONSTRUCTOR_LOOKUP_ERROR` flag.
3. Every later `trustedLookup()` call returns `IMPL_LOOKUP.in(beanClass)`; `Lookup.in()` drops the PRIVATE bit.
4. `LambdaMetafactory` rejects a caller without private access: `LambdaConversionException: Invalid caller`.

This never reproduces under surefire, because JUnit initializes `MethodHandles` long before fastjson2 is loaded — which is why the `Test JDK 8` CI jobs stayed green on `main`.

## Fix

**`JDKUtils` (the actual fix)** — call `MethodHandles.lookup()` before the `Unsafe` read so `MethodHandles.Lookup` is initialized and `IMPL_LOOKUP` is the real trusted lookup. Also read it from `UNSAFE.staticFieldBase(implLookup)` instead of assuming the static-field base is the `Class` mirror.

**`PropertyAccessorFactoryLambda` (safety net)** — for environments where no trusted lookup can be obtained at all (Android, SecurityManager, GraalVM, a future JDK without `Unsafe`):

- check the lookup's PRIVATE mode up front and go straight to the reflection accessors, so a degraded JVM does not throw once per property;
- extract the lambda path into `createLambdaAccessor()` and fall back to `super.create()` if it fails anyway, rethrowing `VirtualMachineError`;
- `getObject(Method)` now uses `lookup(Method)` like every other accessor method, instead of calling `JDKUtils.trustedLookup()` directly.

## Verification

On JDK 8 (zulu8.0.502), a plain `main` calling `JSON.toJSONString(people)` reproduced the exact stack trace of #7691 before the change:

```
RuntimeException: Failed to create lambda for method: public java.lang.String Repro$People.getName()
Caused by: LambdaConversionException: Invalid caller: Repro$People
```

After the change it serializes normally, and `JDKUtils.trustedLookup(Bean.class)` returns `/trusted` with `lookupModes = 15`, so the **lambda fast path is used** rather than the reflection fallback — no performance regression.

Test suites: `core` 7956 tests pass, and also with `-Dfastjson2.creator=reflect`; `core` + `extension` + `fastjson1-compatible` + `kotlin` + `safemode-test` 9397 tests pass.

## Tests

`Issue7691` covers serialization/round-trip for String/Integer/Date/int properties and non-public beans, plus two tests that pin the fallback with a factory whose `lookup()` has no private access and one whose lambda creation fails. Both fail with the `LambdaConversionException` from the issue when the `PropertyAccessorFactoryLambda` change is reverted.

Note: `testTrustedLookupHasPrivateAccess` asserts the invariant the lambda path depends on, but it cannot fail in a surefire JVM, since `MethodHandles.Lookup` is already initialized there. The `JDKUtils` fix was verified manually on JDK 8 as described above.

Closes #7691

### DIFF
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7691.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7691.java
@@ -1,0 +1,246 @@
+package com.alibaba.fastjson2.issues_7000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
+import com.alibaba.fastjson2.introspect.PropertyAccessor;
+import com.alibaba.fastjson2.introspect.PropertyAccessorFactoryLambda;
+import com.alibaba.fastjson2.util.JDKUtils;
+import org.junit.jupiter.api.Test;
+
+import java.lang.invoke.LambdaConversionException;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Method;
+import java.util.Date;
+import java.util.function.Function;
+import java.util.function.ToIntFunction;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue7691 {
+    public static class People {
+        private String name;
+        private Integer age;
+        private Date birthday;
+
+        public String getName() {
+            return name;
+        }
+
+        public void setName(String name) {
+            this.name = name;
+        }
+
+        public Integer getAge() {
+            return age;
+        }
+
+        public void setAge(Integer age) {
+            this.age = age;
+        }
+
+        public Date getBirthday() {
+            return birthday;
+        }
+
+        public void setBirthday(Date birthday) {
+            this.birthday = birthday;
+        }
+    }
+
+    public static class Bean {
+        private int age;
+
+        public int getAge() {
+            return age;
+        }
+
+        public void setAge(int age) {
+            this.age = age;
+        }
+    }
+
+    static class NonPublicBean {
+        private String name;
+        private int value;
+
+        public String getName() {
+            return name;
+        }
+
+        public void setName(String name) {
+            this.name = name;
+        }
+
+        public int getValue() {
+            return value;
+        }
+
+        public void setValue(int value) {
+            this.value = value;
+        }
+    }
+
+    /**
+     * Reproduces the degraded lookup of #7691: {@code JDKUtils.trustedLookup} fell back to
+     * {@code IMPL_LOOKUP.in(beanClass)}, which drops the PRIVATE bit, and LambdaMetafactory then
+     * rejects the call with {@code LambdaConversionException: Invalid caller}.
+     */
+    static class NoPrivateAccessFactory
+            extends PropertyAccessorFactoryLambda {
+        @Override
+        protected MethodHandles.Lookup lookup(Class<?> declaringClass) {
+            return MethodHandles.publicLookup().in(declaringClass);
+        }
+    }
+
+    /**
+     * Has a lookup with private access, so the lambda path is entered, but the lambda creation itself
+     * fails the way LambdaMetafactory does on an affected JVM.
+     */
+    static class LambdaFailingFactory
+            extends PropertyAccessorFactoryLambda {
+        @Override
+        protected MethodHandles.Lookup lookup(Class<?> declaringClass) {
+            return MethodHandles.lookup().in(declaringClass);
+        }
+
+        @Override
+        public Function<Object, Object> getObject(Method method) {
+            throw lambdaFailure(method);
+        }
+
+        @Override
+        public ToIntFunction<Object> getInt(Method method) {
+            throw lambdaFailure(method);
+        }
+
+        private static RuntimeException lambdaFailure(Method method) {
+            return new RuntimeException(
+                    "Failed to create lambda for method: " + method,
+                    new LambdaConversionException("Invalid caller: " + method.getDeclaringClass().getName()));
+        }
+    }
+
+    /**
+     * The invariant the whole lambda path depends on: trustedLookup must have private access,
+     * otherwise LambdaMetafactory rejects every getter/setter. It was violated in #7691 because
+     * IMPL_LOOKUP was read before MethodHandles.Lookup was initialized, which only happens when
+     * nothing has used MethodHandles before fastjson2 is loaded - never the case in a test JVM.
+     */
+    @Test
+    public void testTrustedLookupHasPrivateAccess() {
+        MethodHandles.Lookup lookup = JDKUtils.trustedLookup(People.class);
+        assertTrue((lookup.lookupModes() & MethodHandles.Lookup.PRIVATE) != 0, "lookupModes=" + lookup.lookupModes());
+    }
+
+    @Test
+    public void testFallbackWithoutPrivateAccess() throws Exception {
+        assertFallback(new NoPrivateAccessFactory());
+    }
+
+    @Test
+    public void testFallbackOnLambdaConversionFailure() throws Exception {
+        assertFallback(new LambdaFailingFactory());
+    }
+
+    /**
+     * The factory must return a working accessor instead of throwing
+     * {@code RuntimeException: Failed to create lambda for method}.
+     */
+    private static void assertFallback(PropertyAccessorFactoryLambda factory) throws Exception {
+        Method getName = People.class.getMethod("getName");
+        Method setName = People.class.getMethod("setName", String.class);
+        PropertyAccessor name = factory.create("name", String.class, String.class, getName, setName, null);
+        assertNotNull(name);
+
+        People people = new People();
+        name.setObject(people, "lis");
+        assertEquals("lis", people.getName());
+        assertEquals("lis", name.getObject(people));
+
+        Method getAge = Bean.class.getMethod("getAge");
+        Method setAge = Bean.class.getMethod("setAge", int.class);
+        PropertyAccessor age = factory.create("age", int.class, int.class, getAge, setAge, null);
+        assertNotNull(age);
+
+        Bean bean = new Bean();
+        age.setIntValue(bean, 42);
+        assertEquals(42, bean.getAge());
+        assertEquals(42, age.getIntValue(bean));
+    }
+
+    @Test
+    public void testToJSONString() {
+        People people = new People();
+        people.setName("lis");
+        people.setAge(20);
+        people.setBirthday(new Date());
+
+        String json = JSON.toJSONString(people);
+        assertNotNull(json);
+        assertTrue(json.contains("\"name\":\"lis\""));
+        assertTrue(json.contains("\"age\":20"));
+        assertTrue(json.contains("\"birthday\":"));
+    }
+
+    @Test
+    public void testRoundTrip() {
+        People people = new People();
+        people.setName("lis");
+        people.setAge(20);
+        people.setBirthday(new Date());
+
+        String json = JSON.toJSONString(people);
+        People p2 = JSON.parseObject(json, People.class);
+        assertEquals("lis", p2.getName());
+        assertEquals(20, p2.getAge());
+        assertNotNull(p2.getBirthday());
+    }
+
+    @Test
+    public void testToJSONStringPrimitiveInt() {
+        Bean bean = new Bean();
+        bean.setAge(1);
+
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"age\":1}", json);
+    }
+
+    @Test
+    public void testRoundTripPrimitiveInt() {
+        Bean bean = new Bean();
+        bean.setAge(42);
+
+        String json = JSON.toJSONString(bean);
+        Bean b2 = JSON.parseObject(json, Bean.class);
+        assertEquals(42, b2.getAge());
+    }
+
+    @Test
+    public void testNonPublicBean() {
+        NonPublicBean bean = new NonPublicBean();
+        bean.setName("test");
+        bean.setValue(99);
+
+        String json = JSON.toJSONString(bean);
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7691.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        public void setAge(int age) {
+            this.age = age;
+        }
+    }
+
+    @Test
+    public void testToJSONString() {
+        People people = new People();
```
wenshao, line 176: **[Suggestion]** No test exercises the catch/fallback path — the actual behavioral change in this diff. The test beans (`People`, `Bean`) are `public static` inner classes with standard getters/setters, for which `LambdaMetafactory.metafactory()` succeeds on all JDKs including JDK 8. The `catch (Throwable ignored)` block at line 313 of `PropertyAccessorFactoryLambda.java` — the entire fix — is dead code from the tests' perspective on every JDK. — Failure scenario: if the reflection fallback produces incorrect results for a bean that triggers `LambdaConversionException` (e.g., a non-public class or non-standard classloader), no test would catch it. No test exercises deserialization (`JSON.parseObject`) either.

Consider adding round-trip deserialization assertions (e.g., `People p2 = JSON.parseObject(json, People.class); assertEquals("lis", p2.getName());`). For the fallback path specifically, a test that directly invokes `PropertyAccessorFactoryLambda.create()` with a `Method` from a package-private or otherwise LambdaMetafactory-hostile class would exercise the catch block on any JDK.

wenshao (commented):

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7691.java
wenshao: Good suggestions. Addressed in 5f14bbe9d:

1. **Round-trip deserialization**: Added `testRoundTrip` and `testRoundTripPrimitiveInt` with `JSON.parseObject` assertions verifying field values survive serialization → deserialization.

2. **Non-public class**: Added `NonPublicBean` (package-private) with `testNonPublicBean` covering both serialization and deserialization. On JDK 8 this exercises the reflection fallback path since LambdaMetafactory rejects the trusted lookup for non-public classes. On JDK 11+ the trusted lookup (via Unsafe) succeeds even for non-public classes, so the lambda path is taken — this is a JDK 8-specific limitation that would require JDK 8 CI to directly verify the catch block.

wenshao:
## Code Review

### Overview

Wraps the LambdaMetafactory-based accessor creation in `PropertyAccessorFactoryLambda.create(...)` with `try/catch (Throwable)` so that a lambda-creation failure falls through to `super.create(...)` (reflection), mirroring the existing fallback in `createSupplier(Constructor)`. Adds `Issue7691` with 7 serialization/round-trip tests.

The direction is right — defense in depth here is clearly better than a hard `RuntimeException` that makes bean serialization unusable — but there are a few things worth addressing before merge.

---

### 1. Root cause description is not accurate (and points at a better fix)

The PR body says JDK 8 rejects "a trusted lookup whose `lookupClass` is the Bean class". That isn't what happens: JDK 8's `AbstractValidatingLambdaMetafactory` throws
`LambdaConversionException("Invalid caller: " + caller.lookupClass().getName())` only when `!caller.hasPrivateAccess()`. A genuinely TRUSTED lookup (`allowedModes == -1`) has the PRIVATE bit set and passes that check.

So the reported failure means `JDKUtils.trustedLookup(declaringClass)` did **not** return a trusted lookup — it fell through to `JDKUtils.java:519`:

```java
return IMPL_LOOKUP.in(objectClass);   // Lookup.in() drops PRIVATE/PROTECTED
```

That also explains the odd behaviour in #7691 ("swapping the two lines makes both work", `JSONObject.toJSONString` OK but `JSON.toJSONString` not): `CONSTRUCTOR_LOOKUP_ERROR` is a sticky static — once anything sets it, **every** subsequent `trustedLookup()` call degrades to the non-private `in()` lookup for the rest of the JVM's life.

This suggests a more targeted fix, in addition to (or instead of) the blanket catch: check the lookup's capability *before* calling `LambdaMetafactory`, e.g. in `PropertyAccessorFactoryLambda.lookup(Class)` / at the top of `create`:

```java
MethodHandles.Lookup lookup = JDKUtils.trustedLookup(declaringClass);
if ((lookup.lookupModes() & MethodHandles.Lookup.PRIVATE) == 0) {
    return super.create(...);   // no private access -> lambda path can never work
}
```

That fixes the whole class of problem (getters, setters, `createFunction`, and any future call site) rather than catching after the fact, and costs no exceptions.

### 2. Failure is re-attempted for every property

As written, on an affected JVM every single property pays: build MethodHandle → `LambdaMetafactory` throws `LambdaConversionException` → wrapped in `RuntimeException` (with stack trace) → caught and discarded. Accessor creation is per-class-cached so it's bounded, but for an app with thousands of beans this is a measurable startup cost for zero benefit — the condition is JVM-global, not per-property.

Suggest memoizing: a `static volatile boolean LAMBDA_UNAVAILABLE` set on first failure, checked before entering the lambda path. (The capability check in §1 achieves the same thing more directly.)

### 3. `catch (Throwable ignored)` swallows too much, silently

- It also swallows genuine programming errors from this path, e.g. `IllegalArgumentException("Method must have exactly one parameter")` in `setObject`, and `validateMethodAndReturnType` failures — those now silently degrade instead of surfacing.
- It swallows `Error`s (`OutOfMemoryError`, `StackOverflowError`, `NoClassDefFoundError`). Consider rethrowing `Error` (or at least `VirtualMachineError`) and catching `Exception` + `LambdaConversionException` only.
- Behaviour change beyond JDK 8: on **any** JVM, a lambda-creation failure now becomes a silent switch to the slow path. That is the desired outcome for users, but it makes misconfiguration (module access, agents, GraalVM) invisible. A one-time debug log, or at minimum documenting `-Dfastjson2.creator=reflect` in the issue as the explicit opt-out the reporter asked for, would help.

### 4. Diff churn

The entire block is re-indented, so a ~5-line change reads as 34 removals / 40 additions and is hard to review. Extracting the body into a private helper keeps the diff (and the nesting) small:

```java
if (!lambda && (setter == null || !isChainableSetter(setter))) {
    try {
        PropertyAccessor accessor = createLambdaAccessor(name, propertyClass, propertyType, getter, setter, exceptionHandler);
        if (accessor != null) {
            return accessor;
        }
    } catch (Throwable ignored) {
        // lambda creation failed, fall back to reflection
    }
}
return super.create(name, propertyClass, propertyType, getter, setter, exceptionHandler);
```

(Minor: `propertyType` is mutated inside the `try` before the failing call; it's the same value `super.create` would derive, so harmless — but a helper avoids the question entirely.)

### 5. Test coverage does not cover the fix ⚠️

This is the main gap. I ran the new tests locally with the `PropertyAccessorFactoryLambda` change reverted:

```
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0  -- com.alibaba.fastjson2.issues_7000.Issue7691
```

All 7 pass **without** the fix. They only assert that ordinary bean serialization works, which was never broken on a healthy JVM — and CI's `Test JDK 8` jobs are green on `main`, so they won't catch a regression here either. The tests document the symptom but guard nothing.

To actually pin the behaviour, force the failure deterministically, e.g. a test-local subclass of the factory whose `lookup(Class)` returns `MethodHandles.lookup().in(declaringClass)` (no private access), then assert `create(...)` still returns a working accessor instead of throwing. Something equivalent for `getObject`/`setObject` would also be valuable.

Smaller test nits:
- `import static ...Assertions.*` — the repo overwhelmingly uses explicit static imports (~1770 files vs ~210 with the wildcard).
- `testCallOrder` doesn't assert anything order-specific; as written it's a duplicate of the two tests above it. If it's meant to encode the ordering symptom from #7691, it needs a fresh classloader/JVM to be meaningful.

### 6. Other

- Branch is based on `82eda3e44` (2.0.63); `main` is now on 2.0.64 — worth rebasing.
- Checkstyle passes; `core` builds and the new tests pass on JDK 26 locally.
- No security concerns: the change only broadens a fallback, and the reflective path is the pre-existing, more conservative one.

---

### Summary

| | |
|---|---|
| Correctness | Fix works, but treats a symptom; the real degradation is `trustedLookup()` falling back to a non-private `IMPL_LOOKUP.in(...)` |
| Performance | Per-property exception on affected JVMs; should be detected once |
| Tests | Pass with the fix reverted — no regression protection |
| Style | Large re-indent diff; blanket `catch (Throwable)` |

Recommendation: keep the fallback as a safety net, but add the up-front lookup-capability check (§1), narrow/memoize the catch (§2–3), and add a test that actually exercises the fallback (§5).

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR. Reviewed.

PR #7719, opened 2026-07-29
fix: the getter method is discarded when serializing character fields

TigerZCoder:
### What is the purpose of this PR/why is it needed?

When serializing beans containing char/Character fields, getter methods are silently discarded, leading to NullPointerExceptions or lost field values ​​(serialized as null or missing). This is because `ObjectWriterCreator` passes null instead of the resolved getter methods when constructing FieldWriterCharValue and FieldWriterChar.

在序列化包含 char/Character 字段的 bean 时，getter 方法会被静默丢弃，导致出现空指针异常或者字段值丢失（序列化为 null 或缺失）。这是因为 `ObjectWriterCreator` 在构造 FieldWriterCharValue 和 FieldWriterChar 时传递了 null 而不是已解析的 getter 方法。

### Summary of Changes

In [ObjectWriterCreator#createFieldWriter](url), the resolved getter methods (instead of null) are now passed to the FieldWriterCharValue and FieldWriterChar constructors for char.class and Character.class fields, respectively.
A new ObjectWriter14Test has been added to reproduce the issue and verify the fix, including: 
- When the optional default method of an interface is not implemented (no corresponding field receives a value), a null pointer exception is thrown in `com.alibaba.fastjson2.writer.ObjectWriterCreatorASM#genGetObject` (because both field and method are null, causing member to be null).
- When getter methods contain (protected) logic, this logic is ignored, leading to inconsistent output.

在 `ObjectWriterCreator#createFieldWriter` 中，将已解析的 getter 方法（而不是 null）分别传递给 char.class 和 Character.class 字段的 FieldWriterCharValue 和 FieldWriterChar 构造函数。
新增 ObjectWriter14Test，可复现问题并验证修复，包含：
- 当接口的可选default方法未实现（没有对应的field接收值）时，会导致`com.alibaba.fastjson2.writer.ObjectWriterCreatorASM#genGetObject`里报空指针异常（因为field和method都是null导致member是null）。
- 当getter方法中有（保护）逻辑时，逻辑被忽略，导致输出与预期不一致。

**If this fix is ​​not the intended design, please explain the design considerations behind forcibly passing null here.**
**如果修复不是设计预期的，请帮忙解释此处强传null的设计考虑。**

### DIFF
--- a/core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
+++ b/core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
@@ -2,6 +2,7 @@ package com.alibaba.fastjson2.writer;
 
 import com.alibaba.fastjson2.*;
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONArray;
 import com.alibaba.fastjson2.JSONB;
 import com.alibaba.fastjson2.JSONObject;
 import com.alibaba.fastjson2.JSONPath;
@@ -23,10 +24,11 @@ public class ObjectWriter14Test {
                 Bean.class,
                 Arrays.asList(
                         ObjectWriters.fieldWriter("f0", Bean::getF0),
-                        ObjectWriters.fieldWriter("f1", Bean::getF1)
+                        ObjectWriters.fieldWriter("f1", Bean::getF1),
+                        ObjectWriters.fieldWriter("f2", Bean::getF2)
                 )
         );
-        assertEquals(2, writerAdapter.getFieldWriters().size());
+        assertEquals(3, writerAdapter.getFieldWriters().size());
     }
 
     @Test
@@ -34,17 +36,23 @@ public class ObjectWriter14Test {
         Bean bean = new Bean();
         bean.setF0('0');
         bean.setF1(null);
+        bean.setF2('A');
         String str = JSON.toJSONString(bean);
         Bean bean1 = JSON.parseObject(str, Bean.class);
         assertEquals(bean.getF0(), bean1.getF0());
         assertEquals(bean.getF1(), bean1.getF1());
         assertEquals(String.valueOf(bean.getF1()), JSON.parseObject(str).getString("f1"));
 
+        // the getter of primitive char f2 returns the lowercase form ('a') while
+        // the backing field holds 'A', so the serialized value must come from the getter
+        assertEquals("a", JSON.parseObject(str).getString("f2"));
+
         JSONObject jsonObject = JSONObject.from(bean);
         assertEquals(str, jsonObject.toString());
 
         assertEquals(bean.getF0(), JSONPath.eval(bean, "$.f0"));
         assertEquals(bean.getF1(), JSONPath.eval(bean, "$.f1"));
+        assertEquals(Character.valueOf('a'), JSONPath.eval(bean, "$.f2"));
         assertNull(JSONPath.eval(bean, "$.f100"));
     }
 
@@ -53,7 +61,14 @@ public class ObjectWriter14Test {
         Bean bean = new Bean();
         bean.setF0('0');
         bean.setF1(null);
+        bean.setF2('A');
         byte[] jsonbBytes = JSONB.toBytes(bean);
+
+        JSONObject jsonObject = JSONB.parseObject(jsonbBytes);
+        assertNull(jsonObject.get("f0"));
+        assertEquals(Character.valueOf('0'), jsonObject.get("f1"));
+        assertEquals(Character.valueOf('a'), jsonObject.get("f2"));
+
         Bean bean1 = JSONB.parseObject(jsonbBytes, Bean.class);
         assertEquals(bean.getF0(), bean1.getF0());
         assertEquals(bean.getF1(), bean1.getF1());
@@ -64,7 +79,14 @@ public class ObjectWriter14Test {
         Bean bean = new Bean();
         bean.setF0('0');
         bean.setF1(null);
+        bean.setF2('A');
         byte[] jsonbBytes = JSONB.toBytes(bean, JSONWriter.Feature.BeanToArray);
+
+        JSONArray array = JSONB.parseArray(jsonbBytes);
+        assertNull(array.get(0));
+        assertEquals(Character.valueOf('0'), array.get(1));
+        assertEquals(Character.valueOf('a'), array.get(2));
+
         Bean bean1 = JSONB.parseObject(jsonbBytes, Bean.class, JSONReader.Feature.SupportArrayToBean);
         assertEquals(bean.getF0(), bean1.getF0());
         assertEquals(bean.getF1(), bean1.getF1());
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
```diff
@@ -0,0 +1,93 @@
...
+        byte[] jsonbBytes = JSONB.toBytes(bean, JSONWriter.Feature.BeanToArray);
+        Bean bean1 = JSONB.parseObject(jsonbBytes, Bean.class, JSONReader.Feature.SupportArrayToBean);
+        assertEquals(bean.getF0(), bean1.getF0());
+        assertEquals(bean.getF1(), bean1.getF1());
+    }
+
+    interface BeanIf {
+        default Character getF0() {
+            return null;
+        }
```
wenshao, line 98: **[Suggestion]** Add coverage for the independently changed primitive `char` branch. All properties in this test return `Character`, so a regression that makes `FieldWriterCharValue` discard the getter would leave every new test green while primitive `char` properties serialize the backing field instead of the getter result. A mutation probe that reverted only the `char.class` branch reproduced exactly that false-negative. Add a primitive `char` property whose getter returns a value different from its backing field and assert the serialized value directly.

请补充对独立修改的 primitive `char` 分支的覆盖。当前测试中的属性返回类型全部是 `Character`，因此即使 `FieldWriterCharValue` 再次丢弃 getter，新增测试仍会全部通过，但 primitive `char` 属性会错误地序列化 backing field，而不是 getter 返回值。验证探针仅回退 `char.class` 分支后成功复现了这一漏检。建议增加一个 getter 返回值与 backing field 不同的 primitive `char` 属性，并直接断言序列化结果。

</details>

_— gpt-5.6-sol via Qwen Code /review_

core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Test
+    public void testJsonb() {
+        Bean bean = new Bean();
+        bean.setF0('0');
+        bean.setF1(null);
+        byte[] jsonbBytes = JSONB.toBytes(bean);
+        Bean bean1 = JSONB.parseObject(jsonbBytes, Bean.class);
+        assertEquals(bean.getF0(), bean1.getF0());
+        assertEquals(bean.getF1(), bean1.getF1());
```
wenshao, line 74: **[Suggestion]** Assert the JSONB payload rather than only round-tripping through the same getter-bearing bean. If `f1` is incorrectly serialized from its `null` backing field, deserialization restores `null`, but `bean1.getF1()` maps that back to `'0'`, so these assertions still pass. A probe using JSONB payloads that explicitly contained `null` confirmed that both this test and `testJsonbArray` stay green. Decode the object/array payload and assert the stored value directly.

请直接断言 JSONB 载荷，而不要只通过同一个带计算 getter 的 Bean 做 round-trip。如果 `f1` 被错误地从 `null` backing field 序列化，反序列化后字段仍为 `null`，但 `bean1.getF1()` 会再次把它转换为 `'0'`，因此这些断言仍然通过。验证探针使用显式包含 `null` 的 JSONB 载荷，确认本测试和 `testJsonbArray` 都会错误保持绿色。建议解码 object/array 载荷并直接断言其中保存的值。

</details>

_— gpt-5.6-sol via Qwen Code /review_

wenshao (commented):

<details>
<summary>中文说明</summary>

已审查——无阻断问题。 建议见行内评论。

</details>

_— gpt-5.6-sol via Qwen Code /review_

wenshao (approved):
No issues found. LGTM! ✅

<details>
<summary>中文说明</summary>

未发现问题。LGTM！✅

</details>

wenshao (approved):
No issues found. LGTM! ✅

<details>
<summary>中文说明</summary>

未发现问题。LGTM！✅

</details>

core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
TigerZCoder: Done. Added a primitive char property `f2` to `Bean` whose backing field is `'A'` while the getter `getF2()` returns the lowercase form `'a'` (`f2 >= 'A' && f2 <= 'Z' ? (char)(f2 + 32) : f2`). `test()` now asserts the serialized value directly (`JSON.parseObject(str).getString("f2") == "a", JSONPath.eval(bean, "$.f2") == 'a'`), and JSONB object/array payloads are covered as well. A mutation probe reverting only the `char.class` branch made all three tests fail with `expected: <a> but was: <A>`, confirming the false-negative is eliminated.

已补充。在 `Bean` 中新增 primitive char 属性 `f2`，backing field 为 `'A'`，getter `getF2()` 返回小写形式 `'a'`（`f2 >= 'A' && f2 <= 'Z' ? (char)(f2 + 32) : f2`），并在 `test()` 中直接断言序列化结果（`JSON.parseObject(str).getString("f2") == "a"、JSONPath.eval(bean, "$.f2"`) 为 `'a'`），同时覆盖 JSONB object 与 array 两种载荷。已用变异探针验证：仅回退 `char.class` 分支后，三个测试均以 `expected: <a> but was: <A>` 失败，说明漏检已被消除。

</details>

core/src/test/java/com/alibaba/fastjson2/writer/ObjectWriter14Test.java
TigerZCoder: Updated. `testJsonb()` now decodes the payload with `JSONB.parseObject(jsonbBytes)` and asserts the stored values directly (`f0` null, `f1` `'0'`, `f2` `'a'`); `testJsonbArray()` asserts the decoded `JSONB.parseArray` elements directly (indices 0/1/2 are null, `'0'`, `'a'`). A mutation probe reverting only the `Character.class` branch confirmed the previous round-trip assertions stayed green (the getter re-maps null back to `'0'`) while the new direct payload assertions fail with `expected: <0> but was: <null>`, so the false-negative is eliminated.

已修改。`testJsonb()` 改为通过 `JSONB.parseObject(jsonbBytes)` 解码载荷后直接断言存储值（`f0` 为 null、`f1` 为 `'0'`、`f2` 为 `'a'`）；`testJsonbArray()` 改为通过 `JSONB.parseArray(jsonbBytes)` 直接断言数组元素（索引 0/1/2 分别为 null、`'0'`、`'a'`）。已用变异探针验证：回退 `Character.class` 分支后，round-trip 断言仍通过（getter 把 null 重新映射为 `'0'`），但直接载荷断言以 `expected: <0> but was: <null>` 失败，证明假阴性已被消除。

</details>

wenshao (approved):
No issues found. LGTM! ✅

<details>
<summary>中文说明</summary>

未发现问题。LGTM！✅

</details>

TigerZCoder:
@wenshao 问下，可以合并了吗？还有什么问题吗？我们涉及这部分问题，等发版更新。

wenshao:
预计周日（8月30日）发版

TigerZCoder:
好的，感谢

PR #7731, opened 2026-07-30
fix: apply autoType whitelist to array elements under ErrorOnNotSupportAutoType (#7730)

### LINKED ISSUE #7730: [BUG]Feature.ErrorOnNotSupportAutoType在list场景下反序列化会报错的问题
### 问题描述
我们正在升级 fastjson2，发现Feature.ErrorOnNotSupportAutoType在list场景下反序列化会报错，而单个元素的则不会

### 环境信息
*请填写以下信息：*
 - 版本信息：fastjson 2.0.63  和 fastjson2 2.0.63 均复现
 

### 重现步骤
```java
        record Data(String id) {
        }
        JSONReader.Context context = JSONFactory.createReadContext(JSONReader.Feature.ErrorOnNotSupportAutoType);
        context.getProvider().addAutoTypeAccept("com.example.");

        // 单元素验证
        Data data = new Data("test-data");
        String jsonEl = JSON.toJSONString(data, JSONWriter.Feature.WriteClassName);
        System.out.println(jsonEl);
        Object objEl = JSON.parse(jsonEl, context);
        System.out.println(objEl);

        // list验证
        List<Data> list = List.of(data);
        String jsonLi = JSON.toJSONString(list, JSONWriter.Feature.WriteClassName);
        System.out.println(jsonLi);
        // 此处报错
        Object objLi = JSON.parse(jsonLi, context);
        System.out.println(objLi);
    }
```

输出
```
{"@type":"com.example.FastJsonTest$1Data","id":"test-data"}
Data[id=test-data]
[{"@type":"com.example.FastJsonTest$1Data","id":"test-data"}]
Exception in thread "main" com.alibaba.fastjson2.JSONException: autoType not support : com.example.FastJsonTest$1Data
	at com.alibaba.fastjson2.JSONReader.readObject(JSONReader.java:3746)
	at com.alibaba.fastjson2.JSONReader.readArray(JSONReader.java:4119)
	at com.alibaba.fastjson2.reader.ObjectReaderImplObject.readObject(ObjectReaderImplObject.java:293)
	at com.alibaba.fastjson2.JSON.parse(JSON.java:234)
	at com.example.FastJsonTest.main(FastJsonTest.java:28)
```

如果去掉Feature.ErrorOnNotSupportAutoType则list部分的反序列化是正常的

JingkaiTang:

Fixes #7730.

When `Feature.ErrorOnNotSupportAutoType` is enabled together with an autoType accept whitelist (`addAutoTypeAccept`), a single object with `@type` is restored correctly, but the **same object nested inside an array** throws `autoType not support`:

```java
JSONReader.Context context = JSONFactory.createReadContext(JSONReader.Feature.ErrorOnNotSupportAutoType);
context.getProvider().addAutoTypeAccept("com.example.");

// single object: OK
JSON.parse(JSON.toJSONString(bean, JSONWriter.Feature.WriteClassName), context);

// list of the same bean: throws "autoType not support"
JSON.parse(JSON.toJSONString(List.of(bean), JSONWriter.Feature.WriteClassName), context);
```

**Root cause:** `JSONReader.readArray()` dispatches object elements to the generic `readObject()` branch unless `autoTypeBeforeHandler` or `SupportAutoType` is set. That generic branch throws unconditionally on `@type` (`JSONReader.java:3744-3746`) **without consulting the accept whitelist**, whereas the single-object path (`ObjectReaderImplObject`) does consult it. Hence single object succeeds but array element fails.

### Summary of your change

- In `JSONReader.readArray()`, also route array object elements through `ObjectReaderImplObject` when `ErrorOnNotSupportAutoType` is enabled, so array elements apply the **same** accept whitelist check as single objects.
- The whitelist stays **exclusive**: types outside the accept prefixes are still rejected (verified by test `notAcceptedTypeInListStillThrows`), so this does not relax security.
- `$ref` handling is unaffected: `ObjectReaderImplObject` already supports references, and existing handler/SupportAutoType scenarios already take this path (covered by `ListRefTest`).
- Default behavior (no autoType feature enabled) is unchanged — still uses the fast `readObject()` path.
- Added regression test `Issue7730` covering single object, top-level list, nested list, list-in-map, and the whitelist-exclusion security case.

Local verification:
- `./mvnw -pl core -Dtest=Issue7730 test` — 6/6 pass
- `./mvnw -pl core -Dtest=ListRefTest test` — pass (no $ref regression)
- `./mvnw -pl core clean test` — 7953/7953 pass
- `./mvnw -pl core -Dfastjson2.creator=reflect -Dtest=Issue7730 test` — pass
- `./mvnw -pl core validate` — 0 Checkstyle violations

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -4111,7 +4111,8 @@ public abstract class JSONReader
                     val = readArray();
                     break;
                 case '{':
-                    if (context.autoTypeBeforeHandler != null || (context.features & Feature.SupportAutoType.mask) != 0) {
+                    if (context.autoTypeBeforeHandler != null
+                            || (context.features & (Feature.SupportAutoType.mask | Feature.ErrorOnNotSupportAutoType.mask)) != 0) {
                         val = ObjectReaderImplObject.INSTANCE.readObject(this, null, null, 0);
                     } else if (isReference()) {
                         val = JSONPath.of(readReference());
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -4111,11 +4111,11 @@ public abstract class JSONReader
                     val = readArray();
                     break;
                 case '{':
-                    if (context.autoTypeBeforeHandler != null
+                    if (isReference()) {
+                        val = JSONPath.of(readReference());
+                    } else if (context.autoTypeBeforeHandler != null
                             || (context.features & (Feature.SupportAutoType.mask | Feature.ErrorOnNotSupportAutoType.mask)) != 0) {
                         val = ObjectReaderImplObject.INSTANCE.readObject(this, null, null, 0);
-                    } else if (isReference()) {
-                        val = JSONPath.of(readReference());
                     } else {
                         val = readObject();
                     }
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7730.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7730.java
@@ -15,10 +15,8 @@ import java.util.List;
 import java.util.Map;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * https://github.com/alibaba/fastjson2/issues/7730
@@ -127,12 +125,33 @@ public class Issue7730 {
 
     @Test
     public void whitelistedListVsRejectedListConsistency() {
-        // Whitelisted element restores; non-whitelisted element in an array is rejected.
+        // Two-sided consistency: a whitelisted element restores, and a non-whitelisted
+        // element in an array is rejected. Both paths must apply the accept whitelist
+        // the same way — the rejection path is what makes ErrorOnNotSupportAutoType
+        // meaningful, so guard it alongside the accept path.
         String accepted = JSON.toJSONString(
                 new ArrayList<>(Arrays.asList(new Bean("x"))), JSONWriter.Feature.WriteClassName);
         Object parsed = JSON.parse(accepted, context());
-        assertTrue(parsed instanceof List);
-        assertFalse(((List<?>) parsed).isEmpty());
         assertInstanceOf(Bean.class, ((List<?>) parsed).get(0));
+
+        String rejected = "[{\"@type\":\"java.io.File\",\"path\":\"/tmp/x\"}]";
+        assertThrows(JSONException.class, () -> JSON.parse(rejected, context()));
+    }
+
+    @Test
+    public void refInArrayUnderErrorOnNotSupportAutoType() {
+        // $ref array element under ErrorOnNotSupportAutoType must still resolve when
+        // the caller uses an API that runs handleResolveTasks (e.g. parseObject with
+        // a Type). Routing array elements through ObjectReaderImplObject for the
+        // autoType whitelist must not bypass the isReference() check — otherwise
+        // {"$ref":"$[0]"} would become a literal JSONObject instead of a reference.
+        JSONReader.Context ctx = JSONFactory.createReadContext(JSONReader.Feature.ErrorOnNotSupportAutoType);
+        ctx.getProvider().addAutoTypeAccept("com.alibaba.fastjson2.issues.");
+        ctx.getProvider().addAutoTypeAccept("java.util.");
+
+        String json = "[{\"id\":\"a\"},{\"$ref\":\"$[0]\"}]";
+        Object parsed = JSON.parseObject(json, Object.class, ctx);
+        List<?> list = (List<?>) parsed;
+        assertEquals(list.get(0), list.get(1));
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7730.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7730.java
@@ -140,17 +140,30 @@ public class Issue7730 {
 
     @Test
     public void refInArrayUnderErrorOnNotSupportAutoType() {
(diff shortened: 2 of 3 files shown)


wenshao (commented):

core/src/test/java/com/alibaba/fastjson2/issues/Issue7730.java
```diff
@@ -0,0 +1,138 @@
...
+        // The fix must not relax the whitelist: a type outside the accept prefix must still
+        // be rejected under ErrorOnNotSupportAutoType, just like the single-object path.
+        String json = "[{\"@type\":\"java.io.File\",\"path\":\"/tmp/x\"}]";
+        assertThrows(JSONException.class, () -> JSON.parse(json, context()));
+    }
+
+    @Test
+    public void whitelistedListVsRejectedListConsistency() {
+        // Whitelisted element restores; non-whitelisted element in an array is rejected.
+        String accepted = JSON.toJSONString(
```
wenshao, line 132: **[Suggestion]** Redundant test — `whitelistedListVsRejectedListConsistency` duplicates `list()` without testing the "vs rejected" consistency its name promises. The body only exercises the whitelisted side (which `list()` already covers more thoroughly with field-value assertions on `id`). The rejected side lives entirely in `notAcceptedTypeInListStillThrows`. — Concrete cost: a maintainer modifying the rejection path would see this test name and assume it guards the consistency contract, but it does not; the test adds CI overhead with zero incremental coverage.

Consider either deleting this test or making it a genuine two-sided consistency check:

```suggestion
    public void whitelistedListVsRejectedListConsistency() {
        // Whitelisted element restores; non-whitelisted element in an array is rejected.
        String accepted = JSON.toJSONString(
                new ArrayList<>(Arrays.asList(new Bean("x"))), JSONWriter.Feature.WriteClassName);
        Object parsed = JSON.parse(accepted, context());
        assertInstanceOf(Bean.class, ((List<?>) parsed).get(0));

        String rejected = "[{\"@type\":\"java.io.File\",\"path\":\"/tmp/x\"}]";
        assertThrows(JSONException.class, () -> JSON.parse(rejected, context()));
    }
```

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -4111,7 +4111,8 @@ public List readArray() {
                     val = readArray();
                     break;
                 case '{':
-                    if (context.autoTypeBeforeHandler != null || (context.features & Feature.SupportAutoType.mask) != 0) {
+                    if (context.autoTypeBeforeHandler != null
+                            || (context.features & (Feature.SupportAutoType.mask | Feature.ErrorOnNotSupportAutoType.mask)) != 0) {
                         val = ObjectReaderImplObject.INSTANCE.readObject(this, null, null, 0);
```
wenshao: **[Suggestion]** Routing all `{`-elements through `ObjectReaderImplObject` when `ErrorOnNotSupportAutoType` is set bypasses the `isReference()` check at line 4117, so `{"$ref":"..."}` array elements are no longer resolved as references — they become plain `JSONObject` maps with a literal `$ref` key. — Failure scenario: a user enables `ErrorOnNotSupportAutoType` (without `SupportAutoType`) and deserializes `[{"id":"a"},{"$ref":"$[0]"}]`; the second element silently becomes `{"$ref":"$[0]"}` instead of a resolved reference to the first. Note: the same bypass already exists for `SupportAutoType` (pre-existing pattern), so this is an extension of existing behavior to a new configuration rather than a new regression.

If preserving `$ref` resolution is desired, check `isReference()` before the autoType routing:

```suggestion
                    if (isReference()) {
                        val = JSONPath.of(readReference());
                    } else if (context.autoTypeBeforeHandler != null
                            || (context.features & (Feature.SupportAutoType.mask | Feature.ErrorOnNotSupportAutoType.mask)) != 0) {
                        val = ObjectReaderImplObject.INSTANCE.readObject(this, null, null, 0);
```

JingkaiTang:
Thanks @wenshao for the review — both suggestions adopted and verified.

**Suggestion 2 (source, `JSONReader.java:4116`)** — accepted, this was a real bug in my first push.

Confirmed by experiment that the original ordering bypassed `isReference()`: under `ErrorOnNotSupportAutoType`, deserializing `[{"id":"a"},{"$ref":"$[0]"}]` produced a literal `JSONObject {"$ref":"$[0]"}` for the second element instead of resolving the reference. My earlier claim that "$ref handling is unaffected" was wrong — apologies.

Fixed by moving `isReference()` before the autoType routing (mirrors the ordering already used by the typed `readArray(Type)` path at line 3913). Re-verified:
- `JSON.parseObject(json, Object.class, ctx)` now resolves the second element to `{"id":"a"}` (matches the first).
- Note: `JSON.parse(String, Context)` does not call `handleResolveTasks` regardless of this fix (pre-existing behavior of that overload), so the reference stays as `null` + pending resolveTask there. This is unchanged by this PR.

**Suggestion 1 (test, `Issue7730.java:131`)** — accepted.

`whitelistedListVsRejectedListConsistency` now actually tests both sides: whitelisted element restores, and `{"@type":"java.io.File"}` in an array is still rejected. Also added `refInArrayUnderErrorOnNotSupportAutoType` as a dedicated regression guard for the `$ref` case above.

Local verification (commit 2):
- `Issue7730`: 7/7 pass
- core full: 7954/7954 pass
- reflect mode (`-Dfastjson2.creator=reflect`): 7/7 pass
- checkstyle: 0 violations

JingkaiTang:
Update on the `$ref` resolution issue — the previous commit (4b01ebcf) fixed it for `parseObject(String, Type, Context)` but I found the same problem also affects `JSON.parse(String, Context)` directly: that overload never calls `handleResolveTasks`, so `$ref` array elements stayed `null` (data loss) regardless of the autoType fix.

Pushed commit 72626bed to address it: add the missing `handleResolveTasks` call in `JSON.parse(String text, JSONReader.Context context)`, mirroring the pattern already used by `parse(String)` (line 145) and `parseObject(String, Type, Context)` (line 1151).

So the full picture across entry points now:

| Entry point | `$ref` resolves? |
|---|---|
| `parse(String)` | ✅ (already did) |
| `parse(String, Context)` | ✅ (after commit 72626bed) |
| `parseObject(String, Type, Context)` | ✅ (already did, and now also works under `ErrorOnNotSupportAutoType` after commit 4b01ebcf) |

Added regression guard `refInArrayViaParseWithContext` covering the middle entry point.

Local verification (commit 3):
- `Issue7730`: 8/8 pass
- core full: 7955/7955 pass
- reflect mode: 8/8 pass
- checkstyle: 0 violations

PR #7737, opened 2026-07-31
fix: bound jsonb array item count to remaining input

Support-itecz:

`JSONReaderJSONB` sizes array allocations from the item count on the wire without checking it against the input. `startArray()` returns the `BC_ARRAY`/`BC_BINARY` count from `readInt32Value()` unbounded, and `readObject()`, `readAny()` and `readArray()` each size a `JSONArray`/`ArrayList` from a `BC_ARRAY` length. The 6 byte payload `A4 48 7F FF FF FF` declares `Integer.MAX_VALUE` items, so `JSONB.parseObject(payload, long[].class)` reaches `new long[Integer.MAX_VALUE]` and dies with `OutOfMemoryError`; every primitive, boxed and object array reader behaves the same, and `0x0FFFFFFF` stays under the 256MB cap in `readLength()` while doing the same through `JSONB.parse` and `List`/`Map` targets.

The count is also what keeps element reads inside the message. Under `JSONB.parseObject(byte[], off, len, type)` the reader's `end` is the frame boundary, not the end of the buffer, so the two byte frame `A4 08` declaring eight items runs the item loop past `end` and returns the eight bytes that follow it. For a framed protocol decoding messages out of a shared or pooled buffer, that hands the caller data belonging to the neighboring frame.

### Summary of your change

A JSONB item always costs at least one byte, so a count larger than the bytes left in the input is malformed by construction. `checkItemCnt()` rejects counts that are negative or larger than `end - offset`, applied at the four places a count comes off the wire, which also covers the array readers that reach it through `startArray()`.

The bound sits in the reader rather than in each `ObjectReaderImpl*Array` because that is where `end` is known, and it follows the existing `checkBigintLen` guard. `JSONBArrayItemCountTest` covers the four paths, the read past `end`, and valid arrays including a 100k element one; without the fix it exhausts the forked test JVM, with it the core suite stays green.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
@@ -2556,7 +2556,7 @@ final class JSONReaderJSONB
                 if (type >= BC_ARRAY_FIX_MIN && type <= BC_ARRAY) {
                     int itemCnt;
                     if (type == BC_ARRAY) {
-                        itemCnt = readInt32Value();
+                        itemCnt = checkItemCnt(readInt32Value());
                     } else {
                         itemCnt = type - BC_ARRAY_FIX_MIN;
                     }
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
@@ -398,7 +398,7 @@ final class JSONReaderJSONB
                             offset++;
                             len = itemType;
                         } else {
-                            len = readLength();
+                            len = checkItemCnt(readLength());
                         }
                     } else {
                         len = valueType - BC_ARRAY_FIX_MIN;
@@ -924,7 +924,7 @@ final class JSONReaderJSONB
 
                 if (type >= BC_ARRAY_FIX_MIN && type <= BC_ARRAY) {
                     int len = type == BC_ARRAY
-                            ? readLength()
+                            ? checkItemCnt(readLength())
                             : type - BC_ARRAY_FIX_MIN;
 
                     if (len == 0) {
@@ -1095,7 +1095,7 @@ final class JSONReaderJSONB
             } else if (valueType >= BC_ARRAY_FIX_MIN && valueType <= BC_ARRAY) {
                 offset++;
                 int len = valueType == BC_ARRAY
-                        ? readLength()
+                        ? checkItemCnt(readLength())
                         : valueType - BC_ARRAY_FIX_MIN;
 
                 if (len == 0) {
@@ -1435,14 +1435,14 @@ final class JSONReaderJSONB
         }
 
         if (type == BC_BINARY) {
-            return readInt32Value();
+            return checkItemCnt(readInt32Value());
         }
 
         if (type != BC_ARRAY) {
             throw new JSONException("array not support input " + error(type));
         }
 
-        return readInt32Value();
+        return checkItemCnt(readInt32Value());
     }
 
     public String error(byte type) {
@@ -6307,6 +6307,17 @@ final class JSONReaderJSONB
         if (len < 0 || len > end - offset) {
             throw new JSONException("BC_BIGINT length out of range: " + len + ", available: " + (end - offset));
         }
+    }
+
+    /**
+     * Every array item occupies at least one byte, so a declared item count larger than the
+     * bytes left in the input cannot be satisfied and the array must not be sized from it.
+     */
+    private int checkItemCnt(int itemCnt) {
+        if (itemCnt < 0 || itemCnt > end - offset) {
+            throw new JSONException("array item count out of range: " + itemCnt + ", available: " + (end - offset));
+        }
+        return itemCnt;
     }
 
     static JSONException outOfBoundsCheckFromToIndex(int offset, int end) {
--- a/core/src/test/java/com/alibaba/fastjson2/JSONBArrayItemCountTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/JSONBArrayItemCountTest.java
@@ -1,0 +1,128 @@
+package com.alibaba.fastjson2;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag("regression")
+@Tag("jsonb")
+public class JSONBArrayItemCountTest {
+    // 0xA4 = BC_ARRAY, 0x48 = BC_INT32, 0x7FFFFFFF = Integer.MAX_VALUE
+    static final byte[] ARRAY_MAX_CNT = {
+            (byte) 0xA4,
+            (byte) 0x48,
+            (byte) 0x7F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
+    };
+
+    // 0x0FFFFFFF stays under readLength()'s 256MB cap but still exceeds the input
+    static final byte[] ARRAY_LARGE_CNT = {
+            (byte) 0xA4,
+            (byte) 0x48,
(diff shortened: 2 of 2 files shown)


core/src/test/java/com/alibaba/fastjson2/JSONBArrayItemCountTest.java
```diff
@@ -0,0 +1,128 @@
...
+    static final byte[] OBJECT_ARRAY_LARGE_CNT = {
+            (byte) 0xA6,
+            (byte) 0x4A, (byte) 0x61,
+            (byte) 0xA4, (byte) 0x48, (byte) 0x0F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF,
+            (byte) 0xA5
+    };
+
+    @Test
+    public void testInt32ValueArrayItemCountOverflow() {
+        assertThrows(JSONException.class, () -> JSONB.parseObject(ARRAY_MAX_CNT, int[].class));
```
wenshao, line 48: **[Suggestion]** The `BC_BINARY` branch in `startArray()` (line 1438 of `JSONReaderJSONB.java`) has no test coverage. All 12 tests use `BC_ARRAY` (0xA4) inputs; none exercise `BC_BINARY` (0x91). The mutation "delete `checkItemCnt` from the `BC_BINARY` line" survives every test in this file. — Concrete cost: a future refactor that accidentally drops the guard on the `BC_BINARY` path would reintroduce the OOM vector for payloads like `{0x91, 0x48, 0x7F, 0xFF, 0xFF, 0xFF}` with no test to catch it.

```suggestion
    @Test
    public void testInt32ValueArrayItemCountOverflow() {
        assertThrows(JSONException.class, () -> JSONB.parseObject(ARRAY_MAX_CNT, int[].class));
    }

    @Test
    public void testBinaryArrayItemCountOverflow() {
        // 0x91 = BC_BINARY, 0x48 = BC_INT32, 0x7FFFFFFF = Integer.MAX_VALUE
        byte[] binaryMaxCnt = {
                (byte) 0x91,
                (byte) 0x48,
                (byte) 0x7F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
        };
        assertThrows(JSONException.class, () -> JSONB.parseObject(binaryMaxCnt, int[].class));
    }
```

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -6309,6 +6309,17 @@ static void checkBigintLen(int len, int offset, int end) {
         }
     }
 
+    /**
+     * Every array item occupies at least one byte, so a declared item count larger than the
+     * bytes left in the input cannot be satisfied and the array must not be sized from it.
+     */
+    private int checkItemCnt(int itemCnt) {
+        if (itemCnt < 0 || itemCnt > end - offset) {
+            throw new JSONException("array item count out of range: " + itemCnt + ", available: " + (end - offset));
```
wenshao, line 6318: **[Suggestion]** No test exercises the boundary `itemCnt == end - offset`. The off-by-one mutation `>` → `>=` survives all 12 tests and would silently reject valid arrays whose declared count exactly equals the remaining bytes. — Concrete cost: a future "simplification" of this condition would break parsing of tightly-packed single-byte-element arrays (e.g. `{0xA4, 0x03, 0x01, 0x02, 0x03}`) with no test to catch it.

Consider adding a boundary test:
```java
@Test
public void testItemCountExactlyMatchesRemainingBytes() {
    // BC_ARRAY, count=3 (single-byte int 0x03), three 1-byte items
    byte[] buffer = {
            (byte) 0xA4, 0x03,
            0x01, 0x02, 0x03
    };
    assertArrayEquals(new int[]{1, 2, 3}, JSONB.parseObject(buffer, int[].class));
}
```

core/src/test/java/com/alibaba/fastjson2/JSONBArrayItemCountTest.java
Support-itecz: Good catch. Added testBinaryArrayItemCountOverflow with the 0x91 payload; int[].class routes through the BC_BINARY branch in startArray(), so dropping that guard now fails the test instead of surviving.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
Support-itecz: Added as testItemCountExactlyMatchesRemainingBytes, so the > vs >= boundary is pinned and a tightly packed array still parses. All 14 tests pass locally.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -1095,7 +1095,7 @@ public List readArray() {
             } else if (valueType >= BC_ARRAY_FIX_MIN && valueType <= BC_ARRAY) {
                 offset++;
                 int len = valueType == BC_ARRAY
-                        ? readLength()
+                        ? checkItemCnt(readLength())
                         : valueType - BC_ARRAY_FIX_MIN;
```
wenshao, line 1099: **[Suggestion]** No error-path test covers this `checkItemCnt` call site (the nested-array element path inside `readArray()`). All other four call sites have dedicated overflow tests. — Concrete cost: if a future refactor accidentally drops `checkItemCnt` from this site only, `new JSONArray(0x0FFFFFFF)` at line 1107 would attempt a ~268M-element allocation (OOM), and no test in this PR would fail.

Suggested test: outer `BC_ARRAY` with count 2, first element a small int (`0x01`), second element `BC_ARRAY` + `BC_INT32` + `0x0FFFFFFF`. Assert `JSONException`.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -398,7 +398,7 @@ public Map<String, Object> readObject() {
                             offset++;
                             len = itemType;
                         } else {
-                            len = readLength();
+                            len = checkItemCnt(readLength());
```
wenshao, line 401: **[Suggestion]** No error-path test covers this `checkItemCnt` call site (the `readObject()` field-value branch). `testReadObjectItemCountOverflow` uses `Map.class`, which routes through `ObjectReaderImplMap` → `startArray()` (line 1445) — it does not reach this line. — Concrete cost: if a future refactor drops `checkItemCnt` here, `JSONB.parseObject(bytes)` (the no-type overload) with a malformed object reaches `new JSONArray(0x0FFFFFFF)` → OOM, and no test catches it. Probe confirmed: removing the guard flips from pass to `OutOfMemoryError`.

Suggested test: `assertThrows(JSONException.class, () -> JSONB.parseObject(OBJECT_ARRAY_LARGE_CNT));`

wenshao (changes requested):
**[Critical]** R1-1 skipValue()'s BC_ARRAY branch (line 2559) reads array item count via bare readInt32Value() without checkItemCnt, leaving the exact vulnerability class this PR fixes (OOM/OOB read from malformed count) reachable through any bean deserialization path that skips an unrecognized field containing an array. Probe confirmed: AIOOBE on unmodified code, clean JSONException with the one-line fix. Suggested fix: itemCnt = checkItemCnt(readInt32Value());

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
Support-itecz: Added as testNestedArrayItemCountOverflow. One note: with a List.class target the elements route through readAny(), which has its own guard, so the test goes through JSONB.parseArray to actually land on this branch. Confirmed via stack trace that the exception comes from the readArray() site.

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
Support-itecz: Added as testReadObjectNoTypeItemCountOverflow using the no-type overload. Stack trace confirms it lands on the field-value branch here rather than startArray().

Support-itecz:
Good catch on skipValue(). Its BC_ARRAY branch now runs the count through checkItemCnt() as well, and testSkipValueItemCountOverflow covers it through a bean with an unknown field holding a truncated array. Without the guard that path reproduces the AIOOBE you describe, with it a clean JSONException. Core suite (7964 tests) and checkstyle both pass.

Support-itecz:
any update?

PR #7739, opened 2026-08-01
fix(core): validate cached field writer by value type in toJSONObject

### LINKED ISSUE #7714: [BUG]Payload为Object或泛型时可能会抛JSONException: key get error
### 问题描述

`JSON.toJSON(Object)`（`ObjectWriterAdapter.toJSONObject` 路径）在复用 `FieldWriterObject` 缓存的 `initObjectWriter` 时**不做值类型校验**，与 `Object` / 泛型字段的"首值类型缓存"机制叠加后会产生错误的 writer 复用：

当某个 bean 含有 `Object`（或方法级泛型 `<T> T`，fieldClass 擦除为 `Object`）的字段时：
1. 先用 `JSON.toJSONString(...)` 序列化该字段值为 **A 类型** 的对象（write 路径会把 `A.class` 及其 `ObjectWriter` CAS 缓存进 `FieldWriterObject.initValueClass` / `initObjectWriter`）；
2. 之后用 `JSON.toJSON(...)`（toJSONObject 路径）序列化该字段值为 **B 类型**（如 `String`）的另一对象时，`ObjectWriterAdapter.toJSONObject` 直接通过 `getInitWriter()` 取出缓存的 **A 的 ObjectWriterAdapter**，在**未校验当前值类型**的情况下用它读取 B 值，最终在 A 的属性访问器内抛出 `ClassCastException`，并被包装成 `JSONException: key get error`。

真实触发场景：MQ 消息封装类（payload 为 `Object`/泛型，如 `public <T> T getPayload()`）。其 ObjectWriter 是全局缓存单例，当同一线程/进程内先用 `toJSONString` 序列化过 bean payload 的消息（如消费端打日志），再对 String payload 的消息调用 `JSON.toJSON`（如发送端打日志）时必现。write 路径不受影响（有 `typeMatch` 校验），仅 `toJSONObject` 路径存在该问题。

### 环境信息

- OS信息：Windows 11 / Linux（均可复现，与 OS 无关）
- JDK信息：OpenJDK 17.0.19（Oracle/OpenJDK 均可复现）
- 版本信息：Fastjson2 2.0.62、**2.0.63（最新版，仍复现）**

### 重现步骤

1. 定义一个含 `Object`（或泛型）字段的 bean（模拟 MQ 消息封装），以及一个业务 bean（含 String 属性 `key`）
2. 先 `JSON.toJSONString(...)` 序列化字段值为业务 bean 的对象（触发字段 writer 缓存）
3. 再 `JSON.toJSON(...)` 序列化字段值为 `String` 的同类型对象
4. 出现 `com.alibaba.fastjson2.JSONException: key get error`，根因 `ClassCastException: java.lang.String cannot be cast to DataChangeNotification`

```java

public class IssueRepro {

    // business bean carried as message payload
    public static class DataChangeNotification {
        private String key;
        private String tenantId;
        public String getKey() { return key; }
        public void setKey(String key) { this.key = key; }
        public String getTenantId() { return tenantId; }
        public void setTenantId(String tenantId) { this.tenantId = tenantId; }
    }

    // MQ message envelope with generic payload accessor
    public static class DefaultMessage {
        private Object payLoad;
        public <T> T getPayload() { return (T) payLoad; }
        public boolean setPayload(Object payload) { this.payLoad = payload; return true; }
    }

        // step 1: write path caches DataChangeNotification + its ObjectWriter
        //         into the payload FieldWriterObject (initValueClass/initObjectWriter)
        DefaultMessage m1 = new DefaultMessage();
        DataChangeNotification dcn = new DataChangeNotification();
        dcn.setKey("K1");
        dcn.setTenantId("10000");
        m1.setPayload(dcn);
        System.out.println("step1: " + JSON.toJSONString(m1));

        // step 2: toJSONObject path reuses the polluted cached writer WITHOUT type check
        DefaultMessage m2 = new DefaultMessage();
        m2.setPayload("{\"key\":\"abc\"}");
        Object r = JSON.toJSON(m2);   // <-- throws here
        System.out.println("step2: " + r);
    }
}
```

### 期待的正确结果

`ObjectWriterAdapter.toJSONObject(T object, long features)` 在通过 `FieldWriterObject.getInitWriter()` 取得缓存的 `ObjectWriterAdapter` 后、调用其 `toJSONObject(fieldValue)` 之前，应校验 `fieldValue` 的实际类型与该 writer 的目标类型是否匹配（与 write 路径 `FieldWriterObject.getObjectWriter(jsonWriter, valueClass)` 中的 `typeMatch` 逻辑一致）；类型不匹配时应按 `fieldValue.getClass()` 重新获取 writer（或直接走 `JSON.toJSON(fieldValue)` 兜底），不应抛出 `ClassCastException`。

### 相关日志输出

```
step1: {"payload":{"key":"K1","tenantId":"10000"}}
com.alibaba.fastjson2.JSONException: key get error
	at com.alibaba.fastjson2.introspect.FunctionAccessor.errorForGet(FunctionAccessor.java:109)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactory$FunctionAccessorString.getString(PropertyAccessorFactory.java:4272)
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactory$PropertyAccessorString.getObject(PropertyAccessorFactory.java:1191)
	at com.alibaba.fastjson2.writer.FieldWriterString.getFieldValue(FieldWriterString.java:30)
	at com.alibaba.fastjson2.writer.ObjectWriterAdapter.toJSONObject(ObjectWriterAdapter.java:618)
	at com.alibaba.fastjson2.writer.ObjectWriterAdapter.toJSONObject(ObjectWriterAdapter.java:610)
	at com.alibaba.fastjson2.writer.ObjectWriterAdapter.toJSONObject(ObjectWriterAdapter.java:691)
	at com.alibaba.fastjson2.JSON.toJSON(JSON.java:4084)
	at com.alibaba.fastjson2.JSON.toJSON(JSON.java:4056)
Caused by: java.lang.ClassCastException: class java.lang.String cannot be cast to class IssueRepro$DataChangeNotification
	at com.alibaba.fastjson2.introspect.PropertyAccessorFactory$FunctionAccessorString.getString(PropertyAccessorFactory.java:4270)
	... 8 more
```

#### 附加信息

根因定位（2.0.62 / 2.0.63 源码一致）：

1. **缓存写入方**：`com.alibaba.fastjson2.writer.FieldWriterObject#getObjectWriterVoid`（约 152-160 行）。`Object`/泛型字段（`initValueClass == null`）首次遇到非 null 值时，会将值的 runtime class 及其 ObjectWriter CAS 缓存到字段 writer 上：
   ```java
   boolean success = initValueClassUpdater.compareAndSet(this, null, valueClass);
   formattedWriter = jsonWriter.getObjectWriter(valueClass);
   if (success) {
       initObjectWriterUpdater.compareAndSet(this, null, formattedWriter);
   }
   ```

2. **缓存使用方（问题点）**：`com.alibaba.fastjson2.writer.ObjectWriterAdapter#toJSONObject`（约 683-691 行）。对 `FieldWriterObject` 字段直接取 `getInitWriter()`，若为非空字段表的 `ObjectWriterAdapter` 就用它处理当前值，**全程没有校验当前值类型**：
   ```java
   ObjectWriter valueWriter = fieldWriter.getInitWriter();          // 可能是缓存的其他类型的 writer
   if (valueWriter == null) {
       valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
   }
       ...
       fieldValue = objectWriterAdapter.toJSONObject(fieldValue);   // 类型不匹配时在这里抛 CCE
   }
   ```

3. **对比**：write 路径 `FieldWriterObject#getObjectWriter(JSONWriter, Class)`（约 77-102 行）有 `typeMatch` 校验，缓存类型与当前值类型不匹配时会走 `getObjectWriterTypeNotMatch` 按实际类型取 writer，因此 `JSON.toJSONString(...)` 不会触发该问题——两条序列化路径对同一缓存的复核逻辑不一致，是本 bug 的本质。

fudianchn:

## What

`ObjectWriterAdapter.toJSONObject(T, long)` now checks whether the field writer cached for an `Object` / generic field still matches the current value's runtime type before reusing it, instead of throwing `JSONException: key get error` (a wrapped `ClassCastException`).

## Why

For a bean with an `Object` (or generic) field, `FieldWriterObject` caches the `ObjectWriter` of the **first** value it serializes (`initValueClass` / `initObjectWriter`). The write path (`FieldWriterObject.getObjectWriter(jsonWriter, valueClass)`) validates this cache via `typeMatch` and re-resolves on mismatch. The `toJSONObject` path did not — it pulled the cached writer via `getInitWriter()` and used it unconditionally, so when a later value had a different runtime type the writer's field accessors threw `ClassCastException`, wrapped as `JSONException: key get error`.

This commonly hits MQ-style envelopes with an untyped `payload` field (`public <T> T getPayload()`): after one `JSON.toJSONString(msg)` with a typed payload, a later `JSON.toJSON(msg)` with a `String` payload on the same bean class blows up.

## Root cause

`ObjectWriterAdapter.java:684` resolved the field value writer via `fieldWriter.getInitWriter()` (the cached writer) with no validation that its `initValueClass` still matches `fieldValue.getClass()`.

## How

- Extract the write path's exact-class `typeMatch` predicate into `FieldWriterObject.isTypeMatch(Class)` and call it from both `ObjectWriterAdapter.toJSONObject` and `FieldWriterObject.getObjectWriter`, so the two paths share one predicate and cannot drift. A subclass value no longer matches the parent's cached writer (matching `toJSONString`), so its fields are no longer silently dropped.
- Right after resolving the cached writer, compare `initValueClass` to the actual value. On mismatch, re-resolve the writer via `JSONFactory.getObjectWriter(fieldValue.getClass(), features)` — mirroring the `typeMatch` re-resolution the write path already performs. When the re-resolved writer is not an `ObjectWriterAdapter` (arrays, `Date`, enums, etc.), convert the value via `JSON.toJSON` so its in-memory shape matches the unprimed path and `toJSONString`.
- When `initValueClass` is `null` (nothing cached yet) or matches, behavior is unchanged.

## Testing

New test `Issue7714` (4 cases). 3 fail before the fix (two with the issue's `JSONException: key get error`, one with silent subclass-field loss), all pass after; the 4th guards the unchanged match path:
- `toJSON_afterWriterCachedForDifferentType` — the issue's scenario: cache a `DataChangeNotification` writer via `toJSONString`, then `JSON.toJSON` a `String` payload. Before: `JSONException: key get error` (`ClassCastException: String cannot be cast to DataChangeNotification`). After: the `String` payload is preserved.
- `toJSON_beanToBeanSubclassReparseAndRecurses` — cache the writer for the parent type, then `JSON.toJSON` a subclass payload (`DataChangeNotificationExt`, adds `scope`). Before: `expected: <online> but was: <null>` (the subclass field was dropped by the parent's cached writer). After: payload is a nested `JSONObject` carrying `scope`, and `toJSON` agrees with `toJSONString`.
- `toJSON_nonAdapterValueAfterBeanCacheConvertsLikeUnprimed` — prime the field with a bean, then send an `int[]` payload whose re-resolved writer is not an `ObjectWriterAdapter`. Before: `JSONException: key get error` (`ClassCastException: [I cannot be cast to DataChangeNotification`). After: payload is a `JSONArray`, matching the unprimed path.
- `toJSON_consistentAfterWriterCachedForSameType` — regression guard: a second value of the cached type still converts to a nested `JSONObject` (passes before and after, guarding the unchanged match path).
- `./mvnw checkstyle:check` → `0 Checkstyle violations`.
- Regression: `JSONTest, JSONObjectTest, JSONReader*Test, *ObjectWriter*Test, Issue*` → `Tests run: 2501, Failures: 0`.

## Verification of the original issue

Before the fix:
```
JSON.toJSON(m2) → JSONException: key get error
  Caused by: ClassCastException: class java.lang.String cannot be cast to class ...DataChangeNotification
```

After the fix:
```
JSON.toJSON(m2) → {"payload":"{\"key\":\"abc\"}"}
```
The `String` payload is preserved as-is instead of being fed to the wrong writer.

Fixes #7714

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
@@ -79,13 +79,7 @@ public class FieldWriterObject<T>
         if (initValueClass == null || initObjectWriter == ObjectWriterBaseModule.VoidObjectWriter.INSTANCE) {
             return getObjectWriterVoid(jsonWriter, valueClass);
         } else {
-            boolean typeMatch = initValueClass == valueClass
-                    || (writeUsing && initValueClass.isAssignableFrom(valueClass))
-                    || (initValueClass == Map.class && initValueClass.isAssignableFrom(valueClass))
-                    || (initValueClass == List.class && initValueClass.isAssignableFrom(valueClass));
-            if (!typeMatch && initValueClass.isPrimitive()) {
-                typeMatch = typeMatch(initValueClass, valueClass);
-            }
+            boolean typeMatch = isTypeMatch(valueClass);
 
             if (typeMatch) {
                 ObjectWriter objectWriter;
@@ -99,6 +93,26 @@ public class FieldWriterObject<T>
                 return getObjectWriterTypeNotMatch(jsonWriter, valueClass);
             }
         }
+    }
+
+    /**
+     * Whether a value of {@code valueClass} is compatible with the writer cached for
+     * {@link #initValueClass}. Mirrors the typeMatch check performed on the write path so that
+     * {@link ObjectWriterAdapter#toJSONObject(Object)} keeps {@code toJSON} consistent with
+     * {@code toJSONString}. Notably this requires an exact class match (plus the writeUsing / Map /
+     * List / primitive-boxing exceptions), so a subclass value does NOT match and must be
+     * re-resolved by its actual class to avoid silently dropping the subclass fields.
+     */
+    public boolean isTypeMatch(Class valueClass) {
+        final Class initValueClass = this.initValueClass;
+        boolean typeMatch = initValueClass == valueClass
+                || (writeUsing && initValueClass.isAssignableFrom(valueClass))
+                || (initValueClass == Map.class && initValueClass.isAssignableFrom(valueClass))
+                || (initValueClass == List.class && initValueClass.isAssignableFrom(valueClass));
+        if (!typeMatch && initValueClass.isPrimitive()) {
+            typeMatch = typeMatch(initValueClass, valueClass);
+        }
+        return typeMatch;
     }
 
     protected final ObjectWriter getObjectWriterVoid(JSONWriter jsonWriter, Class valueClass) {
--- a/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
@@ -102,9 +102,14 @@ public class FieldWriterObject<T>
      * {@code toJSONString}. Notably this requires an exact class match (plus the writeUsing / Map /
      * List / primitive-boxing exceptions), so a subclass value does NOT match and must be
      * re-resolved by its actual class to avoid silently dropping the subclass fields.
+     * Returns {@code false} when {@link #initValueClass} is {@code null} (no value has been
+     * observed for this field yet, so nothing is cached to match against).
      */
     public boolean isTypeMatch(Class valueClass) {
         final Class initValueClass = this.initValueClass;
+        if (initValueClass == null) {
+            return false;
+        }
         boolean typeMatch = initValueClass == valueClass
                 || (writeUsing && initValueClass.isAssignableFrom(valueClass))
                 || (initValueClass == Map.class && initValueClass.isAssignableFrom(valueClass))
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
@@ -697,6 +697,13 @@ public class ObjectWriterAdapter<T>
                 if (objectFieldWriter.initValueClass != null
                         && !objectFieldWriter.isTypeMatch(fieldValue.getClass())) {
                     valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
+                    // When the re-resolved writer is not an ObjectWriterAdapter (arrays, Date, enums,
+                    // etc.), convert the value via JSON.toJSON so it matches the shape the unprimed
+                    // path and toJSONString produce, instead of leaving the raw Java object in the
+                    // JSONObject. See issue #7714.
+                    if (!(valueWriter instanceof ObjectWriterAdapter)) {
+                        fieldValue = JSON.toJSON(fieldValue);
+                    }
                 }
                 if (valueWriter instanceof ObjectWriterAdapter) {
                     ObjectWriterAdapter objectWriterAdapter = (ObjectWriterAdapter) valueWriter;
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
@@ -685,6 +685,19 @@ public class ObjectWriterAdapter<T>
                 if (valueWriter == null) {
                     valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
                 }
+                // The cached writer was selected by the first value seen on this field (e.g. an Object
+                // or generic field whose fieldClass erases to Object). When a later value has a
+                // different runtime type, reusing that writer throws ClassCastException (wrapped as
+                // "key get error"); and a subclass value would be written with the parent writer,
+                // silently dropping the subclass fields. Re-resolve by the actual value class using
+                // the same typeMatch check the write path applies in
+                // FieldWriterObject.getObjectWriter(jsonWriter, valueClass), so toJSON stays
+                // consistent with toJSONString. See issue #7714.
+                FieldWriterObject objectFieldWriter = (FieldWriterObject) fieldWriter;
+                if (objectFieldWriter.initValueClass != null
+                        && !objectFieldWriter.isTypeMatch(fieldValue.getClass())) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
+                }
                 if (valueWriter instanceof ObjectWriterAdapter) {
                     ObjectWriterAdapter objectWriterAdapter = (ObjectWriterAdapter) valueWriter;
                     if (!objectWriterAdapter.getFieldWriters().isEmpty()) {
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
@@ -1,6 +1,7 @@
 package com.alibaba.fastjson2.issues;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONArray;
 import com.alibaba.fastjson2.JSONObject;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -145,5 +146,30 @@ public class Issue7714 {
         // aligning toJSONObject's type-match with the write path).
         String expected = JSON.toJSONString(m2);
         assertEquals(expected, ((JSONObject) r).toJSONString());
+    }
+
+    @Test
+    public void toJSON_nonAdapterValueAfterBeanCacheConvertsLikeUnprimed() {
+        // Prime the payload FieldWriterObject with a bean so initValueClass is set, then send a
+        // value whose writer re-resolves to a non-ObjectWriterAdapter (an int[]). The re-resolved
+        // writer is not an adapter, so the value must still be converted via JSON.toJSON to a
+        // JSONArray — matching the unprimed path and toJSONString — rather than stored as a raw
+        // int[] whose shape depends on cache priming order.
+        DefaultMessage m1 = new DefaultMessage();
+        DataChangeNotification dcn = new DataChangeNotification();
+        dcn.setKey("K1");
+        dcn.setTenantId("10000");
+        m1.setPayload(dcn);
+        JSON.toJSONString(m1);
+
+        DefaultMessage m2 = new DefaultMessage();
+        m2.setPayload(new int[]{1, 2, 3});
+        Object r = JSON.toJSON(m2);
+
+        Object payload = ((JSONObject) r).get("payload");
+        assertTrue(payload instanceof JSONArray, "payload should be a JSONArray, not the raw int[]");
+        JSONArray array = (JSONArray) payload;
+        assertEquals(3, array.size());
+        assertEquals(1, array.getIntValue(0));
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
@@ -2,6 +2,7 @@ package com.alibaba.fastjson2.issues;
 
 import com.alibaba.fastjson2.JSON;
 import com.alibaba.fastjson2.JSONObject;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -15,6 +16,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  * {@code JSONException: key get error} (a wrapped ClassCastException), and — for a subclass
  * value — silently dropping the subclass fields by reusing the parent writer.
  */
+@Tag("regression")
 public class Issue7714 {
     public static class DataChangeNotification {
         private String key;
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
@@ -1,0 +1,147 @@
+package com.alibaba.fastjson2.issues;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Regression for issue #7714: when a bean has an {@code Object} (or generic) field,
+ * {@link com.alibaba.fastjson2.JSON#toJSON(Object)} reused the field writer cached for the
+ * first serialized value type without checking it still matches the current value, raising
+ * {@code JSONException: key get error} (a wrapped ClassCastException), and — for a subclass
+ * value — silently dropping the subclass fields by reusing the parent writer.
+ */
+public class Issue7714 {
+    public static class DataChangeNotification {
+        private String key;
+        private String tenantId;
+
+        public String getKey() {
+            return key;
+        }
+
+        public void setKey(String key) {
+            this.key = key;
+        }
+
+        public String getTenantId() {
+            return tenantId;
+        }
+
+        public void setTenantId(String tenantId) {
+            this.tenantId = tenantId;
+        }
+    }
+
+    // Subclass adds a field; if toJSONObject reused the parent's cached writer it would be lost.
+    public static class DataChangeNotificationExt
+            extends DataChangeNotification {
+        private String scope;
+
+        public String getScope() {
+            return scope;
+        }
+
+        public void setScope(String scope) {
+            this.scope = scope;
+        }
+    }
+
+    // MQ-style envelope with an untyped payload; the generic getter makes fieldClass erase to Object.
+    public static class DefaultMessage {
+        private Object payLoad;
+
+        public <T> T getPayload() {
+            return (T) payLoad;
+        }
+
+        public boolean setPayload(Object payload) {
+            this.payLoad = payload;
+            return true;
+        }
+    }
+
+    @Test
+    public void toJSON_afterWriterCachedForDifferentType() {
+        // Step 1: write path caches a DataChangeNotification writer for the payload FieldWriterObject.
+        DefaultMessage m1 = new DefaultMessage();
+        DataChangeNotification dcn = new DataChangeNotification();
+        dcn.setKey("K1");
+        dcn.setTenantId("10000");
+        m1.setPayload(dcn);
+        JSON.toJSONString(m1);
+
+        // Step 2: toJSONObject path with a String payload must not reuse that cached writer.
+        // The String value must be preserved as-is instead of throwing "key get error".
+        DefaultMessage m2 = new DefaultMessage();
+        m2.setPayload("{\"key\":\"abc\"}");
+        Object r = JSON.toJSON(m2);
+
+        JSONObject json = (JSONObject) r;
+        assertEquals("{\"key\":\"abc\"}", json.get("payload"));
+    }
+
+    @Test
+    public void toJSON_consistentAfterWriterCachedForSameType() {
+        // Regression guard: once the writer is cached for DataChangeNotification, a second
+        // DataChangeNotification payload must still convert to a nested JSONObject.
+        DefaultMessage m1 = new DefaultMessage();
+        DataChangeNotification dcn1 = new DataChangeNotification();
+        dcn1.setKey("K1");
+        dcn1.setTenantId("10000");
+        m1.setPayload(dcn1);
+        JSON.toJSONString(m1);
+
+        DefaultMessage m2 = new DefaultMessage();
+        DataChangeNotification dcn2 = new DataChangeNotification();
+        dcn2.setKey("K2");
+        dcn2.setTenantId("20000");
+        m2.setPayload(dcn2);
+        Object r = JSON.toJSON(m2);
+
+        JSONObject payload = (JSONObject) ((JSONObject) r).get("payload");
+        assertEquals("K2", payload.get("key"));
+        assertEquals("20000", payload.get("tenantId"));
+    }
+
+    @Test
+    public void toJSON_beanToBeanSubclassReparseAndRecurses() {
+        // Step 1: cache the writer for the parent type on the payload FieldWriterObject.
+        DefaultMessage m1 = new DefaultMessage();
+        DataChangeNotification dcn = new DataChangeNotification();
+        dcn.setKey("K1");
+        dcn.setTenantId("10000");
+        m1.setPayload(dcn);
+        JSON.toJSONString(m1);
+
+        // Step 2: payload is now a subclass value. The write path (typeMatch) requires an exact
+        // class match, so toJSONObject must re-resolve the writer by the subclass and recurse,
+        // rather than reuse the parent writer and silently drop DataChangeNotificationExt.scope.
+        DefaultMessage m2 = new DefaultMessage();
+        DataChangeNotificationExt ext = new DataChangeNotificationExt();
+        ext.setKey("K2");
+        ext.setTenantId("20000");
+        ext.setScope("online");
+        m2.setPayload(ext);
+        Object r = JSON.toJSON(m2);
+
+        // The payload must be a nested JSONObject built from the subclass writer, not the raw bean.
+        Object payload = ((JSONObject) r).get("payload");
+        assertTrue(payload instanceof JSONObject, "payload should be a nested JSONObject, not the raw bean");
+        assertNotSame(ext, payload, "payload should be a converted JSONObject, not the original bean");
+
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -685,6 +685,16 @@ public JSONObject toJSONObject(T object, long features) {
...
                     valueWriter = JSONFactory.getObjectWriter(fieldWriter.fieldType, this.features | features);
                 }
+                // The cached writer was selected by the first value seen on this field (e.g. an Object
+                // or generic field whose fieldClass erases to Object). When a later value has a
+                // different runtime type, reusing that writer throws ClassCastException (wrapped as
+                // "key get error"). Re-resolve by the actual value class, mirroring the typeMatch
+                // check FieldWriterObject.getObjectWriter(jsonWriter, valueClass) does on the write
+                // path. See issue #7714.
+                Class initValueClass = ((FieldWriterObject) fieldWriter).initValueClass;
+                if (valueWriter != null && initValueClass != null && !initValueClass.isInstance(fieldValue)) {
```
wenshao: **[Suggestion]** The `isInstance` predicate diverges from the write path's `typeMatch` check in `FieldWriterObject.getObjectWriter`: `isInstance` accepts any subclass, while the write path requires exact class identity (plus narrow `writeUsing`/`Map`/`List` exceptions). For subclass values, `toJSONObject` reuses the superclass's cached writer (silently dropping subclass-specific fields) while `toJSONString` re-resolves and includes them — the two APIs produce different JSON for the same object. The comment above claims to mirror `typeMatch` but the predicate does not match. — Concrete cost: a bean with an `Object` field first serialized as `Base`, then as `Sub extends Base` — `toJSONString` includes `Sub`'s extra fields, `toJSON` drops them. A maintainer debugging this divergence reads the "mirroring typeMatch" comment and looks elsewhere.

```suggestion
                    Class initValueClass = ((FieldWriterObject) fieldWriter).initValueClass;
                    if (valueWriter != null && initValueClass != null && initValueClass != fieldValue.getClass()) {
                        valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
```

core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
```diff
@@ -0,0 +1,93 @@
...
+        dcn.setKey("K1");
+        dcn.setTenantId("10000");
+        m1.setPayload(dcn);
+        JSON.toJSONString(m1);
+
+        // Step 2: toJSONObject path with a String payload must not reuse that cached writer.
+        // The String value must be preserved as-is instead of throwing "key get error".
+        DefaultMessage m2 = new DefaultMessage();
+        m2.setPayload("{\"key\":\"abc\"}");
+        Object r = JSON.toJSON(m2);
```
wenshao, line 85: **[Suggestion]** Neither test exercises the fix's re-resolution-then-convert branch with a bean-to-bean type switch. Test 1 uses a `String` payload (re-resolved writer is not an `ObjectWriterAdapter`, so the recursive `toJSONObject` call is not taken). Test 2 uses the same type (no re-resolution). A future regression that discards conversion on mismatch (e.g. simplifying to `valueWriter = null`) would pass both tests, yet `JSON.toJSON` would return a raw bean instead of a nested `JSONObject`. — Concrete cost: the branch the fix actually added (re-resolve → `ObjectWriterAdapter.toJSONObject(fieldValue)`) has no regression guard.

Suggested fix: add a third test that primes the cache via `toJSONString` with `DataChangeNotification`, then calls `toJSON` with a different bean type payload, and asserts the result is a nested `JSONObject` carrying the second bean's fields.

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
fudianchn: `isInstance` is strictly looser than the write path's `typeMatch`, so a subclass value would still reuse the parent's cached writer and silently drop the subclass fields, making `toJSON` diverge from `toJSONString`.

Fixed by extracting the write path's predicate into `FieldWriterObject.isTypeMatch(Class valueClass)` (exact class identity plus the same `writeUsing` / `Map` / `List` / primitive-boxing exceptions) and calling it from `ObjectWriterAdapter.toJSONObject`. Both paths now share one strict predicate, so a subclass value triggers re-resolution by its actual class, consistent with `toJSONString`.

I also pointed `getObjectWriter` itself at the new helper so the two cannot drift again.

Sanity check: with the old `isInstance` logic the new subclass test fails with `expected: <online> but was: <null>` (subclass field dropped); with `isTypeMatch` it passes.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
fudianchn: The original tests left that branch uncovered (Test1's String payload never reaches the recursive `toJSONObject`; Test2 never re-resolves). A future regression that simplifies this back to `valueWriter = null` would pass both while `toJSON` returns the raw bean instead of a nested `JSONObject`.

Added `toJSON_beanToBeanSubclassReparseAndRecurses`: it caches the writer for the parent type (`DataChangeNotification`) via `toJSONString`, then calls `toJSON` with a subclass payload (`DataChangeNotificationExt`, which adds a `scope` field). It asserts:
- the payload is a nested `JSONObject` (not the raw bean, `assertNotSame`),
- the subclass field `scope` is preserved,
- `toJSON` and `toJSONString` produce identical output for the same object.

This exercises the re-resolve-then-recurse path and directly encodes the `toJSON`/`toJSONString` consistency guarantee. Verified it fails on the previous `isInstance` logic (`expected: <online> but was: <null>`).

fudianchn:
Pushed an update addressing both suggestions.

**R1-1 (typeMatch divergence):** Replaced the looser `isInstance` check with the write path's exact-class `typeMatch` predicate. Extracted it into `FieldWriterObject.isTypeMatch(Class)` and called it from both `ObjectWriterAdapter.toJSONObject` and `FieldWriterObject.getObjectWriter`, so the two paths share one predicate and cannot drift. A subclass value now re-resolves the writer by its actual class, matching `toJSONString`.

**R1-2 (missing test branch):** Added `toJSON_beanToBeanSubclassReparseAndRecurses` covering the bean->bean-subclass re-resolve + recursive `toJSONObject` branch, asserting the subclass field is preserved, the payload is a nested `JSONObject` (not the raw bean), and `toJSON` agrees with `toJSONString`.

Verification (commit 6c29d27):
- `Issue7714`: 3/3 pass (incl. the new test). The new test fails on the prior `isInstance` logic with `expected: <online> but was: <null>`, confirming it guards the fix.
- Regression: `JSONTest, JSONObjectTest, JSONReader*Test, *ObjectWriter*Test, Issue*` → 2500/2500 pass.
- `checkstyle:check` (fastjson2-checks.xml): 0 violations.

core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
```diff
@@ -0,0 +1,147 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Regression for issue #7714: when a bean has an {@code Object} (or generic) field,
+ * {@link com.alibaba.fastjson2.JSON#toJSON(Object)} reused the field writer cached for the
+ * first serialized value type without checking it still matches the current value, raising
+ * {@code JSONException: key get error} (a wrapped ClassCastException), and — for a subclass
+ * value — silently dropping the subclass fields by reusing the parent writer.
+ */
+public class Issue7714 {
```
wenshao, line 21: **[Suggestion]** Missing `@Tag("regression")` annotation that 270 of 273 sibling `Issue*.java` files in this directory carry. — Concrete cost: when the test suite is filtered by the regression tag (e.g. `mvn test -Dgroups=regression`), this regression test for issue #7714 is silently excluded, so a future regression in the `toJSONObject` type-match fix would not be caught by tagged regression runs.

```suggestion
@Tag("regression")
public class Issue7714 {
```

wenshao (commented):

core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
fudianchn: Done: added `@Tag("regression")` to `Issue7714` (commit ff1a7e9) so it is picked up by `mvn test -Dgroups=regression`, matching the sibling `Issue*.java` files.

Verified: `Issue7714` 3/3 pass, 0 Checkstyle violations.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -685,6 +685,19 @@ public JSONObject toJSONObject(T object, long features) {
...
+                // different runtime type, reusing that writer throws ClassCastException (wrapped as
+                // "key get error"); and a subclass value would be written with the parent writer,
+                // silently dropping the subclass fields. Re-resolve by the actual value class using
+                // the same typeMatch check the write path applies in
+                // FieldWriterObject.getObjectWriter(jsonWriter, valueClass), so toJSON stays
+                // consistent with toJSONString. See issue #7714.
+                FieldWriterObject objectFieldWriter = (FieldWriterObject) fieldWriter;
+                if (objectFieldWriter.initValueClass != null
+                        && !objectFieldWriter.isTypeMatch(fieldValue.getClass())) {
+                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
```
wenshao, line 699: **[Suggestion]** When the re-resolved writer here is not an `ObjectWriterAdapter` (arrays, `Date`, enums, etc.), the raw Java object is put into the result `JSONObject`, while the unprimed path converts the same value via `JSON.toJSON` — so the in-memory graph shape of `JSON.toJSON` depends on whether a bean payload was previously serialized on the same field in this JVM. — Concrete cost: probe-verified — with an `Object payload` field, a fresh JVM stores an `int[]` payload as `JSONArray` and a `Date` payload as `String`; after `toJSONString` primes the field with a bean payload, the identical inputs store a raw `int[]` / `java.util.Date`. `(JSONArray) ((JSONObject) JSON.toJSON(msg)).get("payload")` passes in unit tests but throws `ClassCastException` in a long-lived process (cache-order heisenbug). Pre-PR the primed case threw, so this is not a regression — but it undercuts the consistency the added comment claims.

```suggestion
                if (objectFieldWriter.initValueClass != null
                        && !objectFieldWriter.isTypeMatch(fieldValue.getClass())) {
                    valueWriter = JSONFactory.getObjectWriter(fieldValue.getClass(), this.features | features);
                    if (!(valueWriter instanceof ObjectWriterAdapter)) {
                        fieldValue = JSON.toJSON(fieldValue);
                    }
```

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
```diff
@@ -101,6 +95,26 @@ public ObjectWriter getObjectWriter(JSONWriter jsonWriter, Class valueClass) {
...
+     * Whether a value of {@code valueClass} is compatible with the writer cached for
+     * {@link #initValueClass}. Mirrors the typeMatch check performed on the write path so that
+     * {@link ObjectWriterAdapter#toJSONObject(Object)} keeps {@code toJSON} consistent with
+     * {@code toJSONString}. Notably this requires an exact class match (plus the writeUsing / Map /
+     * List / primitive-boxing exceptions), so a subclass value does NOT match and must be
+     * re-resolved by its actual class to avoid silently dropping the subclass fields.
+     */
+    public boolean isTypeMatch(Class valueClass) {
+        final Class initValueClass = this.initValueClass;
+        boolean typeMatch = initValueClass == valueClass
```
wenshao, line 113: **[Suggestion]** `isTypeMatch` NPEs when `initValueClass` is null — the precondition is undocumented and both current callers guard externally. — Concrete cost: `initValueClass` is null for every `FieldWriterObject` until the first value is serialized through it; a future caller of this `public` method (its javadoc invites reuse) that omits the external null guard hits `NullPointerException` at `initValueClass.isPrimitive()` / `isAssignableFrom(...)`. The write path's guard (`getObjectWriter` returns `getObjectWriterVoid` when `initValueClass == null`) was not carried into the extracted method.

```suggestion
    public boolean isTypeMatch(Class valueClass) {
        final Class initValueClass = this.initValueClass;
        if (initValueClass == null) {
            return false;
        }
        boolean typeMatch = initValueClass == valueClass
```

wenshao (commented):

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java
fudianchn: Fixed in 514f3af: `isTypeMatch` now returns `false` when `initValueClass == null` (matching the write path's `getObjectWriter` guard that returns `getObjectWriterVoid` in that case), with a javadoc note on the precondition. Both current callers already guard externally, so this is defensive and self-documenting for future callers of the public method.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
fudianchn: Fixed in 514f3af: when the re-resolved writer is not an `ObjectWriterAdapter` (arrays, `Date`, enums, ...), the value is now converted via `JSON.toJSON(fieldValue)` so the stored shape matches the unprimed path and `toJSONString`, mirroring the existing `JSON.toJSON` fallback just below. Added `toJSON_nonAdapterValueAfterBeanCacheConvertsLikeUnprimed` (primed bean field, then `int[]` payload -> asserts `JSONArray`), which fails before the conversion.

The broader primed+match case (e.g. an `Object` field first primed with a bean, later a `Date`) shares the same raw-object trait and predates this PR; left out of scope as it changes behavior outside issue #7714.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java
```diff
@@ -685,6 +685,26 @@ public JSONObject toJSONObject(T object, long features) {
...
+                    // When the re-resolved writer is not an ObjectWriterAdapter (arrays, Date, enums,
+                    // etc.), convert the value via JSON.toJSON so it matches the shape the unprimed
+                    // path and toJSONString produce, instead of leaving the raw Java object in the
+                    // JSONObject. See issue #7714.
+                    if (!(valueWriter instanceof ObjectWriterAdapter)) {
+                        fieldValue = JSON.toJSON(fieldValue);
+                    }
```
wenshao, line 706: **[Suggestion]** R3-1: (fix-induced) The new mismatch conversion leg (`fieldValue = JSON.toJSON(fieldValue)`) renders the value through a default write context, discarding both the field's `@JSONField(format …)` and the caller's `JSONWriter` features — so `JSON.toJSON` still diverges from `JSON.toJSONString` on this newly-reachable mismatch path, the consistency the comment above promises. The round-3 fix closed the raw-put defect; the conversion leg it added re-opens the divergence class at this site. Concretely: prime an Object-typed payload field with a bean via `toJSONString`, then send a non-bean payload — with `format="yyyy-MM-dd"` and a `ZonedDateTime` payload, `toJSONString` emits `"2026-09-07"` while `toJSON` stores `"2026-09-07T15:04:05[Asia/Shanghai]"`; with `BrowserCompatible` and a `long[]{5, 9007199254740993}` payload, `toJSONString` writes `"9007199254740993"` as a string while `toJSON` stores raw numbers `[5,9007199254740993]`. Before this fix both inputs threw the wrapped `ClassCastException`, so the silent divergence is new surface of this change. (`java.util.Date` does not diverge — `FieldWriter.getObjectWriter` has no `java.util.Date` branch, so the write path ignores the format there too.)

Witness:
```
[F1:ZonedDateTime] toJSONString.payload=2026-09-07 | toJSON.payload=2026-09-07T15:04:05[Asia/Shanghai] MATCH=false
[F7:BrowserCompatible] toJSON.tree={"payload":[5,9007199254740993]} vs toJSONString={"payload":[5,"9007199254740993"]} MATCH=false
flip checks (format-aware resolution + render via the resolved writer; context carrying this.features | features): MATCH=true, Issue7714 4/4 green
PR hunk reverted: BASE: JSONException: k get error vs PR: {"payload":[5,9007199254740993]}
```

Suggested fix — one coherent change in this block: resolve format-aware first (`FieldWriter.getObjectWriter(fieldType, fieldClass, format, null, valueClass)` when `format != null`, mirroring `getObjectWriterTypeNotMatch`), then render through the resolved writer with a context carrying `this.features | features` (mirror `JSON.toJSON`'s own fallback, JSON.java:4092-4097) instead of the one-arg `JSON.toJSON`; keep the adapter-recurse vs convert decision on the resolved writer, and share that conversion with the unprimed empty-adapter leg below so primed and unprimed outputs stay identical. Fixing only the resolution is verified not to flip the probe — this conversion leg never dereferences `valueWriter`.

The fix must not change default-context output (the four existing tests pin `int[]` → numeric `JSONArray`) and must not touch the write path's own mismatch leg (`FieldWriterObject.getObjectWriterTypeNotMatch`). Please add the pinning tests to this class — a formatted-temporal variant and a `BrowserCompatible` variant asserting `((JSONObject) r).toJSONString()` equals `JSON.toJSONString(m2, …)` — and confirm they go red if either half of the fix is removed.

_— glm-5.3-flash via Qwen Code /review (v0.23.0)_

core/src/test/java/com/alibaba/fastjson2/issues/Issue7714.java
```diff
@@ -0,0 +1,175 @@
...
+        assertEquals(expected, ((JSONObject) r).toJSONString());
+    }
+
+    @Test
+    public void toJSON_nonAdapterValueAfterBeanCacheConvertsLikeUnprimed() {
+        // Prime the payload FieldWriterObject with a bean so initValueClass is set, then send a
+        // value whose writer re-resolves to a non-ObjectWriterAdapter (an int[]). The re-resolved
+        // writer is not an adapter, so the value must still be converted via JSON.toJSON to a
+        // JSONArray — matching the unprimed path and toJSONString — rather than stored as a raw
+        // int[] whose shape depends on cache priming order.
```
wenshao, line 157: **[Suggestion]** The unprimed baseline this conversion claims parity with ("matching the shape the unprimed path and toJSONString produce") is asserted by no test: all four tests in this class prime the cache before asserting, and no test in `core/src/test` pins the unprimed Object-field shapes. A future refactor of the unprimed branch back to storing raw values would keep this suite green while reintroducing the cache-order-dependent trees this fix exists to kill — primed payloads converted, unprimed ones raw.

Witness:
```
[F3:bean] payload={"u":"uu"} class=com.alibaba.fastjson2.JSONObject
[F3:int[]] payload=[7,8] class=com.alibaba.fastjson2.JSONArray   (fresh never-primed fixture classes)
```

Suggested fix: add one test using bean classes no other test uses (the global writer cache's priming persists for the JVM lifetime, so reused classes would make the assertion suite-order-dependent): unprimed bean payload → `JSONObject` with the bean's fields, unprimed `int[]` payload → `JSONArray`. It must go red if the unprimed empty-adapter conversion in `toJSONObject` stops converting — leave the raw value and this test fails while the primed ones stay green; please verify that mutation before merging.

_— glm-5.3-flash via Qwen Code /review (v0.23.0)_

wenshao (commented):

Not explored to full depth (tool budget reached): `"agent 6b"`: `did not run the repo's Maven suite in the shared worktree (builds in place are barred); relied on code-path verification and the merged PR's CI for the priming …`.



Deferred under the convergence posture (round 4, not a blocker) — recorded, not requested in this round:

- `core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java:643 — [review] Unwrapped branch still lacks the typeMatch guard (deferred: age rule)`
- `core/src/main/java/com/alibaba/fastjson2/writer/FieldWriterObject.java:114 — [review] writeUsing branch of isTypeMatch untested from new consumer (deferred: age rule)`

Convergence: round 4 posted 2 inline comment(s), 2 of them reported for the first time. Findings keep coming back to the same files: `core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterAdapter.java` (findings in round 3; 1 more now). A cluster that keeps producing siblings usually means the fixes are treating instances of a shared root cause — triaging that cause before the next round, or splitting an independent cluster into its own pull request, tends to end the loop faster than fixing them one at a time. (Observation only — nothing was withheld from this review because of this observation.)

_— glm-5.3-flash via Qwen Code /review (v0.23.0)_

PR #7742, opened 2026-08-02
fix: preserve generic element type for custom List implementations

### LINKED ISSUE #7684: TypeReference with custom List implementation loses generic type, elements become JSONObject
## 问题描述

`TypeReference<CustomList<Xxx>>`（其中 `CustomList` 是自定义实现 `List<E>` 接口的类）在 fastjson 2.x 下泛型信息丢失，元素被反序列化为 `JSONObject`。遍历时抛出 `ClassCastException: JSONObject cannot be cast to Xxx`。

fastjson 1.x (1.2.83) 无此问题。

## 环境信息

- fastjson 版本: 2.0.62 (bundled fastjson2: 2.0.61)
- JDK: 8+

## 重现步骤

```java
import com.alibaba.fastjson.JSONException;
import com.alibaba.fastjson.JSONObject;
import com.alibaba.fastjson.TypeReference;

import java.util.*;

public class Repro {

    public static class CustomList<E> implements List<E> {
        private List<E> list = new ArrayList<>();
        public int size() { return list.size(); }
        public boolean isEmpty() { return list.isEmpty(); }
        public boolean contains(Object o) { return list.contains(o); }
        public Iterator<E> iterator() { return list.iterator(); }
        public Object[] toArray() { return list.toArray(); }
        public <T> T[] toArray(T[] a) { return list.toArray(a); }
        public boolean add(E e) { return list.add(e); }
        public boolean remove(Object o) { return list.remove(o); }
        public boolean containsAll(Collection<?> c) { return list.containsAll(c); }
        public boolean addAll(Collection<? extends E> c) { return list.addAll(c); }
        public boolean addAll(int index, Collection<? extends E> c) { return list.addAll(index, c); }
        public boolean removeAll(Collection<?> c) { return list.removeAll(c); }
        public boolean retainAll(Collection<?> c) { return list.retainAll(c); }
        public void clear() { list.clear(); }
        public E get(int index) { return list.get(index); }
        public E set(int index, E element) { return list.set(index, element); }
        public void add(int index, E element) { list.add(index, element); }
        public E remove(int index) { return list.remove(index); }
        public int indexOf(Object o) { return list.indexOf(o); }
        public int lastIndexOf(Object o) { return list.lastIndexOf(o); }
        public ListIterator<E> listIterator() { return list.listIterator(); }
        public ListIterator<E> listIterator(int index) { return list.listIterator(index); }
        public List<E> subList(int fromIndex, int toIndex) { return list.subList(fromIndex, toIndex); }
    }

    public static class Setting {
        private String name;
        private String value;
        public String getName() { return name; }
        public void setName(String name) { this.name = name; }
        public String getValue() { return value; }
        public void setValue(String value) { this.value = value; }
    }

    @SuppressWarnings({"unchecked", "rawtypes"})
        // 场景 A：单元素 → 元素类型丢失为 JSONObject，抛 ClassCastException
        System.out.println("=== A: 单元素 ===");
        Map<String, Object> item = new HashMap<>();
        item.put("name", "key1");
        item.put("value", "val1");
        String singleJson = JSON.toJSONString(Collections.singletonList(item));

        CustomList<?> resultA = JSON.parseObject(
                singleJson,
                new TypeReference<CustomList<Setting>>() {}.getType());

        System.out.println("element is JSONObject? " + (resultA.get(0) instanceof JSONObject));
        // true → 类型丢失

        try {
            CustomList<Setting> typed = (CustomList<Setting>) (List) resultA;
            for (Setting s : typed) { System.out.println(s.getName()); }
        } catch (ClassCastException e) {
            System.out.println("ClassCastException: " + e.getMessage());
        }

        // 场景 B：多元素 → 直接抛 JSONException
        System.out.println("\n=== B: 多元素 ===");
        Map<String, Object> item2 = new HashMap<>();
        item2.put("name", "key2");
        item2.put("value", "val2");
        String multiJson = JSON.toJSONString(Arrays.asList(item, item2));

        try {
            JSON.parseObject(multiJson, new TypeReference<CustomList<Setting>>() {}.getType());
        } catch (JSONException e) {
            System.out.println("JSONException: " + e.getMessage());
        }

        // 场景 C：原生 List 接口正常
        System.out.println("\n=== C: List<Setting> 正常 ===");
        List<Setting> resultC = JSON.parseObject(
                multiJson, new TypeReference<List<Setting>>() {}.getType());
        System.out.println("element class: " + resultC.get(0).getClass().getName());
    }
}
```

## 输出

```
=== A: 单元素 ===
element is JSONObject? true
ClassCastException: com.alibaba.fastjson.JSONObject incompatible with Repro$Setting

=== B: 多元素 ===
JSONException: expect {, but [, class Repro$CustomList, offset 1, ...

=== C: List<Setting> 正常 ===
element class: Repro$Setting
```

## 预期行为

应同 `TypeReference<List<Xxx>>` 一样，正确将元素反序列化为 `Xxx` 而非 `JSONObject`。

## 对比

- fastjson 1.x (1.2.83): ✅ 正常
- fastjson 2.x (2.0.62): ❌ 异常
- jackson: ✅ 正常

## 备注

已知 issue #3926 已修复类似的 List 子类问题，但该修复针对的是「作为另一个对象的字段」的场景（如 `MyDTO.myList`），不涵盖 `TypeReference<CustomList<Xxx>>` 直接解析的场景。当前问题仍存在。

fudianchn:

## What
A custom `List` implementation used as a `TypeReference` generic argument now keeps its element type.

## Why
```java
public class CustomList<E> extends AbstractList<E> { ... }
CustomList<Bean> r = JSON.parseObject(json, new TypeReference<CustomList<Bean>>() {});
```
Before, the element generic type was lost. The failure mode differs by entry point:
- compat module (`com.alibaba.fastjson`): `r.get(0)` is `JSONObject`, `ClassCastException` on use.
- core (`com.alibaba.fastjson2`): `JSONException` ("expect {, but [").

After: `r.get(0)` is `Bean`.

Issue #7684 (regression from fastjson 1.x, where this worked). Both entry points reach the same core path (`ObjectReaderProvider.getObjectReaderInternal`), so fixing core fixes both.

## Root cause
In `ObjectReaderProvider.getObjectReader`, the parameterised-type fallback that routes a 1-arg generic type to `ObjectReaderImplList` gated on `ArrayList.class.isAssignableFrom(rawClass)`. A `List` that doesn't extend `ArrayList` (e.g. one extending `AbstractList`, or any plain `implements List`) failed that check and fell through to bean reflection, losing the element generic type. The adjacent 2-arg branch already used the interface `Map.class.isAssignableFrom`, so the List branch was asymmetric.

## How
Use `List.class.isAssignableFrom(rawClass)` instead of `ArrayList.class`, mirroring the Map branch.

Broadening the gate to `List.class` makes a non-`ArrayList` `List` subclass that declares a custom deserializer (`@JSONType(deserializer = ...)` / `deserializeUsing`) match the list fast-path. Before this change such a subclass fell through to `createObjectReader`, which honors `beanInfo.deserializer` (same logic as `ObjectReaderCreator.createObjectReader`). To preserve that, this branch first runs the annotation processors via `getBeanInfo(beanInfo, rawClass)` and, if a deserializer is declared, instantiates and returns it the same way `createObjectReader` does; only when no deserializer is declared does it return `ObjectReaderImplList.of(...)`. Both the deserializer reader and the list reader are cached via `getPreviousObjectReader`, so a reused `TypeReference` returns the same instance.

## Testing
New `Issue7684` (`@Tag("regression")`):
- `customListGenericPreserved`: a `CustomList<Bean>` extending `AbstractList` round-trips with `Bean` elements.
- `customDeserializerOnListSubclassHonored`: a `@JSONType(deserializer = ...)` `AbstractList` subclass used as a `TypeReference` argument invokes the declared reader instead of the list fast-path.

`./mvnw -pl core test`: full core suite green.

Fixes #7684

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -1194,10 +1194,10 @@ public class ObjectReaderProvider
                     }
                 }
                 if (typeArguments.length == 1 && List.class.isAssignableFrom(rawClass)) {
-                    // Honor a custom deserializer declared on the List subclass (e.g.
-                    // @JSONType(deserializer = ...), @JSONType(deserializeUsing = ...) or Jackson's
-                    // @JsonDeserialize(using = ...)) before the list fast-path. Otherwise the generic
-                    // list reader is returned and the user's deserializer directive is silently bypassed.
+                    // Honor a custom deserializer declared on the List subclass — any ObjectReader
+                    // populated into beanInfo.deserializer by getBeanInfo (e.g. @JSONType(deserializer = ...))
+                    // — before the list fast-path. Otherwise the generic list reader is returned and the
+                    // user's deserializer directive is silently bypassed.
                     BeanInfo beanInfo = new BeanInfo(this);
                     getBeanInfo(beanInfo, rawClass);
                     if (beanInfo.deserializer != null && ObjectReader.class.isAssignableFrom(beanInfo.deserializer)) {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -1194,7 +1194,28 @@ public class ObjectReaderProvider
                     }
                 }
                 if (typeArguments.length == 1 && List.class.isAssignableFrom(rawClass)) {
-                    return ObjectReaderImplList.of(objectType, rawClass, 0);
+                    // Honor a custom deserializer declared on the List subclass (e.g.
+                    // @JSONType(deserializer = ...), @JSONType(deserializeUsing = ...) or Jackson's
+                    // @JsonDeserialize(using = ...)) before the list fast-path. Otherwise the generic
+                    // list reader is returned and the user's deserializer directive is silently bypassed.
+                    BeanInfo beanInfo = new BeanInfo(this);
+                    getBeanInfo(beanInfo, rawClass);
+                    if (beanInfo.deserializer != null && ObjectReader.class.isAssignableFrom(beanInfo.deserializer)) {
+                        ObjectReader deserializerReader;
+                        try {
+                            Constructor<?> constructor = beanInfo.deserializer.getDeclaredConstructor();
+                            constructor.setAccessible(true);
+                            deserializerReader = (ObjectReader) constructor.newInstance();
+                        } catch (InstantiationException | IllegalAccessException | NoSuchMethodException
+                                | InvocationTargetException e) {
+                            throw new JSONException("create deserializer error", e);
+                        }
+                        ObjectReader previous = getPreviousObjectReader(fieldBased, objectType, deserializerReader);
+                        return previous != null ? previous : deserializerReader;
+                    }
+                    ObjectReader listReader = ObjectReaderImplList.of(objectType, rawClass, 0);
+                    ObjectReader previous = getPreviousObjectReader(fieldBased, objectType, listReader);
+                    return previous != null ? previous : listReader;
                 }
 
                 if (typeArguments.length == 2 && Map.class.isAssignableFrom(rawClass)) {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
@@ -1193,7 +1193,7 @@ public class ObjectReaderProvider
                         return rawClassReader;
                     }
                 }
-                if (typeArguments.length == 1 && ArrayList.class.isAssignableFrom(rawClass)) {
+                if (typeArguments.length == 1 && List.class.isAssignableFrom(rawClass)) {
                     return ObjectReaderImplList.of(objectType, rawClass, 0);
                 }
 
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7684.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7684.java
@@ -21,7 +21,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  * lost its generic element type — elements were decoded as JSONObject.
  */
 @Tag("regression")
-public class Issue7684Test {
+public class Issue7684 {
     public static class CustomList<E> extends AbstractList<E> {
         private final List<E> list = new ArrayList<>();
 
--- a/core/src/test/java/com/alibaba/fastjson2/issues/Issue7684Test.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues/Issue7684Test.java
@@ -1,10 +1,14 @@
 package com.alibaba.fastjson2.issues;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONReader;
 import com.alibaba.fastjson2.TypeReference;
+import com.alibaba.fastjson2.annotation.JSONType;
+import com.alibaba.fastjson2.reader.ObjectReader;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import java.lang.reflect.Type;
 import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -1193,7 +1193,7 @@ private ObjectReader getObjectReaderInternal(Type objectType, boolean fieldBased
                         return rawClassReader;
                     }
                 }
-                if (typeArguments.length == 1 && ArrayList.class.isAssignableFrom(rawClass)) {
+                if (typeArguments.length == 1 && List.class.isAssignableFrom(rawClass)) {
```
wenshao, line 1196: **[Critical]** This widened branch returns `ObjectReaderImplList.of(...)` before the creator fallback, so a custom deserializer declared with `@JSONType(deserializer = ...)` / `@JSONType(deserializeUsing = ...)` (or Jackson's `@JsonDeserialize(using = ...)`) on a non-`ArrayList` `List` subclass is now silently ignored when the type is used as a parameterized `TypeReference` argument.

Before this change, a custom (non-`ArrayList`) `List` subclass fell through to `creator.createObjectReader(...)`, which populates `beanInfo.deserializer` and returns the user's reader (`ObjectReaderCreator.java:1182`). After the change, `List.class.isAssignableFrom(rawClass)` is true, so `ObjectReaderImplList.of(...)` is returned immediately and the user's explicit deserializer directive is bypassed — with no error or warning. The same bypass already existed for `ArrayList` subclasses; this change newly extends it to every other `List` subclass.

— Failure scenario: `@JSONType(deserializer = MyListReader.class) class MyList<E> extends AbstractList<E>` parsed via `new TypeReference<MyList<Bean>>() {}` → before: `MyListReader` is invoked; after: it is silently bypassed and generic list logic is used (A/B probe confirmed: base `customReaderUsed=true`, PR `customReaderUsed=false`).

Suggested fix: before returning from this branch, honor a custom deserializer annotation the same way `ObjectReaderCreator.createObjectReader` does (build a `BeanInfo`, run the modules' annotation processors via `getBeanInfo(beanInfo, rawClass)`, and if `beanInfo.deserializer` is an `ObjectReader` subclass, instantiate and return it); only fall through to `ObjectReaderImplList.of(...)` when no custom deserializer is declared.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
```diff
@@ -1193,7 +1193,7 @@ private ObjectReader getObjectReaderInternal(Type objectType, boolean fieldBased
...
                     return ObjectReaderImplList.of(objectType, rawClass, 0);
```
wenshao: **[Suggestion]** This branch returns the freshly built reader without caching it. Before this change, a custom non-`ArrayList` `List` subclass fell through to the generic creator path, which caches via `getPreviousObjectReader` / `cache.putIfAbsent`; after the change it hits this uncached branch, so the reader is rebuilt on every parse. A probe confirmed two `getObjectReader` calls with the same `TypeReference` return different instances. (The adjacent `Map` branch shares this trait, so it may be a deliberate house pattern — but the widening of the `List` branch is introduced by this change.)

— Concrete cost: hot-looping `JSON.parseObject(json, REF)` with a reused `static final TypeReference<CustomList<Bean>>` rebuilds `ObjectReaderImplList.of` each parse (parameterized-type resolution, class-name switch, allocation, lazy constructor reflection) instead of returning a cached reader.

```suggestion
                    ObjectReader listReader = ObjectReaderImplList.of(objectType, rawClass, 0);
                    ObjectReader previous = getPreviousObjectReader(fieldBased, objectType, listReader);
                    return previous != null ? previous : listReader;
```

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
fudianchn: Fixed in ede7adb: before returning the list reader I now build a `BeanInfo` through the modules' annotation processors (`getBeanInfo(beanInfo, rawClass)`, mirroring `ObjectReaderCreator.createObjectReader`) and, if `beanInfo.deserializer` is an `ObjectReader` subclass, instantiate and return it; only when no custom deserializer is declared does it fall through to `ObjectReaderImplList.of`. Added `customDeserializerOnListSubclassHonored` (a `@JSONType(deserializer = ...)` `AbstractList` subclass parsed via `TypeReference`); it fails on the old code and passes now.

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderProvider.java
fudianchn: Fixed in the same commit: both the deserializer reader and the list reader now go through `getPreviousObjectReader(fieldBased, objectType, reader)` so a reused `TypeReference` returns the cached instance instead of rebuilding on every parse (mirrors the existing pattern at line 1189).

PR #7764, opened 2026-08-05
perf: detect illegal autoType type name chars while scanning them

wenshao:
### Motivation

Since #7753 (rejecting hash-cache hits), every text autoType occurrence goes through the full string-based authorization. A type name carrying `:` or `!` is rejected by `hasIllegalTypeNameChars` (two `String.indexOf` scans) — but only after the string has been materialized via `getString()` and re-scanned. For malicious or JSON-LD payloads (`@type` is a JSON-LD keyword whose IRI value always carries `:`), all of that work is spent on a name that will never resolve.

### Changes

- `IOUtils.containsTypeNameSpecialChar(byte[]/char[], off, end)`: SWAR detection of `!` (0x21) and `:` (0x3A), 8 bytes per lane / 4 chars per lane, following the existing `indexOfSlashV` / `containsSlashOrQuoteUTF16` idioms.
- `JSONReader.typeNameIllegal` / `isTypeNameIllegal()`: set by `readTypeHashCode()` overrides in `JSONReaderUTF8` (inherited by `JSONReaderASCII`), `JSONReaderUTF16` and `JSONReaderJSONB` (long-form ASCII fast path plus the direct-string branches of `readTypeHashCode0`).
- Early reject at the reader-adjacent consult points only: `ObjectReaderBean.checkAutoType0`, `ObjectReaderAdapter.autoType` (the ASM text path), `JSONReaderJSONB.checkAutoType` and `JSONReader.getObjectReaderAutoType(long, Class, long)`. When the flag is set they skip `getString()` and the provider lookup and produce exactly the unresolvable outcome the provider would return.
- The flag is conservative: escaped names, JSONB symbol references, UTF-16 strings and names at or above the 192-char limit are not flagged and keep the existing downstream behaviour (including the 192-char length error). Consult points also require both the context-level and the provider-level `AutoTypeBeforeHandler` to be null, so custom handlers still receive and may resolve illegal type names, exactly as before.
- All downstream checks (`hasIllegalTypeNameChars` in the provider, handlers and `loadClass`) stay as defense in depth.

### Benchmark (JDK 26, local)

Parsing `{"@type":"...","id":101}` into the expected bean with `SupportAutoType`:

| case | baseline | this change |
|---|---|---|
| legal type name | 87.9 ns/op | 97.6 ns/op (+10 ns) |
| illegal name (`jar:http://evil/x.jar!/Bad`) | 85.3 ns/op | 79.6 ns/op (−5.7 ns, early reject) |

Note: fusing the detection into `readValueHashCode`'s existing `isASCII` pass was tried and measured slower (+16.6 ns on the legal case), so the wrapper approach was kept.

### Tests

New `AutoTypeIllegalTypeNameTest` (14 cases): `IOUtils` boundary coverage (lengths 1–33 × every position × near-miss characters around 0x21/0x3A), per-reader detection, escaped names falling back to downstream validation, the 192-char length error kept, custom handlers still resolving illegal names, JSON-LD map fallback. Full core suite passes (7988 tests), checkstyle clean, safemode-test 4/4.

<details>
<summary>中文</summary>

### 动机

#7753 拒绝 hash-cache 命中后，文本路径每次 autoType 解析都走完整的字符串授权。类型名中的 `:` 或 `!` 会被 `hasIllegalTypeNameChars`（两次 `String.indexOf` 扫描）拒绝，但此前字符串已经通过 `getString()` 物化并重新扫描。对于恶意构造或 JSON-LD 载荷（`@type` 是 JSON-LD 关键字，IRI 值必含 `:`），这些开销全部花在注定无法解析的名字上。

### 改动

- `IOUtils.containsTypeNameSpecialChar(byte[]/char[], off, end)`：SWAR 检测 `!`（0x21）与 `:`（0x3A），8 字节/lane、4 字符/lane，沿用 `indexOfSlashV` / `containsSlashOrQuoteUTF16` 的既有写法。
- `JSONReader.typeNameIllegal` / `isTypeNameIllegal()`：在 `JSONReaderUTF8`（`JSONReaderASCII` 继承）、`JSONReaderUTF16`、`JSONReaderJSONB`（长格式 ASCII 快路径 + `readTypeHashCode0` 直接字符串分支）的 `readTypeHashCode()` 覆盖中设置。
- 仅在 reader 相邻的消费点提前拒绝：`ObjectReaderBean.checkAutoType0`、`ObjectReaderAdapter.autoType`（ASM 文本路径）、`JSONReaderJSONB.checkAutoType`、`JSONReader.getObjectReaderAutoType(long, Class, long)`。标记命中时跳过 `getString()` 与 provider 查找，结果与 provider 返回 null 完全一致。
- 标记是保守的：转义名、JSONB symbol 引用、UTF-16 字符串、≥192 字符的名字不标记，下游行为（包括 192 长度异常）原样保留；消费点同时要求 context 级与 provider 级 `AutoTypeBeforeHandler` 均为 null，自定义 handler 仍能收到并解析非法类型名，与之前完全一致。
- 下游所有检查（provider、handler、`loadClass` 中的 `hasIllegalTypeNameChars`）保留，作为防御纵深。

### 性能（JDK 26，本机）

将 `{"@type":"...","id":101}` 解析为期望 Bean，开启 `SupportAutoType`：

| 场景 | 基线 | 本次改动 |
|---|---|---|
| 合法类型名 | 87.9 ns/op | 97.6 ns/op（+10 ns） |
| 非法名（`jar:http://evil/x.jar!/Bad`） | 85.3 ns/op | 79.6 ns/op（−5.7 ns，提前拒绝） |

注：尝试过把检测融合进 `readValueHashCode` 已有的 `isASCII` 扫描，实测反而更慢（合法场景 +16.6 ns），因此保留包装方案。

### 测试

新增 `AutoTypeIllegalTypeNameTest`（14 个用例）：`IOUtils` 全边界（长度 1–33 × 每个位置 × 0x21/0x3A 邻近字符）、各 reader 检测、转义名走下游兜底、192 长度错误保留、自定义 handler 仍可解析非法名、JSON-LD 回退。core 全量 7988 个测试通过；checkstyle 无违规；safemode-test 4/4。

</details>

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -150,6 +150,7 @@ public abstract class JSONReader
     protected int nameBegin;
 
     protected boolean nameEscape;
+    protected boolean typeNameIllegal;
     protected boolean valueEscape;
     protected boolean wasNull;
     protected boolean boolValue;
@@ -1619,6 +1620,22 @@ public abstract class JSONReader
      */
     public long readTypeHashCode() {
         return readValueHashCode();
+    }
+
+    /**
+     * Returns whether the type name scanned by the last {@link #readTypeHashCode()} call
+     * contains a character that cannot appear in a Java binary class name ({@code ':'} or
+     * {@code '!'}), making it illegal for autoType. Detecting this during the scan lets
+     * callers reject the type name without materializing and re-scanning it.
+     *
+     * <p>The result is conservative: it is only set when the scan saw the raw type name
+     * bytes or chars, so escaped or symbol-referenced names report false and are still
+     * validated by the downstream checks.
+     *
+     * @return true if the last scanned type name contains {@code ':'} or {@code '!'}
+     */
+    public final boolean isTypeNameIllegal() {
+        return typeNameIllegal;
     }
 
     /**
@@ -6798,6 +6815,11 @@ public abstract class JSONReader
      * @return An ObjectReader for the specified type, or null if not found
      */
     public ObjectReader getObjectReaderAutoType(long typeHash, Class expectClass, long features) {
+        if (typeNameIllegal
+                && context.autoTypeBeforeHandler == null
+                && context.provider.getAutoTypeBeforeHandler() == null) {
+            return null;
+        }
         return context.getObjectReaderAutoType(getString(), expectClass, context.features | features);
     }
 
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
@@ -1317,7 +1317,10 @@ final class JSONReaderJSONB
                 autoTypeError();
             }
 
-            autoTypeObjectReader = context.provider.getObjectReader(getString(), expectClass, features2);
+            autoTypeObjectReader = typeNameIllegal
+                    && context.provider.getAutoTypeBeforeHandler() == null
+                    ? null
+                    : context.provider.getObjectReader(getString(), expectClass, features2);
             if (autoTypeObjectReader == null) {
                 if ((features2 & Feature.ErrorOnNotSupportAutoType.mask) == 0) {
                     return null;
@@ -1672,6 +1675,8 @@ final class JSONReaderJSONB
                 this.strBegin = strBegin;
                 this.strlen = typelen;
                 this.offset = offset;
+                this.typeNameIllegal = typelen < 192
+                        && IOUtils.containsTypeNameSpecialChar(bytes, strBegin, strBegin + typelen);
 
                 return hashCode;
             }
@@ -1680,6 +1685,7 @@ final class JSONReaderJSONB
     }
 
     public long readTypeHashCode0() {
+        typeNameIllegal = false;
         final byte[] bytes = this.bytes;
         byte strtype = this.strtype = bytes[offset];
         if (strtype == BC_SYMBOL) {
@@ -1788,6 +1794,17 @@ final class JSONReaderJSONB
             strBegin = offset;
         } else {
             throw readStringError();
+        }
+
+        // scan-time illegal type name detection; exact for single-byte encodings, since ':'
+        // and '!' never appear inside a multibyte UTF-8 sequence. UTF-16 variants, symbol
+        // references and names at or above the 192 limit stay conservative and are
+        // validated downstream
+        if (strlen < 192
+                && (strtype == BC_STR_UTF8
+                || strtype == BC_STR_ASCII
+                || (strtype >= BC_STR_ASCII_FIX_MIN && strtype <= BC_STR_ASCII_FIX_MAX))) {
+            typeNameIllegal = IOUtils.containsTypeNameSpecialChar(bytes, strBegin, strBegin + strlen);
         }
 
         long hashCode;
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -1668,6 +1668,18 @@ final class JSONReaderUTF16
         return readValueHashCode0();
     }
 
+    @Override
+    public long readTypeHashCode() {
+        long hashCode = readValueHashCode();
+        // a name at or above the 192 limit stays conservative and keeps the downstream
+        // length error
+        typeNameIllegal = hashCode != -1
+                && !nameEscape
+                && nameEnd - nameBegin < 192
+                && IOUtils.containsTypeNameSpecialChar(chars, nameBegin, nameEnd);
+        return hashCode;
+    }
+
     public final long readValueHashCode0() {
         final char quote = ch;
         if (quote != '"' && quote != '\'') {
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
@@ -2809,6 +2809,18 @@ class JSONReaderUTF8
         return nameHashCode;
     }
 
+    @Override
+    public long readTypeHashCode() {
+        long hashCode = readValueHashCode();
+        // gate on the byte length; the char length is never larger, so a name at or above
+        // the 192 limit stays conservative and keeps the downstream length error
+        typeNameIllegal = hashCode != -1
+                && !nameEscape
+                && nameEnd - nameBegin < 192
+                && IOUtils.containsTypeNameSpecialChar(bytes, nameBegin, nameEnd);
+        return hashCode;
+    }
+
     private long readValueHashCode0() {
         int ch = this.ch;
         if (ch != '"' && ch != '\'') {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
@@ -248,14 +248,20 @@ public class ObjectReaderAdapter<T>
 
         ObjectReader autoTypeObjectReader = null;
         if (jsonReader.isSupportAutoTypeOrHandler(featuresAll)) {
-            String typeName = jsonReader.getString();
-            autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, featuresAll);
+            if (jsonReader.isTypeNameIllegal()
+                    && context.getContextAutoTypeBeforeHandler() == null
+                    && context.getProvider().getAutoTypeBeforeHandler() == null) {
+                // an illegal type name is unresolvable, the same as the provider returning null
+            } else {
+                String typeName = jsonReader.getString();
+                autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, featuresAll);
+            }
 
             if (autoTypeObjectReader == null) {
                 if (expectClass == objectClass) {
                     autoTypeObjectReader = this;
                 } else {
-                    throw new JSONException(jsonReader.info("autoType not support : " + typeName));
+                    throw new JSONException(jsonReader.info("autoType not support : " + jsonReader.getString()));
                 }
             }
         }
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderBean.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderBean.java
@@ -176,10 +176,10 @@ public abstract class ObjectReaderBean<T>
         JSONReader.Context context = jsonReader.getContext();
         long features3 = jsonReader.features(features | this.features);
         JSONReader.AutoTypeBeforeHandler autoTypeFilter = context.getContextAutoTypeBeforeHandler();
-        String typeName = jsonReader.getString();
 
         ObjectReader autoTypeObjectReader;
         if (autoTypeFilter != null) {
+            String typeName = jsonReader.getString();
             Class<?> filterClass = autoTypeFilter.apply(typeName, expectClass, features);
 
             if (filterClass != null && !expectClass.isAssignableFrom(filterClass)) {
@@ -191,7 +191,11 @@ public abstract class ObjectReaderBean<T>
             }
 
             autoTypeObjectReader = context.getObjectReader(filterClass);
+        } else if (jsonReader.isTypeNameIllegal()
+                && context.getProvider().getAutoTypeBeforeHandler() == null) {
+            throw auotypeError(jsonReader);
         } else {
+            String typeName = jsonReader.getString();
             autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, features3);
 
             if (autoTypeObjectReader == null) {
--- a/core/src/main/java/com/alibaba/fastjson2/util/IOUtils.java
+++ b/core/src/main/java/com/alibaba/fastjson2/util/IOUtils.java
@@ -2451,6 +2451,88 @@ public class IOUtils {
     }
 
     /**
+     * Checks whether a byte array region contains a character that cannot appear in a Java
+     * binary class name but is meaningful in a URL: {@code ':'} (0x3A) or {@code '!'} (0x21).
+     * Used to reject autoType type names while scanning them, see
+     * {@link TypeUtils#hasIllegalTypeNameChars(String)}.
+     *
+     * @param buf the byte array to check
+     * @param off the starting offset in the array
+     * @param end the end offset (exclusive)
+     * @return true if the region contains {@code ':'} or {@code '!'}
+     */
+    public static boolean containsTypeNameSpecialChar(byte[] buf, int off, int end) {
+        int i = off;
+        long address = ARRAY_BYTE_BASE_OFFSET + off;
+        int upperBound = off + ((end - off) & ~7);
+        while (i < upperBound
+                && notContains2(UNSAFE.getLong(buf, address), 0x3A3A3A3A3A3A3A3AL, 0x2121212121212121L)) {
+            i += 8;
+            address += 8;
+        }
+        while (i < end) {
+            byte b = buf[i++];
+            if (b == ':' || b == '!') {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean notContains2(long value, long v0, long v1) {
+        long x0 = value ^ v0;
+        long x1 = value ^ v1;
+        return ((((x0 - 0x0101010101010101L) & ~x0) | ((x1 - 0x0101010101010101L) & ~x1)) & 0x8080808080808080L) == 0;
+    }
+
+    /**
+     * Checks whether a char array region contains a character that cannot appear in a Java
+     * binary class name but is meaningful in a URL: {@code ':'} or {@code '!'}.
+     * Used to reject autoType type names while scanning them, see
+     * {@link TypeUtils#hasIllegalTypeNameChars(String)}.
+     *
+     * @param buf the char array to check
+     * @param off the starting offset in the array
+     * @param end the end offset (exclusive)
+     * @return true if the region contains {@code ':'} or {@code '!'}
+     */
+    public static boolean containsTypeNameSpecialChar(char[] buf, int off, int end) {
+        int i = off;
+        long address = ARRAY_CHAR_BASE_OFFSET + ((long) off << 1);
+        int upperBound = off + ((end - off) & ~3);
+        while (i < upperBound
+                && notContains2UTF16(UNSAFE.getLong(buf, address), 0x003A_003A_003A_003AL, 0x0021_0021_0021_0021L)) {
+            i += 4;
+            address += 8;
+        }
+        while (i < end) {
+            char c = buf[i++];
+            if (c == ':' || c == '!') {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean notContains2UTF16(long value, long v0, long v1) {
+        /*
+          for (int i = 0; i < 4; ++i) {
+            short c = (short) value;
+            if (c == (short) v0 || c == (short) v1) {
+                return false;
+            }
+            value >>>= 16;
+          }
+          return true;
+         */
+        long x0 = value ^ v0;
+        long x1 = value ^ v1;
+        x0 = (x0 - 0x0001_0001_0001_0001L) & ~x0;
+        x1 = (x1 - 0x0001_0001_0001_0001L) & ~x1;
+        return ((x0 | x1) & 0x8000_8000_8000_8000L) == 0;
+    }
+
+    /**
      * Checks if a byte array region matches a prefix string.
      * This method compares a segment of a byte array with a string prefix to see
      * if they match exactly.
--- a/core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeIllegalTypeNameTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeIllegalTypeNameTest.java
@@ -1,0 +1,270 @@
+package com.alibaba.fastjson2.autoType;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONB;
+import com.alibaba.fastjson2.JSONException;
+import com.alibaba.fastjson2.JSONFactory;
+import com.alibaba.fastjson2.JSONReader;
+import com.alibaba.fastjson2.JSONWriter;
+import com.alibaba.fastjson2.SymbolTable;
+import com.alibaba.fastjson2.util.IOUtils;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for the scan-time detection of illegal autoType type name characters
+ * ({@code ':'} and {@code '!'}), see {@link JSONReader#isTypeNameIllegal()}.
+ */
+@Tag("autotype")
+public class AutoTypeIllegalTypeNameTest {
+    static final String BEAN_NAME = "com.alibaba.fastjson2.autoType.AutoTypeIllegalTypeNameTest$Bean";
+
+    // =========================================================================
+    // IOUtils.containsTypeNameSpecialChar boundary coverage
+    // =========================================================================
+
+    @Test
+    public void testContainsTypeNameSpecialCharBytes() {
+        for (int len = 1; len <= 33; len++) {
+            byte[] buf = new byte[len];
+            for (int pos = 0; pos < len; pos++) {
+                Arrays.fill(buf, (byte) 'a');
+
+                buf[pos] = ':';
+                assertTrue(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos);
+
+                buf[pos] = '!';
+                assertTrue(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos);
+
+                // near-miss characters around ':' (0x3A) and '!' (0x21)
+                for (byte b : new byte[] {'9', ';', ' ', '"', '@', '/', (byte) 0x80, (byte) 0xFF}) {
+                    buf[pos] = b;
+                    assertFalse(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos + ", b=" + b);
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testContainsTypeNameSpecialCharChars() {
+        for (int len = 1; len <= 17; len++) {
+            char[] buf = new char[len];
+            for (int pos = 0; pos < len; pos++) {
+                Arrays.fill(buf, 'a');
+
+                buf[pos] = ':';
+                assertTrue(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos);
+
+                buf[pos] = '!';
+                assertTrue(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos);
+
+                for (char c : new char[] {'9', ';', ' ', '"', '@', '/', 0x3A00, 0x2100, 0xFF3A, 0xFF01}) {
+                    buf[pos] = c;
+                    assertFalse(IOUtils.containsTypeNameSpecialChar(buf, 0, len), "len=" + len + ", pos=" + pos + ", c=" + (int) c);
+                }
+            }
+        }
+    }
+
+    // =========================================================================
+    // readTypeHashCode scan-time detection, one test per reader implementation
+    // =========================================================================
+
+    @Test
+    public void testDetectUtf8() {
+        assertFalse(readTypeNameIllegalUtf8("com.example.Bean"));
+        assertTrue(readTypeNameIllegalUtf8("com.example.Bean:bad"));
+        assertTrue(readTypeNameIllegalUtf8("com.example.Bean!bad"));
+        assertTrue(readTypeNameIllegalUtf8(":com.example.Bean"));
+        assertTrue(readTypeNameIllegalUtf8("com.example.Bean:"));
+        assertTrue(readTypeNameIllegalUtf8("a:b"));
+        // multibyte characters never alias the raw bytes of ':' or '!'
+        assertFalse(readTypeNameIllegalUtf8("com.example.Bean中"));
+        assertTrue(readTypeNameIllegalUtf8("com.example.Bean中:bad"));
+    }
+
+    @Test
+    public void testDetectAscii() {
+        assertFalse(readTypeNameIllegalOfString("com.example.Bean"));
+        assertTrue(readTypeNameIllegalOfString("com.example.Bean:bad"));
+        assertTrue(readTypeNameIllegalOfString("com.example.Bean!bad"));
+        assertTrue(readTypeNameIllegalOfString("a!b"));
+    }
+
+    @Test
+    public void testDetectUtf16() {
+        assertFalse(readTypeNameIllegalUtf16("com.example.Bean"));
+        assertTrue(readTypeNameIllegalUtf16("com.example.Bean:bad"));
+        assertTrue(readTypeNameIllegalUtf16("com.example.Bean!bad"));
+        assertTrue(readTypeNameIllegalUtf16(":com.example.Bean"));
+        assertTrue(readTypeNameIllegalUtf16("com.example.Bean!"));
+    }
+
+    private static boolean readTypeNameIllegalUtf8(String typeName) {
+        JSONReader reader = JSONReader.of(quote(typeName).getBytes(StandardCharsets.UTF_8));
+        reader.readTypeHashCode();
+        return reader.isTypeNameIllegal();
+    }
+
+    private static boolean readTypeNameIllegalOfString(String typeName) {
+        JSONReader reader = JSONReader.of(quote(typeName));
+        reader.readTypeHashCode();
+        return reader.isTypeNameIllegal();
+    }
+
+    private static boolean readTypeNameIllegalUtf16(String typeName) {
+        JSONReader reader = JSONReader.of(quote(typeName).toCharArray());
+        reader.readTypeHashCode();
+        return reader.isTypeNameIllegal();
+    }
+
+    private static String quote(String typeName) {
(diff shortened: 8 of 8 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -1790,6 +1796,17 @@ public long readTypeHashCode0() {
...
+        // scan-time illegal type name detection; exact for single-byte encodings, since ':'
+        // and '!' never appear inside a multibyte UTF-8 sequence. UTF-16 variants, symbol
+        // references and names at or above the 192 limit stay conservative and are
+        // validated downstream
+        if (strlen < 192
+                && (strtype == BC_STR_UTF8
+                || strtype == BC_STR_ASCII
+                || (strtype >= BC_STR_ASCII_FIX_MIN && strtype <= BC_STR_ASCII_FIX_MAX))) {
+            typeNameIllegal = IOUtils.containsTypeNameSpecialChar(bytes, strBegin, strBegin + strlen);
+        }
```
wenshao, line 1808: **[Critical]** Out-of-bounds heap read via `Unsafe` on malformed/truncated JSONB. In `readTypeHashCode0()` this scan runs BEFORE any bounds-checked access to the declared string bytes: `strBegin + strlen` is attacker-controlled and may exceed `bytes.length`, and the SWAR loop's `UNSAFE.getLong` performs no bounds check. The gate admits `strlen` up to 191, so the maximum unchecked read is 184 bytes past the array end. Pre-PR, the identical input deterministically threw `ArrayIndexOutOfBoundsException` in the bounds-checked hash loops below; the `readTypeHashCode()` fast path is unaffected (its checked FNV loop runs first), as are the text readers (scan region always quote-bounded). — Failure scenario: a crafted 2-byte JSONB document `{BC_TYPED_ANY, (byte)(BC_STR_ASCII_FIX_MIN + 47)}` reaches this code with `strlen=47, strBegin=2` on a 2-byte array — the scan sweeps words `[2, 42)`, 40 bytes entirely outside the array. No `SupportAutoType` needed: `checkAutoType` calls `readTypeHashCode()` unconditionally on `BC_TYPED_ANY`, so plain `JSONB.parseObject(malicious, Bean.class)` reaches it. Usually a silent garbage read, but a JVM crash (SIGSEGV) if the overrun crosses into an unmapped page — attacker-crafted JSONB becomes probabilistic process-level DoS on exactly the untrusted-input path this PR hardens. — Witness (A/B probe, same payload on base and PR trees, deterministic 5/5): base P2 throws `Index 2 out of bounds for length 2`, PR P2 throws `Index 42 out of bounds for length 2`; base P4 `Index 4 … length 4`, PR P4 `Index 188 … length 4`. The checked code can only first throw at index 2/4 — index 42/188 proves the SWAR loop swept words far outside the array first; patching the bounds gate flips all outcomes to the guarded path. — Suggested fix: add a region guard (preserves base behavior exactly — truncated input then throws in the checked hash loop immediately after, as before), or move the scan after the validating hash loops like the fast path does:

```suggestion
        if (strlen < 192
                && strBegin + strlen <= bytes.length
                && (strtype == BC_STR_UTF8
                || strtype == BC_STR_ASCII
                || (strtype >= BC_STR_ASCII_FIX_MIN && strtype <= BC_STR_ASCII_FIX_MAX))) {
            typeNameIllegal = IOUtils.containsTypeNameSpecialChar(bytes, strBegin, strBegin + strlen);
        }
```

**[Critical]** 畸形/截断 JSONB 会触发 `Unsafe` 堆越界读。在 `readTypeHashCode0()` 中，该扫描先于任何对声明字符串字节的带边界检查访问执行：`strBegin + strlen` 由攻击者控制，可能超出 `bytes.length`，而 SWAR 循环的 `UNSAFE.getLong` 不做边界检查。门控允许 `strlen` 最大 191，因此最大越界读可达数组末尾之后 184 字节。本 PR 之前，相同输入会在下方带边界检查的 hash 循环中确定性地抛出 `ArrayIndexOutOfBoundsException`；`readTypeHashCode()` 快路径不受影响（其带检查的 FNV 循环先执行），文本 reader 也不受影响（扫描区域始终由引号界定）。

失败场景：构造的 2 字节 JSONB 文档 `{BC_TYPED_ANY, (byte)(BC_STR_ASCII_FIX_MIN + 47)}` 在 2 字节数组上以 `strlen=47, strBegin=2` 到达此处——扫描读取 `[2, 42)` 的字，其中 40 字节完全在数组之外。无需开启 `SupportAutoType`：`checkAutoType` 对 `BC_TYPED_ANY` 无条件调用 `readTypeHashCode()`，因此普通的 `JSONB.parseObject(malicious, Bean.class)` 即可到达。通常只是静默读取垃圾数据，但越界读越过未映射页时会导致 JVM 崩溃（SIGSEGV）——攻击者构造的 JSONB 恰好在本 PR 加强的不可信输入路径上形成概率性进程级 DoS。

证据（A/B 探针，base 与 PR 两棵树相同载荷，5/5 确定性复现）：base P2 抛 `Index 2 out of bounds for length 2`，PR P2 抛 `Index 42 out of bounds for length 2`；base P4 `Index 4`，PR P4 `Index 188`（长度 4 的数组）。带检查代码最早只能在索引 2/4 抛出——索引 42/188 证明 SWAR 循环先扫过了远超数组边界的字；为边界门控打补丁后所有结果翻转为受保护路径。

修复建议：增加区域守卫（精确保留 base 行为——截断输入随后会在紧接其后的带检查 hash 循环中抛异常，与之前一致），或像快路径那样把扫描移到校验 hash 循环之后。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReaderJSONB.java
```diff
@@ -1672,6 +1675,8 @@ public long readTypeHashCode() {
                 this.strBegin = strBegin;
                 this.strlen = typelen;
                 this.offset = offset;
+                this.typeNameIllegal = typelen < 192
+                        && IOUtils.containsTypeNameSpecialChar(bytes, strBegin, strBegin + typelen);
```
wenshao, line 1679: **[Suggestion]** The conservative `< 192` gates this PR adds have no discriminating test — this JSONB fast-path gate, the `readTypeHashCode0` gate below, and the UTF16 twin. The only reader-level ≥192 test, `testLongIllegalNameKeepsLengthError`, uses String input — on JDK 9+ that routes to `JSONReaderASCII` (inherits the UTF8 gate), so the UTF16 gate is pinned only as a side effect of JDK-8 String routing, and the JSONB gates not at all (`testJsonbCheckAutoTypeShortIllegalName` passes whether the gate flags or not, because downstream returns the same null). — Concrete cost: removing the `strlen`/`typelen` gate makes a ≥192-byte JSONB type name carrying ':' silently fall back to the expected class instead of throwing the downstream length error `autoType is not support. <name>` — and no test changes state. — Witness (probe): 202-byte `BC_STR_UTF8` type name (198×'a' + '中' + ':'): CLEAN `isTypeNameIllegal=false` + `checkAutoType` threw `autoType is not support. aaa…`; MUTANT `isTypeNameIllegal=true` + `checkAutoType` returned null (silent fallback); all PR JSONB test assertions unchanged under the mutant. — Suggested fix: mirror `testLongIllegalNameKeepsLengthError` for JSONB — a typed JSONB object with a ≥192-byte name carrying ':' asserting the `JSONException`, plus flag-level `assertFalse(isTypeNameIllegal())` for JSONB names ≥192 and for multibyte UTF-8 JSONB names.

**[Suggestion]** 本 PR 新增的保守 `< 192` 门控没有判别性测试——包括此处的 JSONB 快路径门控、下方 `readTypeHashCode0` 的门控以及 UTF16 孪生门控。唯一的 reader 级 ≥192 测试 `testLongIllegalNameKeepsLengthError` 使用 String 输入——JDK 9+ 上 String 路由到 `JSONReaderASCII`（继承 UTF8 门控），因此 UTF16 门控仅作为 JDK 8 String 路由的副作用被固定，JSONB 门控则完全未被固定（`testJsonbCheckAutoTypeShortIllegalName` 无论门控是否标记都通过，因为下游返回同样的 null）。

具体代价：移除 `strlen`/`typelen` 门控后，≥192 字节含 ':' 的 JSONB 类型名会静默回退到期望类，而不再抛出下游长度错误 `autoType is not support. <name>`——且没有任何测试状态改变。

证据（探针）：202 字节 `BC_STR_UTF8` 类型名（198×'a' + '中' + ':'）：CLEAN `isTypeNameIllegal=false` 且 `checkAutoType` 抛 `autoType is not support. aaa…`；MUTANT `isTypeNameIllegal=true` 且 `checkAutoType` 返回 null（静默回退）；PR 的全部 JSONB 测试断言在突变体下不变。

修复建议：为 JSONB 复制 `testLongIllegalNameKeepsLengthError`——≥192 字节含 ':' 的带类型 JSONB 对象断言 `JSONException`，并对 ≥192 与多字节 UTF-8 的 JSONB 名做 flag 级 `assertFalse(isTypeNameIllegal())`。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
```diff
@@ -1668,6 +1668,18 @@ public final long readValueHashCode() {
...
 
+    @Override
+    public long readTypeHashCode() {
+        long hashCode = readValueHashCode();
+        // a name at or above the 192 limit stays conservative and keeps the downstream
+        // length error
+        typeNameIllegal = hashCode != -1
+                && !nameEscape
+                && nameEnd - nameBegin < 192
+                && IOUtils.containsTypeNameSpecialChar(chars, nameBegin, nameEnd);
```
wenshao, line 1679: **[Suggestion]** This UTF16 `< 192` gate has no direct discriminating test; it is pinned on CI only as a side effect of JDK-8 String routing. `JSONReader.of(String)` routes LATIN1 strings to `JSONReaderASCII` on JDK 9+ (which inherits the UTF8 gate) — the UTF16 reader sees char[]/Reader/non-LATIN1 input on all JDKs, but the only reader-level ≥192 test (`testLongIllegalNameKeepsLengthError`) uses String input, and all `testDetectUtf16` names are short. — Concrete cost: deleting `nameEnd - nameBegin < 192 &&` makes char[]/Reader input with a ≥192-char illegal name silently fall back to the expected class (or the generic `auotypeError`) instead of throwing the downstream length error — an encoding-dependent policy divergence between char[] and byte[]/LATIN1-String inputs — and the mutant survives on JDK 11/17/21/25, caught only on the JDK 8 leg. — Witness (4-arm probe, gate-intact vs gate-removed × JDK 11 vs JDK 8): GATE INTACT: String/char[]/Reader all THROW `autoType is not support`; MUTANT JDK 11: String THROW (ASCII reader unaffected), char[] FALLBACK (bean.id=0), Reader FALLBACK; MUTANT JDK 8: all FALLBACK. — Suggested fix: add char[] (and ideally Reader) variants to `testLongIllegalNameKeepsLengthError`, asserting the same `autoType is not support` message, so every reader implementation's 192 gate is pinned on every JDK leg.

**[Suggestion]** 该 UTF16 `< 192` 门控没有直接的判别性测试；它在 CI 上仅作为 JDK 8 String 路由的副作用被固定。JDK 9+ 上 `JSONReader.of(String)` 把 LATIN1 字符串路由到 `JSONReaderASCII`（继承 UTF8 门控）——UTF16 reader 在所有 JDK 上接收 char[]/Reader/非 LATIN1 输入，但唯一的 reader 级 ≥192 测试（`testLongIllegalNameKeepsLengthError`）使用 String 输入，且 `testDetectUtf16` 的名字都很短。

具体代价：删除 `nameEnd - nameBegin < 192 &&` 后，char[]/Reader 输入的 ≥192 字符非法类型名会静默回退到期望类（或泛化的 `auotypeError`），而不再抛出下游长度错误——char[] 与 byte[]/LATIN1-String 输入之间出现依赖编码的策略分歧——该突变体在 JDK 11/17/21/25 上存活，仅在 JDK 8 分支被捕获。

证据（四臂探针，门控完整 vs 移除 × JDK 11 vs JDK 8）：门控完整：String/char[]/Reader 均抛出 `autoType is not support`；突变体 JDK 11：String 抛出（ASCII reader 不受影响）、char[] 回退（bean.id=0）、Reader 回退；突变体 JDK 8：全部回退。

修复建议：为 `testLongIllegalNameKeepsLengthError` 增加 char[]（最好再加 Reader）变体，断言相同的 `autoType is not support` 消息，使每个 reader 实现的 192 门控在每个 JDK 分支上都被固定。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderBean.java
```diff
@@ -191,7 +191,11 @@ protected final ObjectReader checkAutoType0(JSONReader jsonReader, long features
             }
 
             autoTypeObjectReader = context.getObjectReader(filterClass);
+        } else if (jsonReader.isTypeNameIllegal()
+                && context.getProvider().getAutoTypeBeforeHandler() == null) {
+            throw auotypeError(jsonReader);
```
wenshao, line 196: **[Suggestion]** This early-reject branch ships with no gating test, and the provider-handler guard is unpinned (the ObjectReaderAdapter skip branch is in the same state — separate comment there). Reverting this hunk to the merge base leaves the new 14-test class AND the entire autoType package (268 tests) green — measured by revert probe — because the fast path and the old path end in the same observable outcome (`auotypeError` → `JSONException("autoType not support")`; the base path just reaches it after `getString()` materialization + provider consultation). No test installs a provider-level `AutoTypeBeforeHandler` with an illegal name, so the `provider.getAutoTypeBeforeHandler() == null` guard here is unpinned at all four consult sites. — Failure scenario: a future refactor deleting this branch turns no test red, silently restoring the slow path this PR exists to remove; a future edit dropping the provider null-check silently rejects type names a registered handler used to resolve — the PR's stated "custom AutoTypeBeforeHandler behavior intact" guarantee regresses invisibly. — Witness (measured): this file at merge base — new tests 14/14 green AND autoType package 268/268 green. — Suggested fix: a whitebox test that observes the branch was taken — e.g. a recording context/provider wrapper counting `getObjectReaderAutoType` consultations, asserting zero consultations when the flag is set and no handler is configured, then asserting the expected `JSONException` — plus a provider-level handler variant (property or `setAutoTypeBeforeHandler`) with an illegal name whose resolution is observably distinct from the fallback.

**[Suggestion]** 该提前拒绝分支没有任何门控测试，provider handler 守卫也未被固定（ObjectReaderAdapter 的跳过分支同样——见另一条评论）。将此 hunk 回退到 merge base，新增 14 个测试与整个 autoType 包（268 个测试）仍然全绿——回退探针实测——因为快路径与旧路径的可见结果相同（`auotypeError` → `JSONException("autoType not support")`；base 路径只是经过 `getString()` 物化 + provider 咨询后才到达同一结果）。没有测试用非法名安装 provider 级 `AutoTypeBeforeHandler`，因此此处及全部四个消费点的 `provider.getAutoTypeBeforeHandler() == null` 守卫均未被固定。

失败场景：未来重构删除该分支不会让任何测试变红，静默恢复本 PR 要消除的慢路径；未来删除 provider 空检查会静默拒绝注册 handler 原本能解析的类型名——PR 声明的"自定义 AutoTypeBeforeHandler 行为不变"保证会无声回归。

证据（实测）：本文件回退到 merge base——新测试 14/14 全绿且 autoType 包 268/268 全绿。

修复建议：白盒测试观察分支确实被走到——例如记录型 context/provider 包装器统计 `getObjectReaderAutoType` 咨询次数，断言 flag 置位且无 handler 时次数为零，再断言预期的 `JSONException`——外加 provider 级 handler 变体（属性或 `setAutoTypeBeforeHandler`），使用非法名且其解析结果与回退可区分。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
```diff
@@ -248,14 +248,20 @@ public Object autoType(JSONReader jsonReader, Class expectClass, long features)
 
         ObjectReader autoTypeObjectReader = null;
         if (jsonReader.isSupportAutoTypeOrHandler(featuresAll)) {
-            String typeName = jsonReader.getString();
-            autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, featuresAll);
+            if (jsonReader.isTypeNameIllegal()
+                    && context.getContextAutoTypeBeforeHandler() == null
+                    && context.getProvider().getAutoTypeBeforeHandler() == null) {
```
wenshao, line 253: **[Suggestion]** This early-skip branch ships with no gating test, and the existing handler test cannot detect its bypass (the ObjectReaderBean branch is in the same state — separate comment there). Reverting this hunk to the merge base leaves all 268 autoType tests green — measured by revert probe — and `testCustomHandlerStillReceivesIllegalName` cannot discriminate "handler consulted" from "handler bypassed": the handler resolves `"custom:Bean"` to exactly `expectClass`, so deleting the handler null-checks from this skip condition still passes (`autoTypeObjectReader` stays null → `expectClass == objectClass` → `this` → `bean.id` still 101). — Failure scenario: a future refactor removing the branch silently restores `getString()` materialization + provider lookup on every illegal name (the regression this PR targets); a future edit dropping a handler null-check silently bypasses a registered handler — nothing in the suite fails either way. — Witness (measured): this file at merge base — autoType package 268/268 green. — Suggested fix: strengthen the oracle — record the type name the handler received (e.g. an `AtomicReference` asserted after the parse) and/or have the handler resolve to a subclass and assert on the resulting instance's class; and assert zero `getObjectReaderAutoType` consultations via a recording context when the flag is set and no handler is configured.

**[Suggestion]** 该提前跳过分支没有任何门控测试，且现有 handler 测试无法检测其被绕过（ObjectReaderBean 分支同样——见另一条评论）。将此 hunk 回退到 merge base，全部 268 个 autoType 测试仍然全绿——回退探针实测——并且 `testCustomHandlerStillReceivesIllegalName` 无法区分"handler 被咨询"与"handler 被绕过"：handler 把 `"custom:Bean"` 解析为恰好等于 `expectClass` 的类，因此删除跳过条件中的 handler 空检查测试仍然通过（`autoTypeObjectReader` 保持 null → `expectClass == objectClass` → `this` → `bean.id` 仍为 101）。

失败场景：未来重构移除该分支会静默恢复每个非法名的 `getString()` 物化 + provider 查找（正是本 PR 针对的回归）；未来删除某个 handler 空检查会静默绕过注册 handler——两种情况都没有任何测试失败。

证据（实测）：本文件回退到 merge base——autoType 包 268/268 全绿。

修复建议：加强测试判据——记录 handler 收到的类型名（例如解析后断言的 `AtomicReference`），和/或让 handler 解析到子类并断言结果实例的类型；并用记录型 context 断言 flag 置位且无 handler 时 `getObjectReaderAutoType` 咨询次数为零。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -6798,6 +6815,11 @@ public BigInteger apply(Integer integer, int[] mag) {
      * @return An ObjectReader for the specified type, or null if not found
      */
     public ObjectReader getObjectReaderAutoType(long typeHash, Class expectClass, long features) {
+        if (typeNameIllegal
+                && context.autoTypeBeforeHandler == null
+                && context.provider.getAutoTypeBeforeHandler() == null) {
+            return null;
+        }
```
wenshao, line 6822: **[Suggestion]** This new early-reject sits in a public method with zero in-tree callers and zero test coverage, and it gains a hidden freshness precondition. Repo-wide grep across all modules finds no caller of this `getObjectReaderAutoType(long, Class, long)` overload (the ASM creator emits no such call), and the new test suite never exercises it — so the branch ships unexercised and a future regression in it would be invisible. Pre-PR the method always materialized and validated the current string; now it additionally consults `typeNameIllegal`, which reflects the LAST `readTypeHashCode()` call and is never reset elsewhere. — Failure scenario: an external caller reusing a reader whose most recent `readTypeHashCode()` saw an illegal name (flag left true, parse continued via fallback) and later invoking this method without an intervening `readTypeHashCode()` gets null where it previously resolved — a spurious "autoType not support". In-tree the readTypeHashCode()→consult pairing always holds; the risk is external-API-only. — Suggested fix: document the freshness precondition in the javadoc (mirroring `isTypeNameIllegal()`'s "last readTypeHashCode()" contract), or keep this method validating unconditionally and leave the fast-reject to the internal sites that control freshness; either way, add a direct unit test (illegal name → null; with handler → consulted; legal name → unchanged).

**[Suggestion]** 该提前拒绝位于一个树内零调用者、零测试覆盖的 public 方法中，并引入了隐式的新鲜度前置条件。全仓（所有模块）grep 找不到 `getObjectReaderAutoType(long, Class, long)` 重载的任何调用者（ASM 生成器也不生成该调用），新测试套件从未覆盖它——该分支上线即无人执行，未来的回归也不可见。本 PR 之前该方法总是物化并校验当前字符串；现在它还额外读取 `typeNameIllegal`，而该 flag 反映的是上一次 `readTypeHashCode()` 调用，且从不在其他位置重置。

失败场景：外部调用者复用一个最近一次 `readTypeHashCode()` 见过非法名（flag 置 true，解析经回退继续）的 reader，之后在没有再次调用 `readTypeHashCode()` 的情况下调用本方法，会得到 null 而之前能解析——虚假的 "autoType not support"。树内 readTypeHashCode()→消费配对始终成立；风险仅存在于外部 API。

修复建议：在 javadoc 中记录该新鲜度前置条件（与 `isTypeNameIllegal()` 的"上一次 readTypeHashCode()"契约一致），或让本方法保持无条件校验、把快拒绝留给控制新鲜度的内部消费点；无论哪种，补一个直接单测（非法名 → null；有 handler → 被咨询；合法名 → 行为不变）。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
```diff
@@ -2809,6 +2809,18 @@ public long readValueHashCode() {
...
 
+    @Override
+    public long readTypeHashCode() {
+        long hashCode = readValueHashCode();
+        // gate on the byte length; the char length is never larger, so a name at or above
+        // the 192 limit stays conservative and keeps the downstream length error
+        typeNameIllegal = hashCode != -1
+                && !nameEscape
+                && nameEnd - nameBegin < 192
+                && IOUtils.containsTypeNameSpecialChar(bytes, nameBegin, nameEnd);
```
wenshao, line 2820: **[Suggestion]** The scan-time policy is duplicated across 4 copies and coupled to the downstream rules by bare literals. This flag expression appears verbatim in `JSONReaderUTF16.readTypeHashCode()` and twice more in `JSONReaderJSONB` (`readTypeHashCode()` and `readTypeHashCode0()`); the bare literal `192` appears four new times to mirror the unnamed downstream limit (`typeNameLength >= 192` in `ObjectReaderProvider.checkAutoType`, `className.length() >= 192` in `TypeUtils.loadClass` — no named constant exists repo-wide); and the SWAR char set (0x3A/0x21) independently duplicates `TypeUtils.hasIllegalTypeNameChars`. — Failure scenario: if the downstream limit is ever lowered to N < 192, the four scan gates keep using 192 and an illegal name with N ≤ len < 192 is silently swallowed by the scan-time early-reject (fallback/null) instead of throwing the length error — no compiler or existing test catches it (`testLongIllegalNameKeepsLengthError` hardcodes a 199-char name, safely conservative). Conversely, if `hasIllegalTypeNameChars` ever narrows (drops '!'), the scan still flags those names and all four consult points reject type names the provider would now resolve. A policy edit missed at one of the 4 copies makes `isTypeNameIllegal()` encoding-dependent. — Suggested fix: give the invariant one home — a named constant (e.g. `TypeUtils.MAX_AUTO_TYPE_NAME_LENGTH = 192`) referenced by `checkAutoType`, `loadClass` and the four scan gates, plus a note on `hasIllegalTypeNameChars` that `IOUtils.containsTypeNameSpecialChar` must reject a superset of its predicate (or a test pinning the two definitions together); optionally implement the flag policy once in base `JSONReader` with a one-line per-reader region hook.

**[Suggestion]** 扫描期策略存在 4 份拷贝，且通过裸字面量与下游规则耦合。该 flag 表达式在 `JSONReaderUTF16.readTypeHashCode()` 中逐字重复，在 `JSONReaderJSONB` 中还有两处（`readTypeHashCode()` 与 `readTypeHashCode0()`）；裸字面量 `192` 新增 4 处，用于镜像未命名的下游限制（`ObjectReaderProvider.checkAutoType` 的 `typeNameLength >= 192`、`TypeUtils.loadClass` 的 `className.length() >= 192`——全仓不存在命名常量）；SWAR 字符集（0x3A/0x21）也独立重复了 `TypeUtils.hasIllegalTypeNameChars`。

失败场景：若下游限制将来降到 N < 192，四个扫描门控仍用 192，长度在 N ≤ len < 192 的非法名会被扫描期提前拒绝（回退/null）静默吞掉，而不再抛长度错误——编译器和现有测试都抓不到（`testLongIllegalNameKeepsLengthError` 硬编码 199 字符名，安全地落在保守侧）。反之若 `hasIllegalTypeNameChars` 收窄（去掉 '!'），扫描仍会标记这些名字，四个消费点会拒绝 provider 本将解析的类型名。4 份拷贝中任何一份漏改都会使 `isTypeNameIllegal()` 依赖编码。

修复建议：给该不变量一个家——命名常量（如 `TypeUtils.MAX_AUTO_TYPE_NAME_LENGTH = 192`）供 `checkAutoType`、`loadClass` 与四个扫描门控引用，并在 `hasIllegalTypeNameChars` 上注明 `IOUtils.containsTypeNameSpecialChar` 必须拒绝其谓词的超集（或用测试把两个定义钉在一起）；可选：在基类 `JSONReader` 中实现一次 flag 策略，各 reader 提供一行区域钩子。

</details>

core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderAdapter.java
```diff
@@ -248,14 +248,20 @@ public Object autoType(JSONReader jsonReader, Class expectClass, long features)
...
+                // an illegal type name is unresolvable, the same as the provider returning null
+            } else {
+                String typeName = jsonReader.getString();
+                autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, featuresAll);
+            }
```
wenshao, line 258: **[Suggestion]** Inverted guard with an empty `if` body — the actual work sits in the `else`, forcing every reader to mentally De Morgan the three-clause negative condition; the sibling consult in `ObjectReaderBean.checkAutoType0` expresses the same predicate as a positive `else if` carrying the action. One rejection rule now has two opposite shapes in the same PR; a future change to the handler-exemption rule must be applied to both, and the empty then-block reads like a missing statement (it passes the `EmptyBlock` checkstyle rule only because of the comment inside). — Concrete cost: lockstep maintenance of two opposite shapes of one rule, plus readability. The flat positive form is behaviour-identical (`autoTypeObjectReader` stays null either way):

```suggestion
            // an illegal type name is unresolvable, the same as the provider returning null;
            // skip the consult only when no handler could resolve the name
            if (!jsonReader.isTypeNameIllegal()
                    || context.getContextAutoTypeBeforeHandler() != null
                    || context.getProvider().getAutoTypeBeforeHandler() != null) {
                String typeName = jsonReader.getString();
                autoTypeObjectReader = context.getObjectReaderAutoType(typeName, expectClass, featuresAll);
            }
```

**[Suggestion]** 反向守卫带空 `if` 体——实际工作位于 `else` 中，读者必须对三子句否定条件做德摩根展开；`ObjectReaderBean.checkAutoType0` 中的兄弟消费点用带动作的正向 `else if` 表达同一谓词。同一条拒绝规则现在在同一 PR 中有两种相反形状；将来修改 handler 豁免规则必须同步改两处；空 then 块看起来像漏写语句（仅因内部注释才通过 `EmptyBlock` checkstyle 规则）。

具体代价：同一规则的两种相反形状需同步维护，外加可读性。扁平正向形式行为完全等价（两种写法下 `autoTypeObjectReader` 都保持 null），见 suggestion 块。

</details>

core/src/main/java/com/alibaba/fastjson2/util/IOUtils.java
```diff
@@ -2450,6 +2450,88 @@ public static int indexOfSlashV(byte[] buf, int fromIndex, int max) {
...
+     * {@link TypeUtils#hasIllegalTypeNameChars(String)}.
+     *
+     * @param buf the byte array to check
+     * @param off the starting offset in the array
+     * @param end the end offset (exclusive)
+     * @return true if the region contains {@code ':'} or {@code '!'}
+     */
+    public static boolean containsTypeNameSpecialChar(byte[] buf, int off, int end) {
+        int i = off;
+        long address = ARRAY_BYTE_BASE_OFFSET + off;
```
wenshao, line 2466: **[Suggestion]** Both new sweeps (`testContainsTypeNameSpecialCharBytes`/`Chars`) always pass `off = 0`, so mutations to the offset arithmetic survive: `ARRAY_BYTE_BASE_OFFSET + off` → `ARRAY_BYTE_BASE_OFFSET` (byte version, here) or `ARRAY_CHAR_BASE_OFFSET + ((long) off << 1)` → `+ off` (char version). Every production caller passes `strBegin`/`nameBegin` > 0. The current implementation is correct — probe: decoys outside the region ignored, hits inside detected in both vector and scalar tail — this is purely a test gap. — Failure scenario: the dropped-offset mutant flips real cases: ':' at index 10 in region `[8,24)`: clean impl `true` → mutant `false`, while the full byte sweep replayed under the mutant still passes. The false-negative direction is absorbed today because downstream `hasIllegalTypeNameChars` stays authoritative, but a future change relying on the scan alone would not be. — Witness (probe): byte sweep replayed under the mutant passes (len 1..33, all positions + near-misses); the off>0 flip above; all four reader-level UTF8 detect inputs still true under the mutant. — Suggested fix: run the same position sweep on a padded buffer with the region at a nonzero start — allocate `len + 8`, fill the prefix with junk including ':'/'!' decoys, scan `[8, 8+len)`, and assert the same true/false matrix (both byte and char overloads).

**[Suggestion]** 两个新扫描测试（`testContainsTypeNameSpecialCharBytes`/`Chars`）始终传 `off = 0`，因此偏移地址算术的突变体可以存活：`ARRAY_BYTE_BASE_OFFSET + off` → `ARRAY_BYTE_BASE_OFFSET`（byte 版，此处）或 `ARRAY_CHAR_BASE_OFFSET + ((long) off << 1)` → `+ off`（char 版）。所有生产调用者都传 `strBegin`/`nameBegin` > 0。当前实现是正确的——探针：区域外的诱饵字符被忽略，区域内命中在向量段与标量尾段均被检出——这纯粹是测试缺口。

失败场景：丢弃偏移的突变体会翻转真实用例：区域 `[8,24)` 索引 10 处的 ':'：干净实现 `true` → 突变体 `false`，而完整 byte 扫描在突变体下重放仍然通过。误漏方向今天被下游 `hasIllegalTypeNameChars` 兜底吸收，但将来依赖扫描本身的改动则不会。

证据（探针）：突变体下重放 byte 扫描全部通过（长度 1..33 × 每个位置 × 邻近字符）；上述 off>0 翻转；四个 reader 级 UTF8 检测输入在突变体下仍为 true。

修复建议：用非零起始的区域重放同样的位置扫描——分配 `len + 8`，前缀填充包含 ':'/'!' 诱饵的垃圾字节，扫描 `[8, 8+len)`，断言相同的 true/false 矩阵（byte 与 char 两个重载）。

</details>

core/src/test/java/com/alibaba/fastjson2/autoType/AutoTypeIllegalTypeNameTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        JSONReader reader2 = JSONReader.ofJSONB(jsonbBytes);
+        assertThrows(JSONException.class, () -> reader2.checkAutoType(
+                Object.class,
+                0,
+                JSONReader.Feature.SupportAutoType.mask | JSONReader.Feature.ErrorOnNotSupportAutoType.mask));
+    }
+
+    /**
+     * Short type names (up to 8 bytes) take the conservative path in JSONB and are
```
wenshao, line 186: **[Suggestion]** This javadoc misdescribes the code: short type names do NOT take a conservative path in JSONB. `readTypeHashCode0`'s gate explicitly INCLUDES the `BC_STR_ASCII_FIX_MIN..BC_STR_ASCII_FIX_MAX` range, and the JSONB writer emits `"A:B"` as a FIX strtype — probe: `writeTypeName("A:B")` emits `[BC_TYPED_ANY, 76 (= BC_STR_ASCII_FIX_MIN+3), 'A', ':', 'B']` and `isTypeNameIllegal()` measures `true`. So short names are flagged at scan time and early-rejected via `typeNameIllegal`; they never reach "the downstream validation". The assertions pass either way — only the comment is wrong. — Concrete cost: a maintainer reads this as evidence that JSONB has an "≤ 8 bytes" blind spot covered by downstream validation — no such blind spot exists; the comment could misdirect a future fix (e.g. deleting the scan-time gate for FIX strings believing this test still guards the short-name case via downstream checks, when it would then silently validate something else). — Suggested fix: state that short single-byte names are caught by the same scan-time check (via the FIX strtype range), or make the test actually cover a conservative JSONB path (e.g. a symbol-referenced type name) if that was the intent.

**[Suggestion]** 该 javadoc 与代码不符：短类型名在 JSONB 中并不走保守路径。`readTypeHashCode0` 的门控显式包含 `BC_STR_ASCII_FIX_MIN..BC_STR_ASCII_FIX_MAX` 范围，且 JSONB writer 把 `"A:B"` 写为 FIX strtype——探针：`writeTypeName("A:B")` 输出 `[BC_TYPED_ANY, 76 (= BC_STR_ASCII_FIX_MIN+3), 'A', ':', 'B']`，实测 `isTypeNameIllegal()` 为 true。因此短名在扫描期即被标记并经 `typeNameIllegal` 提前拒绝，从不经过"下游校验"。断言在两种情况下都通过——只有注释是错的。

具体代价：维护者会据此认为 JSONB 存在由下游校验兜底的 "≤ 8 字节" 盲区——该盲区不存在；注释可能误导未来修复（例如误以为该测试通过下游检查守护短名场景而删除 FIX 字符串的扫描期门控，届时该测试将静默地验证别的东西）。

修复建议：改为说明短的单字节名由同一扫描期检查（经 FIX strtype 范围）捕获；或者如果本意如此，让测试真正覆盖 JSONB 保守路径（如 symbol 引用的类型名）。

</details>

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR. Reviewed. Suggestions are inline.

<details>
<summary>中文说明</summary>

⚠️ 已从请求修改降级为评论：self-PR。 已审查。 建议见行内评论。

</details>

PR #7770, opened 2026-08-07
feat: add JSON.copyTo(Object, Object, JSONWriter.Feature...) for in-place property copy

xzxiaoshan:
## Summary

The existing `JSON.copyTo(Object, Class<T>, JSONWriter.Feature...)` always creates a **new** target instance. This is not only unnecessary overhead when a target object already exists, but in some scenarios the target object simply **cannot be re-instantiated** (e.g. framework-managed beans, proxied objects, objects with complex lifecycle state).

This PR adds an overload that copies properties **in-place** into an existing target object:

```java
static <T> T copyTo(Object object, Object targetObject, JSONWriter.Feature... features)
```

This is especially handy for **Map → target object** incremental property setting — copy only the fields you need onto an existing object, no new instance required.

The implementation delegates to `toJSON` + `JSONObject.copyTo`, keeping the same null-field semantics (null skipped by default, `WriteNulls` to overwrite).

<details>
<summary>中文</summary>

## 概述

现有 `JSON.copyTo(Object, Class<T>, ...)` 每次都会创建**新**的目标实例。这在目标对象已存在时不仅是不必要的开销，更关键的是某些场景下目标对象**无法重新实例化**（如框架管理的 Bean、代理对象、带有复杂生命周期状态的对象）。

本 PR 新增一个重载方法，将属性**就地**拷贝到已存在的目标对象中：

```java
static <T> T copyTo(Object object, Object targetObject, JSONWriter.Feature... features)
```

特别是 **Map → 目标对象**的增量属性设置场景非常好用——只把需要的字段拷到已有对象上，无需重新实例化。

实现委托 `toJSON` + `JSONObject.copyTo`，null 字段语义保持一致（默认跳过 null，开启 `WriteNulls` 则覆盖）。
</details>

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSON.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSON.java
@@ -4633,6 +4633,27 @@ public interface JSON {
     }
 
     /**
+     * Copies the properties of the specified source object into the given target object (in-place).
+     * No new instance is created; the target object is modified directly. Only fields with
+     * matching names between source and target are copied.
+     *
+     * @param object the source object whose properties will be copied
+     * @param targetObject the target object to receive the copied properties
+     * @param features the specified features are applied to serialization
+     * @param <T> the target type
+     * @return the target object after properties have been copied into it
+     * @since 2.0.65
+     */
+    @SuppressWarnings("unchecked")
+    static <T> T copyTo(Object object, Object targetObject, JSONWriter.Feature... features) {
+        if (object == null || targetObject == null) {
+            return (T) targetObject;
+        }
+        ((JSONObject) toJSON(object, features)).copyTo(targetObject);
+        return (T) targetObject;
+    }
+
+    /**
      * Configure the Enum classes as a JavaBean
      *
      * @param enumClasses the enum classes to configure as JavaBeans
--- a/core/src/test/java/com/alibaba/fastjson2/util/JSON_copyToObjectTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/util/JSON_copyToObjectTest.java
@@ -1,0 +1,177 @@
+package com.alibaba.fastjson2.util;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONWriter;
+import com.alibaba.fastjson2.annotation.JSONField;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDate;
+import java.util.Date;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for {@code JSON.copyTo(Object, Object, JSONWriter.Feature...)} — the in-place
+ * overload that copies properties from a source object into an existing target object.
+ *
+ * @since 2.0.65
+ */
+@Tag("util")
+public class JSON_copyToObjectTest {
+    @Test
+    public void copyTo_basic() {
+        Source source = new Source();
+        source.setId(101);
+        source.setName("fastjson");
+        source.setActive(true);
+
+        Target target = new Target();
+        Target result = JSON.copyTo(source, target);
+
+        assertSame(target, result, "copyTo should return the same target instance");
+        assertEquals(101, target.getId());
+        assertEquals("fastjson", target.getName());
+        assertTrue(target.isActive());
+    }
+
+    @Test
+    public void copyTo_nullEdgeCases() {
+        // null source: target unchanged
+        Target target = new Target();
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/alibaba/fastjson2/JSON.java
```diff
@@ -4632,6 +4632,27 @@ static <T> T copyTo(Object object, Class<T> targetClass, JSONWriter.Feature... f
...
+     *
+     * @param object the source object whose properties will be copied
+     * @param targetObject the target object to receive the copied properties
+     * @param features the specified features are applied to serialization
+     * @param <T> the target type
+     * @return the target object after properties have been copied into it
+     * @since 2.0.65
+     */
+    @SuppressWarnings("unchecked")
+    static <T> T copyTo(Object object, Object targetObject, JSONWriter.Feature... features) {
```
wenshao, line 4648: **[Critical]** R1-1: The return type is not constrained by `targetObject`, and the literal-null call does not reach this overload. — Failure scenario: `String value = JSON.copyTo(source, new Target())` compiles and then throws `ClassCastException` at the caller-side cast. Separately, `Target value = JSON.copyTo(source, null)` selects the existing `Class<T>` overload and fails instead of applying this method's null-target behavior. Bind the parameter as `T targetObject` and resolve the null-overload ambiguity (for example, also guard a null `targetClass`, or use a non-conflicting API signature).

**R1-1：** 返回类型没有受 `targetObject` 约束，而且传入字面量 `null` 时不会进入这个重载。失败场景：`String value = JSON.copyTo(source, new Target())` 可以编译，但返回时会在调用方类型转换处抛出 `ClassCastException`。此外，`Target value = JSON.copyTo(source, null)` 会选择现有的 `Class<T>` 重载并失败，而不是执行本方法声明的 null target 行为。建议将参数改为 `T targetObject`，并消除 null 重载歧义（例如同时处理 null `targetClass`，或采用不冲突的 API 签名）。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

core/src/main/java/com/alibaba/fastjson2/JSON.java
```diff
@@ -4632,6 +4632,27 @@ static <T> T copyTo(Object object, Class<T> targetClass, JSONWriter.Feature... f
...
+        if (object == null || targetObject == null) {
+            return (T) targetObject;
+        }
+        ((JSONObject) toJSON(object, features)).copyTo(targetObject);
```
wenshao, line 4652: **[Critical]** R1-2: The `toJSON` → `JSONObject.copyTo` bridge does not preserve the semantics required by this in-place copy API. — Failure scenario: the cast throws for collection/array/scalar sources; `Map`/`JSONObject` targets remain unchanged; `FieldBased` is not applied to the target; `rootName` wrappers prevent property matching; `ReferenceDetection` can overflow on cycles or duplicate shared children; and the fast path ignores `NotWriteDefaultValue`, null-substitution features, `IgnoreErrorGetter`, and `WriteEnumUsingToString`. These paths either throw, silently leave the target unchanged, overwrite values that should be preserved, or store the wrong representation. Please generalize the existing direct `ObjectWriterAdapter`/`ObjectReaderBean` copy path to accept an existing target and derive writer/reader features once, with an explicit validated fallback for unsupported shapes.

**R1-2：** `toJSON` → `JSONObject.copyTo` 的桥接无法保持就地拷贝 API 所需的语义。失败场景包括：集合、数组或标量源对象在强制转换处抛异常；`Map`/`JSONObject` 目标保持不变；`FieldBased` 没有应用到目标读取；`rootName` 包装导致属性无法匹配；`ReferenceDetection` 在循环引用时可能栈溢出、在共享子对象时破坏引用身份；快路径还忽略 `NotWriteDefaultValue`、null 替换特性、`IgnoreErrorGetter` 和 `WriteEnumUsingToString`。这些路径会抛异常、静默不修改目标、覆盖本应保留的值，或写入错误的表示形式。建议扩展现有的 `ObjectWriterAdapter`/`ObjectReaderBean` 直接拷贝路径，使其接受已有目标对象并统一派生 writer/reader features，仅对不支持的形态使用显式校验后的 fallback。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

wenshao (changes requested):
Not reviewed: reverse audit — did not converge within the reverse-audit round cap of 5.

<details>
<summary>中文说明</summary>

未审查：反向审计——在 5 轮的反审轮数上限内未收敛。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

PR #7778, opened 2026-08-12
fix: make JSONPath.contains parse JSON text like eval

### LINKED ISSUE #7771: [BUG]JSONPath.contains和JSONPath.eval输出结果不一致
### 问题描述
JSONPath.contains结果为false，但JSONPath.eval可以取到值

### 环境信息
 - OS信息：  NA
 - JDK信息：Openjdk 21.0.8
 - 版本信息：Fastjson2 2.0.60
 

### 重现步骤
    private void test1() {
        String  s = """
            {"status": 200}
            """;

        boolean contains = JSONPath.contains(s, "$.status");
        Object eval = JSONPath.eval(s, "$.status");
        System.out.println(eval); // 返回 200
        System.out.println(contains); // 返回 false 期望返回 true
    }

kalayciburak:
## Summary

`JSONPath.eval(String, String)` treats the first argument as JSON text (via `extract`), but `JSONPath.contains` only had an `Object` overload. Passing a JSON string therefore resolved to bean-style `contains` and returned `false` for ordinary paths such as `$.status` even when `eval` returned the value.

Add `JSONPath.contains(String, String)` that parses the text and delegates to the existing object path, so path presence matches `eval`/`extract` and null-valued keys still count as present.

Fixes #7771

<details>
<summary>中文说明</summary>

`JSONPath.eval(String, String)` 会把第一个参数当作 JSON 文本解析，而 `contains` 原先只有 `Object` 重载。传入 JSON 字符串时会走 Java bean 路径，导致 `$.status` 这类路径返回 `false`，与 `eval` 不一致。

新增 `JSONPath.contains(String, String)`：先 `JSON.parse`，再委托现有对象路径判断，与 `eval`/`extract` 语义对齐，且 `null` 值字段仍视为路径存在。

</details>

## Test plan

- [x] `mvn -pl core -Dtest=Issue7771,Issue643 test` — 6/6 GREEN (RED→GREEN on Issue7771)
- [x] `mvn -pl core -Dtest='com.alibaba.fastjson2.jsonpath.**,com.alibaba.fastjson2.issues.Issue643,com.alibaba.fastjson2.issues_7000.Issue7771' test` — 403/403 GREEN
- [x] `mvn -pl core validate` GREEN

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
@@ -227,6 +227,26 @@ public abstract class JSONPath {
     }
 
     /**
+     * Checks if the specified path exists in JSON text.
+     * <p>
+     * Mirrors {@link #eval(String, String)} / {@link #extract(String, String)}: the first
+     * argument is parsed as JSON, not treated as a Java bean. Without this overload,
+     * {@code contains(jsonText, path)} resolves to {@link #contains(Object, String)} and
+     * incorrectly returns false for ordinary property paths such as {@code $.status}.
+     *
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        return JSONPath.of(path).contains(JSON.parse(json));
+    }
+
+    /**
      * Sets a value in an object using the specified path
      *
      * @param rootObject the root object to modify
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
@@ -246,9 +246,6 @@ public abstract class JSONPath {
         }
 
         Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
-        if ("$".equals(path)) {
-            return parsed != null;
-        }
         return contains(parsed, path);
     }
 
@@ -1557,7 +1554,7 @@ public abstract class JSONPath {
 
         @Override
         public boolean contains(Object object) {
-            return false;
+            return object != null;
         }
 
         @Override
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
@@ -233,6 +233,8 @@ public abstract class JSONPath {
      * argument is parsed as JSON, not treated as a Java bean. Without this overload,
      * {@code contains(jsonText, path)} resolves to {@link #contains(Object, String)} and
      * incorrectly returns false for ordinary property paths such as {@code $.status}.
+     * The root path {@code $} is present for every non-null document.
+     * {@code $ref} is not resolved, matching {@link #extract(String, String)}.
      *
      * @param json the JSON text to check
      * @param path the JSONPath expression
@@ -243,7 +245,11 @@ public abstract class JSONPath {
             return false;
         }
 
-        return contains(JSON.parse(json), path);
+        Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
+        if ("$".equals(path)) {
+            return parsed != null;
+        }
+        return contains(parsed, path);
     }
 
     /**
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPath.java
@@ -243,7 +243,7 @@ public abstract class JSONPath {
             return false;
         }
 
-        return JSONPath.of(path).contains(JSON.parse(json));
+        return contains(JSON.parse(json), path);
     }
 
     /**
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
@@ -34,6 +34,17 @@ abstract class JSONPathSegment {
         return context.value != null;
     }
 
+    static boolean hasMatch(Object value) {
+        if (value instanceof Collection) {
+            return !((Collection<?>) value).isEmpty();
+        }
+        if (value instanceof JSONPath.Sequence) {
+            List<?> values = ((JSONPath.Sequence) value).values;
+            return values != null && !values.isEmpty();
+        }
+        return value != null;
+    }
+
     public boolean remove(JSONPath.Context context) {
         throw new JSONException("UnsupportedOperation " + getClass());
     }
@@ -266,6 +277,12 @@ abstract class JSONPathSegment {
             }
 
             throw new JSONException("TODO");
+        }
+
+        @Override
+        public boolean contains(JSONPath.Context context) {
+            eval(context);
+            return hasMatch(context.value);
         }
 
         @Override
@@ -562,6 +579,12 @@ abstract class JSONPathSegment {
         }
 
         @Override
+        public boolean contains(JSONPath.Context context) {
+            eval(context);
+            return hasMatch(context.value);
+        }
+
+        @Override
         public void accept(JSONReader jsonReader, JSONPath.Context context) {
             if (context.parent != null
                     && context.parent.current instanceof CycleNameSegment
@@ -777,6 +800,24 @@ abstract class JSONPathSegment {
                 array.add(fieldValue);
             }
             context.value = array;
+        }
+
+        @Override
+        public boolean contains(JSONPath.Context context) {
+            Object object = context.parent == null
+                    ? context.root
+                    : context.parent.value;
+            if (object instanceof Map) {
+                Map<?, ?> map = (Map<?, ?>) object;
+                for (String name : names) {
+                    if (map.containsKey(name)) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+            eval(context);
+            return hasMatch(context.value);
         }
 
         @Override
@@ -1055,6 +1096,12 @@ abstract class JSONPathSegment {
             }
             context.value = array;
             context.eval = true;
+        }
+
+        @Override
+        public boolean contains(JSONPath.Context context) {
+            eval(context);
+            return hasMatch(context.value);
         }
 
         @Override
@@ -1388,6 +1435,22 @@ abstract class JSONPathSegment {
             }
 
             context.eval = true;
+        }
+
+        @Override
+        public boolean contains(JSONPath.Context context) {
+            Object object = context.parent == null
+                    ? context.root
+                    : context.parent.value;
+            List values = new JSONArray();
+            Consumer action;
+            if (shouldRecursive()) {
+                action = new MapRecursive(context, values, 0);
+            } else {
+                action = new MapLoop(context, values);
+            }
+            action.accept(object);
+            return !values.isEmpty();
         }
 
         @Override
--- a/core/src/test/java/com/alibaba/fastjson2/JSONPathTypedMultiTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/JSONPathTypedMultiTest.java
@@ -200,6 +200,7 @@ public class JSONPathTypedMultiTest {
                 new String[]{"$", "$.values[0].name", "$.values[0].date"},
                 new Type[]{JSONObject.class, String.class, Date.class}
         );
+        assertTrue(JSONPath.contains(object, "$"));
         assertFalse(jsonPath.contains(object));
         assertTrue(jsonPath.isRef());
     }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
@@ -33,4 +33,10 @@ public class Issue7771 {
         assertTrue(JSONPath.contains(object, "$.status"));
         assertFalse(JSONPath.contains(object, "$.missing"));
     }
+
+    @Test
+    public void containsOnJsonNullLiteralIsFalse() {
+        assertFalse(JSONPath.contains("null", "$.status"));
+        assertFalse(JSONPath.contains("null", "$"));
+    }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
@@ -38,5 +38,30 @@ public class Issue7771 {
     public void containsOnJsonNullLiteralIsFalse() {
         assertFalse(JSONPath.contains("null", "$.status"));
         assertFalse(JSONPath.contains("null", "$"));
+        assertFalse(JSONPath.contains((String) null, "$.status"));
+        assertFalse(JSONPath.contains("", "$.status"));
+    }
+
+    @Test
+    public void containsOnRootPathMatchesEval() {
+        assertTrue(JSONPath.contains("{\"status\":200}", "$"));
+        assertTrue(JSONPath.contains("123", "$"));
+        assertTrue(JSONPath.contains("[1,2]", "$"));
+        assertEquals(JSON.parse("{\"status\":200}"), JSONPath.eval("{\"status\":200}", "$"));
+    }
+
+    @Test
+    public void containsOnArrayRoot() {
+        assertTrue(JSONPath.contains("[1,2,3]", "$[1]"));
+        assertFalse(JSONPath.contains("[1,2,3]", "$[5]"));
+        assertFalse(JSONPath.contains("[null]", "$.status"));
+        assertFalse(JSONPath.contains("[null,{\"status\":1}]", "$.status"));
+        assertTrue(JSONPath.contains("[{\"status\":1},null]", "$.status"));
+    }
+
+    @Test
+    public void containsOnRefCycleDoesNotOverflow() {
+        assertTrue(JSONPath.contains("{\"a\":{\"$ref\":\"$\"}}", "$..a"));
+        assertTrue(JSONPath.contains("{\"x\":{\"y\":{\"$ref\":\"$\"}}}", "$..x"));
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
@@ -1,12 +1,14 @@
 package com.alibaba.fastjson2.issues_7000;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONException;
 import com.alibaba.fastjson2.JSONObject;
 import com.alibaba.fastjson2.JSONPath;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class Issue7771 {
@@ -48,6 +50,7 @@ public class Issue7771 {
         assertTrue(JSONPath.contains("123", "$"));
         assertTrue(JSONPath.contains("[1,2]", "$"));
         assertEquals(JSON.parse("{\"status\":200}"), JSONPath.eval("{\"status\":200}", "$"));
+        assertTrue(JSONPath.contains(JSON.parse("{\"a\":1}"), "$"));
     }
 
     @Test
@@ -57,6 +60,33 @@ public class Issue7771 {
         assertFalse(JSONPath.contains("[null]", "$.status"));
         assertFalse(JSONPath.contains("[null,{\"status\":1}]", "$.status"));
         assertTrue(JSONPath.contains("[{\"status\":1},null]", "$.status"));
+        assertTrue(JSONPath.contains("[null]", "$[0]"));
+        assertTrue(JSONPath.contains("{\"a\":[null]}", "$.a[0]"));
+        assertTrue(JSONPath.contains("[[]]", "$[0]"));
+    }
+
+    @Test
+    public void containsTreatsPresentNullAsPresent() {
+        assertTrue(JSONPath.contains("{\"a\":[{\"status\":null}]}", "$.a.status"));
+        assertTrue(JSONPath.contains("{\"a\":{\"status\":null}}", "$.a.status"));
+        assertTrue(JSONPath.contains("[{\"status\":null}]", "$.status"));
+        assertTrue(JSONPath.contains("{\"a\":[{\"status\":null}]}", "$.a[0].status"));
+    }
+
+    @Test
+    public void containsOnEmptyMatchSetIsFalse() {
+        assertFalse(JSONPath.contains("{}", "$.*"));
+        assertFalse(JSONPath.contains("{}", "$..a"));
+        assertFalse(JSONPath.contains("[1,2]", "$[5,6]"));
+        assertFalse(JSONPath.contains("[1,2]", "$[?(@>9)]"));
+        assertFalse(JSONPath.contains("[1,2]", "$[5:6]"));
+        assertFalse(JSONPath.contains("[1,2]", "$[0:0]"));
+        assertFalse(JSONPath.contains("{}", "$['x','y']"));
+    }
+
+    @Test
+    public void containsOnMalformedJsonThrows() {
+        assertThrows(JSONException.class, () -> JSONPath.contains("{invalid json", "$.status"));
     }
 
     @Test
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
@@ -1,0 +1,36 @@
+package com.alibaba.fastjson2.issues_7000;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONObject;
+import com.alibaba.fastjson2.JSONPath;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue7771 {
+    @Test
+    public void containsOnJsonStringMatchesEval() {
+        String json = "{\"status\": 200}";
+
+        assertEquals(200, JSONPath.eval(json, "$.status"));
+        assertTrue(JSONPath.contains(json, "$.status"));
+        assertFalse(JSONPath.contains(json, "$.missing"));
+    }
+
+    @Test
+    public void containsOnJsonStringNullValueStillPresent() {
+        String json = "{\"status\": null}";
+
+        assertTrue(JSONPath.contains(json, "$.status"));
+        assertFalse(JSONPath.contains(json, "$.missing"));
+    }
+
+    @Test
+    public void containsOnParsedObjectUnchanged() {
+        JSONObject object = JSON.parseObject("{\"status\": 200}");
+        assertTrue(JSONPath.contains(object, "$.status"));
+        assertFalse(JSONPath.contains(object, "$.missing"));
+    }
+}
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathFilter.java
@@ -32,6 +32,12 @@ abstract class JSONPathFilter
 
     abstract boolean apply(JSONPath.Context context, Object object);
 
+    @Override
+    public boolean contains(JSONPath.Context context) {
+        eval(context);
+        return hasMatch(context.value);
+    }
+
     enum Operator {
         EQ,
         NE,
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentIndex.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentIndex.java
@@ -200,6 +200,53 @@ final class JSONPathSegmentIndex
     }
 
     @Override
+    public boolean contains(JSONPath.Context context) {
+        Object object = context.parent == null
+                ? context.root
+                : context.parent.value;
+        if (object == null) {
+            return false;
+        }
+        if (object instanceof List) {
+            int size = ((List<?>) object).size();
+            int i = index >= 0 ? index : size + index;
+            return i >= 0 && i < size;
+        }
+        if (object instanceof Object[]) {
+            int size = ((Object[]) object).length;
+            int i = index >= 0 ? index : size + index;
+            return i >= 0 && i < size;
+        }
+        Class<?> objectClass = object.getClass();
+        if (objectClass.isArray()) {
+            int size = Array.getLength(object);
+            int i = index >= 0 ? index : size + index;
+            return i >= 0 && i < size;
+        }
+        if (object instanceof Collection) {
+            int size = ((Collection<?>) object).size();
+            int i = index >= 0 ? index : size + index;
+            return i >= 0 && i < size;
+        }
+        if (object instanceof JSONPath.Sequence) {
+            for (Object item : ((JSONPath.Sequence) object).values) {
+                context.value = item;
+                JSONPath.Context itemContext = new JSONPath.Context(
+                        context.path, context, context.current, context.next, context.readerFeatures);
+                if (contains(itemContext)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+        if (object instanceof Map) {
+            Map<?, ?> map = (Map<?, ?>) object;
+            return map.containsKey(index) || map.containsKey(Integer.toString(index));
+        }
+        return index == 0;
+    }
+
+    @Override
     public void set(JSONPath.Context context, Object value) {
         Object object = context.parent == null
                 ? context.root
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
@@ -98,9 +98,10 @@ class JSONPathSegmentName
                 }
 
                 if (item instanceof Map) {
-                    if (((Map<?, ?>) item).get(name) != null) {
+                    if (((Map<?, ?>) item).containsKey(name)) {
                         return true;
                     }
+                    continue;
                 }
 
                 ObjectWriter<?> objectWriter = context.path
@@ -126,9 +127,10 @@ class JSONPathSegmentName
                 }
 
                 if (item instanceof Map) {
-                    if (((Map<?, ?>) item).get(name) != null) {
+                    if (((Map<?, ?>) item).containsKey(name)) {
                         return true;
                     }
+                    continue;
                 }
 
                 ObjectWriter<?> objectWriter = context.path
@@ -154,9 +156,10 @@ class JSONPathSegmentName
                 }
 
                 if (item instanceof Map) {
-                    if (((Map) item).get(name) != null) {
+                    if (((Map) item).containsKey(name)) {
                         return true;
                     }
+                    continue;
                 }
 
                 ObjectWriter<?> objectWriter = context.path
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSingleName.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSingleName.java
@@ -111,6 +111,10 @@ final class JSONPathSingleName
 
     @Override
     public boolean contains(Object root) {
+        if (root == null) {
+            return false;
+        }
+
         if (root instanceof Map) {
             return ((Map) root).containsKey(name);
         }
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathTypedMulti.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathTypedMulti.java
@@ -50,6 +50,9 @@ class JSONPathTypedMulti
     @Override
     public boolean contains(Object object) {
         for (JSONPath jsonPath : paths) {
+            if (jsonPath instanceof RootPath) {
+                continue;
+            }
             if (jsonPath.contains(object)) {
                 return true;
             }


core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
...
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        return JSONPath.of(path).contains(JSON.parse(json));
```
wenshao: **[Critical]** Parsed JSON `null` can still reach a null-unsafe path implementation — Failure scenario: `JSONPath.contains("null", "$.status")` (and parser-accepted empty input) produces a null root, then `JSONPathSingleName.contains` dereferences `root.getClass()`, throwing `NullPointerException` instead of returning `false`. Delegate the parsed value through the existing null-safe `contains(Object, String)` overload and add null-literal/empty-input regression cases.

**问题：** 解析后的 JSON `null` 仍可能进入不支持空根对象的路径实现。

**失败场景：** `JSONPath.contains("null", "$.status")`（以及解析器接受的空输入）会得到 `null` 根对象，随后 `JSONPathSingleName.contains` 解引用 `root.getClass()`，抛出 `NullPointerException`，而不是返回 `false`。建议将解析结果委托给已有的空值安全重载 `contains(Object, String)`，并补充 JSON `null`/空输入回归测试。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
...
+     * Mirrors {@link #eval(String, String)} / {@link #extract(String, String)}: the first
+     * argument is parsed as JSON, not treated as a Java bean. Without this overload,
+     * {@code contains(jsonText, path)} resolves to {@link #contains(Object, String)} and
+     * incorrectly returns false for ordinary property paths such as {@code $.status}.
+     *
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
```
wenshao, line 243: **[Critical]** This overload cannot preserve existing Java `String`-root behavior or fix previously compiled callers — Failure scenario: after recompilation, ordinary non-JSON strings bind to `contains(String, String)` and can throw `JSONException`; without recompilation, existing bytecode remains linked to `contains(Object, String)` and keeps the original incorrect result for JSON text. Use an explicit JSON-text API or another compatibility strategy that does not reinterpret every Java `String` root and clearly addresses already compiled callers.

**问题：** 该重载既无法保留已有 Java `String` 根对象语义，也无法修复已编译调用方。

**失败场景：** 重新编译后，普通非 JSON 字符串会绑定到 `contains(String, String)`，并可能抛出 `JSONException`；不重新编译时，已有字节码仍链接到 `contains(Object, String)`，对 JSON 文本继续得到原来的错误结果。建议使用显式的 JSON 文本 API，或采用既不会重新解释所有 Java `String` 根对象、又能明确处理已编译调用方的兼容方案。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
wenshao: **[Critical]** A present `null` array element is reported as absent — Failure scenario: `JSONPath.contains("[null]", "$[0]")` parses an in-bounds one-element array, but index containment is implemented as `value != null`, so index 0 returns `false` although it exists, contradicting the PR's present-null property contract. Make index containment check bounds/presence independently of the element value and add present-null/out-of-range array tests.

**问题：** 数组中实际存在但值为 `null` 的元素会被判断为不存在。

**失败场景：** `JSONPath.contains("[null]", "$[0]")` 解析得到包含一个元素的数组，但索引存在性使用 `value != null` 判断，因此索引 0 明明存在却返回 `false`，与本 PR 对对象属性“值为 null 仍视为存在”的语义冲突。建议按索引边界/位置判断存在性，不依赖元素值，并补充 null 元素与越界索引测试。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

wenshao: **[Suggestion]** The convenience overload materializes the entire JSON document — Concrete cost: for a large document whose requested property appears near the beginning, `JSON.parse` still allocates and parses the complete irrelevant tail before `contains` runs, adding memory and GC cost proportional to the whole input. Consider a `JSONReader`-based presence operation that skips irrelevant values while distinguishing present `null` from a missing path.

**建议：** 该便捷重载会完整构建整个 JSON 文档。

**具体成本：** 即使目标属性位于大型文档开头，`JSON.parse` 仍会分配并解析后续全部无关内容，导致内存占用与 GC 成本随完整输入增长。可考虑基于 `JSONReader` 的存在性检查，跳过无关值，同时区分“存在但为 null”和“路径缺失”。

</details>

_— gpt-5.6-sol via Qwen Code /review (v0.21.10)_

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
         return jsonPath.contains(rootObject);
     }
 
+    /**
+     * Checks if the specified path exists in JSON text.
+     * <p>
+     * Mirrors {@link #eval(String, String)} / {@link #extract(String, String)}: the first
+     * argument is parsed as JSON, not treated as a Java bean. Without this overload,
```
wenshao, line 233: **[Critical]** `contains(jsonText, "$")` returns `false` for **every** non-null JSON document, contradicting this javadoc — "Mirrors `eval(String, String)` / `extract(String, String)`" and "@return true if the path exists". The root path exists in every non-null document, and `eval("{\"status\":200}", "$")` returns the parsed object while `contains` says it is absent. It is also self-inconsistent: `contains(json, "$.status")` = `true` while its parent path `contains(json, "$")` = `false`.

**Failure scenario:** probe-verified on this PR's compiled classes — `contains("{\"status\":200}", "$")`, `contains("123", "$")`, `contains("[1,2]", "$")` all return `false`, while `eval`/`extract` of the same texts return the root. Root cause is pre-existing (`RootPath.contains` is hardcoded `return false`, JSONPath.java ~1553), but this diff creates the public text entry point and the contract that stub breaks. A caller replacing the natural idiom `JSONPath.eval(json, "$") != null` with `JSONPath.contains(json, "$")` gets an inverted answer for every valid document. The only `"$"` test added uses the `"null"` document, which short-circuits at the Object overload's null guard and never reaches `RootPath.contains` — so the suite passes under either semantics (flip check: making `RootPath.contains` return `object != null` turns all the probes above to `true` and keeps all four PR tests green).

**Suggested fix:** make `RootPath.contains` return `object != null` (fixes both overloads; probe-verified the PR tests stay green), or special-case `"$"` in the new overload — at minimum, document the limitation and pin the chosen semantics with `assertTrue(JSONPath.contains("{\"status\":200}", "$"))`.

**[严重]** `contains(jsonText, "$")` 对**任何**非 null JSON 文档都返回 `false`，与这段 javadoc 矛盾——"Mirrors `eval(String, String)` / `extract(String, String)`" 与 "@return true if the path exists"。根路径在每个非 null 文档中都存在，`eval("{\"status\":200}", "$")` 返回解析出的对象，而 `contains` 却判定不存在。且自相矛盾：`contains(json, "$.status")` = `true`，其父路径 `contains(json, "$")` = `false`。

**失败场景：** 已在本 PR 编译产物上探针实测——`contains("{\"status\":200}", "$")`、`contains("123", "$")`、`contains("[1,2]", "$")` 全部返回 `false`，而同一文本的 `eval`/`extract` 返回根对象。根因是既有代码（`RootPath.contains` 硬编码 `return false`，JSONPath.java ~1553），但本 diff 新增了公开文本入口并写下了被该桩实现打破的契约。调用方把自然写法 `JSONPath.eval(json, "$") != null` 替换为 `JSONPath.contains(json, "$")` 时，对每个有效文档都得到相反答案。新增的唯一 `"$"` 测试使用 `"null"` 文档，在 Object 重载的 null 防护处就短路，到不了 `RootPath.contains`——因此套件在两种语义下均通过（flip check：把 `RootPath.contains` 改为 `return object != null` 后上述探针全部翻正，且 PR 的 4 个测试仍全绿）。

**建议修复：** 将 `RootPath.contains` 改为 `return object != null`（同时修复两个重载；已探针验证 PR 测试仍全绿），或在新重载中对 `"$"` 特判——至少应文档化该限制，并用 `assertTrue(JSONPath.contains("{\"status\":200}", "$"))` 钉住所选语义。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
...
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        return contains(JSON.parse(json), path);
```
wenshao: **[Critical]** Array JSON text whose first element is null crashes this overload with `NullPointerException` — exactly the contains/eval divergence issue #7771 reports. `JSON.parse("[null]")` yields a `JSONArray` whose element 0 is null; `JSONPathSingleName.contains` takes its `root instanceof List` branch and recurses with `list.get(0)`; the null falls through both `instanceof` checks to `provider.getObjectWriter(root.getClass())` → NPE.

**Failure scenario:** probe-verified on this PR's compiled classes — `contains("[null]", "$.status")` and `contains("[null,{\"status\":1}]", "$.status")` throw NPE, while `contains("[{\"status\":1},null]", "$.status")` returns `true`, `eval("[null]", "$.status")` returns `null`, and the pre-PR binding `contains((Object) "[null]", "$.status")` returned `false`. So this overload flips well-formed input from "returns false" to "throws NPE". The new tests never exercise an array root, so this ships green.

**Suggested fix:** make `JSONPathSingleName.contains` null-safe (e.g. `if (root == null) return false;` at the top, or skip a null first element in its List branch) and add a regression case such as `assertFalse(JSONPath.contains("[null]", "$.status"));` to `Issue7771`.

**[严重]** 首元素为 null 的数组 JSON 文本会使本重载抛出 `NullPointerException`——这正是 issue #7771 所抱怨的 contains/eval 不一致。`JSON.parse("[null]")` 得到首元素为 null 的 `JSONArray`；`JSONPathSingleName.contains` 走 `root instanceof List` 分支并以 `list.get(0)` 递归；null 穿过两个 `instanceof` 分支落到 `provider.getObjectWriter(root.getClass())` → NPE。

**失败场景：** 已在本 PR 编译产物上探针实测——`contains("[null]", "$.status")` 与 `contains("[null,{\"status\":1}]", "$.status")` 抛 NPE，而 `contains("[{\"status\":1},null]", "$.status")` 返回 `true`、`eval("[null]", "$.status")` 返回 `null`、PR 前绑定 `contains((Object) "[null]", "$.status")` 返回 `false`。即本重载把合法输入从"返回 false"翻转为"抛 NPE"。新测试从未覆盖数组根，因此可以静默合入。

**建议修复：** 让 `JSONPathSingleName.contains` 空值安全（如顶部加 `if (root == null) return false;`，或其 List 分支跳过 null 首元素），并在 `Issue7771` 补回归用例如 `assertFalse(JSONPath.contains("[null]", "$.status"));`。

</details>

wenshao: **[Critical]** This overload is the only text-based JSONPath API that dies with `StackOverflowError` on well-formed JSON containing a `$ref` cycle, because `JSON.parse` resolves `$ref` into a real cyclic structure and the `$..name` walk (`CycleNameSegment`'s `MapLoop` — `shouldRecursive` is only true for `*`/empty names) has no visited-set or depth guard. A 16-byte attacker-controlled document kills the thread with an `Error` that `catch (Exception)` does not intercept.

**Failure scenario:** probe-verified on this PR's compiled classes — `contains("{\"a\":{\"$ref\":\"$\"}}", "$..a")` → `StackOverflowError` (recursing in `JSONPathSegment$CycleNameSegment$MapLoop.accept`); same for `{"x":{"y":{"$ref":"$"}}}` + `$..x`. The same input through the sibling APIs this javadoc cites all survive: `eval`/`extract` (streaming, never materialize the cycle) return `[{"$ref":"$"}]`, `set`/`remove` (mutate before descending, breaking the cycle) complete. Pre-PR binding `contains((Object) text, "$..a")` returned `true` — so this overload flips well-formed input from "returns a boolean" to "thread-killing Error", an attacker-reachable DoS on the new public API. (Overflow on hand-built cyclic objects via the Object overload is pre-existing; what this diff adds is the single-call text entry point and the divergence from every other text API.)

**Suggested fix:** align with the other text APIs: implement `contains(String, String)` over a `JSONReader` the way `extract(String, String)` does (probe-verified viable and semantically correct on this input), or parse without resolving `$ref` into cycles for this predicate, or add a visited-set/depth guard to the `MapLoop` walk (which would also fix the pre-existing Object-path overflow).

**[严重]** 本重载是唯一会被含 `$ref` 循环的合法 JSON 打死的文本系 JSONPath API：`JSON.parse` 会把 `$ref` 解析成真实循环结构，而 `$..name` 的行走（`CycleNameSegment` 的 `MapLoop`——`shouldRecursive` 仅对 `*`/空名为 true）没有访问集或深度防护。16 字节的攻击者可控文档即可用 `catch (Exception)` 拦不住的 `Error` 杀死线程。

**失败场景：** 已在本 PR 编译产物上探针实测——`contains("{\"a\":{\"$ref\":\"$\"}}", "$..a")` → `StackOverflowError`（在 `JSONPathSegment$CycleNameSegment$MapLoop.accept` 中无限递归）；`{"x":{"y":{"$ref":"$"}}}` + `$..x` 同样。同一输入走 javadoc 点名的兄弟 API 全部存活：`eval`/`extract`（流式，不物化循环）返回 `[{"$ref":"$"}]`，`set`/`remove`（先改写再下探，打断循环）正常完成。PR 前绑定 `contains((Object) text, "$..a")` 返回 `true`——本重载把合法输入从"返回布尔"翻转为"线程级 Error"，是新公开 API 上攻击者可达的 DoS。（Object 重载对手工构造循环对象的溢出为既有问题；本 diff 新增的是单调用文本入口及与所有其他文本 API 的分歧。）

**建议修复：** 与其他文本 API 对齐：像 `extract(String, String)` 那样基于 `JSONReader` 实现（已探针验证在该输入上可行且语义正确），或为本谓词解析时不把 `$ref` 解析成循环，或在 `MapLoop` 行走上加 visited-set/深度防护（顺带修复既有 Object 路径溢出）。

</details>

wenshao: **[Suggestion]** Array-root JSON text is never exercised — all four tests use object roots or the `"null"` literal, so a mutant replacing `JSON.parse(json)` with `JSON.parseObject(json)` passes the whole new suite unchanged.

**Failure scenario:** mutation-verified by probe — on every test input the `parseObject` mutant behaves identically, but it diverges on `JSONPath.contains("[1,2,3]", "$[1]")`: the original returns `true`, the mutant throws (`parseObject` rejects array text). Top-level arrays are a common shape for `contains` (cf. the existing `Issue3331`, which runs `contains` against an array root).

**Suggested fix:** add `assertTrue(JSONPath.contains("[1,2,3]", "$[1]"));` and `assertFalse(JSONPath.contains("[1,2,3]", "$[5]"));` — both kill this mutant.

**[建议]** 数组根 JSON 文本完全未被覆盖——4 个测试全部使用对象根或 `"null"` 字面量，因此把 `JSON.parse(json)` 替换为 `JSON.parseObject(json)` 的突变体可以不改动地通过全部新测试。

**失败场景：** 已经探针实测突变体——对每个现有测试输入 `parseObject` 突变体行为一致，但在 `JSONPath.contains("[1,2,3]", "$[1]")` 上发散：原实现返回 `true`，突变体抛异常（`parseObject` 拒绝数组文本）。顶层数组是 `contains` 的常见输入形态（参见既有 `Issue3331`，它对数组根做 `contains`）。

**建议修复：** 补 `assertTrue(JSONPath.contains("[1,2,3]", "$[1]"));` 与 `assertFalse(JSONPath.contains("[1,2,3]", "$[5]"));`——两者即可杀死该突变体。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
...
+     * Checks if the specified path exists in JSON text.
+     * <p>
+     * Mirrors {@link #eval(String, String)} / {@link #extract(String, String)}: the first
+     * argument is parsed as JSON, not treated as a Java bean. Without this overload,
+     * {@code contains(jsonText, path)} resolves to {@link #contains(Object, String)} and
+     * incorrectly returns false for ordinary property paths such as {@code $.status}.
+     *
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
```
wenshao, line 241: **[Suggestion]** The contract documented here omits two probe-verified divergences from the methods this javadoc cites:

1. **Non-JSON input throws.** `contains("not json", "$.a")` throws `JSONException` (via `JSON.parse`), where the pre-PR binding returned `false` — and `eval("not json", "$.a")` returns `null`, so the javadoc's own reference method contradicts the "Mirrors eval" claim on identical input. The "@return true if the path exists, false otherwise" line omits the exception entirely.
2. **Trailing content.** `JSON.parse` enforces an end-of-input check (`JSON.java:152-154`, throws "input not end") while `extract`/`eval` do not — `contains("{\"a\":1} trailing", "$.a")` throws, `eval` of the same text returns `1`.

**Failure scenario:** a service using `contains` as a boolean predicate over arbitrary strings gains a new uncaught-exception path on recompilation; a caller trusting "Mirrors eval" gets an exception on trailing-content input where `eval` returns a value. (For the record: `eval("", "$.a")` returns `null` rather than throwing, so empty input is *not* a divergence.)

**Suggested fix:** either document the `JSONException` behavior and scope the "Mirrors" claim accordingly (no `@throws` precedent exists in this file; `set(String, String, Object)` has the identical exposure undocumented), or make the predicate total (`catch JSONException → false`).

**[建议]** 此处文档化的契约遗漏了两处经探针实测、与 javadoc 点名方法相悖的分歧：

1. **非 JSON 输入抛异常。** `contains("not json", "$.a")` 抛 `JSONException`（来自 `JSON.parse`），而 PR 前绑定返回 `false`——且 `eval("not json", "$.a")` 返回 `null`，即 javadoc 点名的参照方法自身就与同一输入上的 "Mirrors eval" 声明矛盾。"@return true if the path exists, false otherwise" 完全没有提及异常。
2. **尾随内容。** `JSON.parse` 强制输入结束检查（`JSON.java:152-154`，抛 "input not end"），而 `extract`/`eval` 不做——`contains("{\"a\":1} trailing", "$.a")` 抛异常，`eval` 同一文本返回 `1`。

**失败场景：** 以 `contains` 作为任意字符串布尔判定的服务在重新编译后获得新的未捕获异常路径；信任 "Mirrors eval" 的调用方在尾随内容输入上一边得到异常、一边得到值。（另：`eval("", "$.a")` 返回 `null` 而非抛异常，故空输入**不**构成分歧。）

**建议修复：** 二选一——文档化 `JSONException` 行为并相应限定 "Mirrors" 的适用范围（本文件无 `@throws` 先例；`set(String, String, Object)` 有完全相同的未文档化暴露面）；或把谓词做成全函数（`catch JSONException → false`）。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
         return jsonPath.contains(rootObject);
     }
 
+    /**
+     * Checks if the specified path exists in JSON text.
+     * <p>
+     * Mirrors {@link #eval(String, String)} / {@link #extract(String, String)}: the first
```
wenshao, line 232: **[Suggestion]** For array-root JSON text with a property path, `contains` consults only element 0 — existence becomes order-dependent and disagrees with `eval`/`extract` in both directions, breaking the "Mirrors" claim this javadoc makes.

**Failure scenario:** probe-verified — `JSONPathSingleName.contains`'s List branch recurses with `list.get(0)` only, so `contains("[{\"a\":1},{\"status\":2}]", "$.status")` = `false` although the key exists in element 1, and the same multiset reordered returns `true`. The other direction: `contains("[{\"status\":1}]", "$.status")` = `true` while `eval`/`extract` return `null` on the same input (`JSONPathSingleName.extract` gates on `nextIfObjectStart()`, which fails on array input). So `contains ⟺ eval != null` fails both ways on array roots. The element-0 semantics are pre-existing (equally reachable via `contains(Object, String)` with any List) and changing them would also alter the Object overload — but this diff routes every array document into that branch and invites the eval/extract comparison, so the semantics should at least be documented and pinned.

**Suggested fix:** qualify the javadoc (for array roots `contains` only inspects the first element; `eval`/`extract` do not project property paths over arrays, so "mirrors" holds for object roots) and pin the chosen semantics with tests; whether to scan all elements is a maintainer decision, not one this PR should make silently.

**[建议]** 数组根 JSON 文本 + 属性路径时，`contains` 只咨询首元素——存在性变得依赖元素顺序，且与 `eval`/`extract` 双向背离，打破了这段 javadoc 的 "Mirrors" 声明。

**失败场景：** 已探针实测——`JSONPathSingleName.contains` 的 List 分支仅以 `list.get(0)` 递归，因此 `contains("[{\"a\":1},{\"status\":2}]", "$.status")` = `false`（键明明在元素 1），同一多重集换序后返回 `true`。反向背离：`contains("[{\"status\":1}]", "$.status")` = `true`，而 `eval`/`extract` 同一输入返回 `null`（`JSONPathSingleName.extract` 以 `nextIfObjectStart()` 为门，数组输入进不去）。即数组根上 `contains ⟺ eval != null` 双向都不成立。元素 0 语义为既有实现（经 `contains(Object, String)` 对任何 List 同样可达），改动它也会同时改变 Object 重载——但本 diff 把每个数组文档都路由进该分支并主动邀请与 eval/extract 的比较，因此至少应文档化并钉住该语义。

**建议修复：** 在 javadoc 中限定语义（数组根时 `contains` 只检查首元素；`eval`/`extract` 不对数组投影属性路径，故 "mirrors" 仅对对象根成立），并用测试钉住所选语义；是否扫描全部元素应由维护者决定，不应由本 PR 隐式完成。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,26 @@ public static boolean contains(Object rootObject, String path) {
...
+     * argument is parsed as JSON, not treated as a Java bean. Without this overload,
+     * {@code contains(jsonText, path)} resolves to {@link #contains(Object, String)} and
+     * incorrectly returns false for ordinary property paths such as {@code $.status}.
+     *
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
```
wenshao, line 244: **[Suggestion]** This null guard is ungated by any test — verified by mutant probe: deleting the guard statement keeps all four `Issue7771` tests green, and flipping `return false` to `return true` survives too. The guard is behaviourally redundant today (`JSON.parse(null)` returns `null`, and `contains(Object, String)` already returns `false` for a null root), so the only observable failure mode of this branch — a wrong return value — is precisely what a test could pin.

**Failure scenario:** a future refactor that inlines the parse-and-dispatch (round 1 of this PR did exactly that) or a `JSON.parse` change rejecting null would regress `JSONPath.contains(null, "$.x")` from `false` to an exception with no failing test.

**Suggested fix:** add `assertFalse(JSONPath.contains((String) null, "$.status"));` to `Issue7771` (the cast is required so the literal binds to the `String` overload).

**[建议]** 该 null 防护没有任何测试钉住——已经突变体探针验证：删除该防护语句后 `Issue7771` 的 4 个测试仍全绿，把 `return false` 翻转为 `return true` 的突变体同样存活。该防护目前行为上冗余（`JSON.parse(null)` 返回 `null`，而 `contains(Object, String)` 对 null 根本就返回 `false`），因此本分支唯一可观察的失败模式——返回值写错——恰是测试可以钉住的。

**失败场景：** 未来把解析后分发内联的重构（本 PR 第 1 轮实现正是如此），或 `JSON.parse` 变更为拒绝 null，都会使 `JSONPath.contains(null, "$.x")` 从 `false` 静默回归为异常而无测试失败。

**建议修复：** 在 `Issue7771` 补 `assertFalse(JSONPath.contains((String) null, "$.status"));`（需强转使字面量绑定 `String` 重载）。

</details>

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void containsOnParsedObjectUnchanged() {
+        JSONObject object = JSON.parseObject("{\"status\": 200}");
+        assertTrue(JSONPath.contains(object, "$.status"));
+        assertFalse(JSONPath.contains(object, "$.missing"));
+    }
+
+    @Test
+    public void containsOnJsonNullLiteralIsFalse() {
+        assertFalse(JSONPath.contains("null", "$.status"));
```
wenshao, line 41: **[Suggestion]** The round-1 blocker (comment 3769064079) explicitly requested "null-literal/**empty-input** regression cases" — only the null-literal cases were added; there is no empty-input regression test.

**Failure scenario:** `JSONPath.contains("", "$.status")` returns `false` today only because `JSON.parse("")` returns `null` and the delegation goes through the null-safe `contains(Object, String)`. Probe-verified: round 1's inline form (`JSONPath.of(path).contains(JSON.parse(json))`) throws NPE on empty input — so if a future change inlines the parse-and-dispatch again, empty input regresses to an exception and nothing fails. The existing `"null"`-literal test would catch that revert, but `""` has no such pin.

**Suggested fix:** add `assertFalse(JSONPath.contains("", "$.status"));` here.

**[建议]** 第 1 轮 blocker（comment 3769064079）明确要求补充 JSON `null` 字面量/**空输入**两类回归用例——目前只加了 null 字面量，空输入回归测试缺失。

**失败场景：** `JSONPath.contains("", "$.status")` 现在返回 `false` 仅因 `JSON.parse("")` 返回 `null` 后走空值安全的 `contains(Object, String)`。已探针实测：第 1 轮的内联写法（`JSONPath.of(path).contains(JSON.parse(json))`）对空输入抛 NPE——未来若再次把解析后分发内联，空输入会静默回归为异常而无测试失败。现有 `"null"` 字面量测试能捕获那种回退，但 `""` 没有对应的钉子。

**建议修复：** 在此处补 `assertFalse(JSONPath.contains("", "$.status"));`。

</details>

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,32 @@ public static boolean contains(Object rootObject, String path) {
...
+            return false;
+        }
+
+        Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
+        if ("$".equals(path)) {
+            return parsed != null;
+        }
+        return contains(parsed, path);
```
wenshao, line 249: **[Critical]** R3-1: The presence contract this PR establishes (javadoc `@return true if the path exists`, tests pinning present-null members as present and absent paths as false) is broken across the containment machinery the new overload delegates to — probe-verified in both directions on this PR's compiled classes. The root cause is structural: the default `JSONPathSegment.contains` (`eval` + `value != null`) and per-branch value-based proxies are wrong in both directions, and each audit round yields further entrances — so this is one class-level finding, not an enumeration.

Evidence (all probed at HEAD, unmodified classes):
1. **Index presence conflates null value with absence.** `contains("[null]", "$[0]")` = `false` although index 0 exists — this is the round-1 open blocker (comment 3769064089, no author reply); deep site `contains("{\"a\":[null]}", "$.a[0]")` = `false` too (dispatches via `JSONPathTwoSegment`, so a `JSONPathSingleIndex`-only fix cannot close it). Controls: `contains("[1,2]", "$[0]")` = `true`, `contains("[1,2]", "$[5]")` = `false`.
2. **Name paths descending into a collection use `get(name) != null` instead of `containsKey`.** `contains("{\"a\":[{\"status\":null}]}", "$.a.status")` = `false` although the key exists, while three sibling spellings of the same element — `contains("{\"a\":{\"status\":null}}", "$.a.status")`, `contains("[{\"status\":null}]", "$.status")`, `contains("{\"a\":[{\"status\":null}]}", "$.a[0].status")` — all return `true`.
3. **Match-set shapes report `true` when nothing matches.** `contains("{}", "$.*")` = `true`, `contains("{}", "$..a")` = `true`, `contains("[1,2]", "$[5,6]")` = `true` (both indices out of bounds — while this PR's own test pins `contains("[1,2,3]", "$[5]")` = `false`), `contains("[1,2]", "$[?(@>9)]")` = `true`.
4. **Same defect on slices and multi-name selectors.** `contains("[1,2]", "$[5:6]")` = `true`, `contains("[1,2]", "$[0:0]")` = `true`, `contains("{}", "$['x','y']")` = `true` (`MultiNameSegment.eval` pads a fixed-size list with nulls, so a non-empty-list override alone cannot fix it).

Witness (flip tests): a bounds-based `JSONPathSegmentIndex.contains(Context)`, `containsKey` in `JSONPathSegmentName`'s Collection/Sequence/Object[] branches, and per-segment non-empty-match overrides flip every false-negative/false-positive above to the correct answer, while all controls stay correct (out-of-bounds stays `false`, absent keys stay `false`, `contains("[[]]", "$[0]")` stays `true`, Issue7771 mirror 22/22 PASS).

Suggested fix: one coherent semantics pass — bounds-based `contains` in `JSONPathSegmentIndex` (covers deep + root), `containsKey` in `JSONPathSegmentName`'s Collection/Sequence/Object[] branches, per-segment non-empty-match `contains` for All/CycleName/MultiIndex/Filter/Range segments plus containsKey-per-name for `MultiNameSegment`. Do **not** add a blanket empty-collection check to the default `JSONPathSegment.contains` (it would break `contains("[[]]", "$[0]")` = `true`, where the matched value itself is an empty array). Alternative: if the machinery semantics should stay, scope the javadoc/tests to what `contains` actually guarantees and answer the blocker thread on the record.

**[严重]** R3-1：本 PR 确立的存在性契约（javadoc `@return true if the path exists`、测试钉住“值为 null 的成员仍视为存在、不存在的路径返回 false”）在其委托的整条 contains 机制上被打破——已在本 PR 编译产物上双向探针实测。根因是结构性的：默认的 `JSONPathSegment.contains`（`eval` + `value != null`）与各分支基于“值非空”的代理判断在两个方向上都是错的，且每轮审计都会发现新的入口——故按一个类级问题报告，不再逐入口枚举。

证据（均在 HEAD 未改动产物上探针实测）：
1. **索引存在性把“值为 null”混同于“不存在”。** `contains("[null]", "$[0]")` = `false`，而索引 0 明明存在——这正是第 1 轮未关闭的 blocker（评论 3769064089，无回复）；深层位置 `contains("{\"a\":[null]}", "$.a[0]")` 同样 = `false`（经 `JSONPathTwoSegment` 分发，只改 `JSONPathSingleIndex` 无法覆盖）。对照：`contains("[1,2]", "$[0]")` = `true`、`contains("[1,2]", "$[5]")` = `false`。
2. **属性路径下探到集合时使用 `get(name) != null` 而非 `containsKey`。** `contains("{\"a\":[{\"status\":null}]}", "$.a.status")` = `false`（键存在），而同一元素的三种兄弟写法 `contains("{\"a\":{\"status\":null}}", "$.a.status")`、`contains("[{\"status\":null}]", "$.status")`、`contains("{\"a\":[{\"status\":null}]}", "$.a[0].status")` 均返回 `true`。
3. **匹配集类路径在零命中时返回 `true`。** `contains("{}", "$.*")` = `true`、`contains("{}", "$..a")` = `true`、`contains("[1,2]", "$[5,6]")` = `true`（双双越界——而本 PR 自己的测试钉住 `contains("[1,2,3]", "$[5]")` = `false`）、`contains("[1,2]", "$[?(@>9)]")` = `true`。
4. **切片与多键选择器同样。** `contains("[1,2]", "$[5:6]")` = `true`、`contains("[1,2]", "$[0:0]")` = `true`、`contains("{}", "$['x','y']")` = `true`（`MultiNameSegment.eval` 会填充定长 null 列表，仅“列表非空”的覆写无法修复）。

见证（flip 测试）：按边界实现的 `JSONPathSegmentIndex.contains(Context)`、`JSONPathSegmentName` 集合分支改用 `containsKey`、各匹配集段“至少命中一个”的覆写，能把上述误报/漏报全部翻正，且所有对照保持正确（越界仍 `false`、缺失键仍 `false`、`contains("[[]]", "$[0]")` 保持 `true`，Issue7771 镜像断言 22/22 通过）。

建议修复：一次统一的语义修正——`JSONPathSegmentIndex` 按边界实现 `contains`（同时覆盖深层与根）、`JSONPathSegmentName` 的 Collection/Sequence/Object[] 分支改用 `containsKey`、All/CycleName/MultiIndex/Filter/Range 段各自实现“至少命中一个”的 `contains` 且 `MultiNameSegment` 按键做 `containsKey`。**不要**在默认的 `JSONPathSegment.contains` 里加笼统的空集合判断（会破坏 `contains("[[]]", "$[0]")` = `true`——匹配到的值本身是空数组）。备选：若决定保留现机制语义，则在 javadoc/测试中限定 `contains` 实际保证的范围，并在 blocker 线程中留下书面决定。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,32 @@ public static boolean contains(Object rootObject, String path) {
...
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
+        if ("$".equals(path)) {
```
wenshao: **[Suggestion]** R3-2: The `"$"` special case (the round-2 R2-1 fix) makes the two public `contains` overloads disagree on the root path — Failure scenario: `contains("{\"a\":1}", "$")` = `true` while `contains(JSON.parse("{\"a\":1}"), "$")` = `false` — identical logical document, opposite answers depending on which overload binds (probe-verified); a caller refactoring between text and pre-parsed documents gets a silent flip, and `eval(obj, "$")` returns the object, so the `contains ⟺ eval != null` parity the Issue7771 tests build on breaks for the object overload. (`RootPath.contains` at JSONPath.java:1559 returns `false` unconditionally; the new javadoc claim "The root path `$` is present for every non-null document" holds only for this overload.)

Fix options:
```java
// A: align both overloads — change RootPath.contains to:
return object != null;
// then this "$" special case can be dropped entirely.
// B: keep the divergence — document on contains(Object, String)
// that "$" returns false there.
```

**[建议]** R3-2：`"$"` 特判（第 2 轮 R2-1 的修复方式）使两个公开的 `contains` 重载在根路径上给出相反答案——失败场景：`contains("{\"a\":1}", "$")` = `true`，而 `contains(JSON.parse("{\"a\":1}"), "$")` = `false`；同一逻辑文档仅因绑定重载不同而答案相反（已探针实测），调用方在文本与已解析对象之间重构时会静默翻转。且 `eval(obj, "$")` 返回对象本身，Issue7771 测试所依赖的 `contains ⟺ eval != null` 等价关系在 Object 重载上被打破。（`RootPath.contains`（JSONPath.java:1559）无条件返回 `false`；新 javadoc “根路径 `$` 在每个非 null 文档中都存在”仅对本文本重载成立。）修复选项：A. 将 `RootPath.contains` 改为 `return object != null;`（两个重载随之对齐，本特判可整体删除）；B. 若有意保留分歧，在 `contains(Object, String)` 上注明 `"$"` 恒返回 `false`。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,32 @@ public static boolean contains(Object rootObject, String path) {
...
+     * @param json the JSON text to check
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
```
wenshao, line 248: **[Suggestion]** R3-3: The overload's error path — malformed JSON throwing `JSONException` out of `JSON.parse` — is pinned by no test — Concrete cost: `contains("{invalid json", "$.status")` throws at HEAD (probe-verified; whitespace-only input throws too, even for path `"$"`, since parsing precedes the `"$"` shortcut). The throw-vs-false contract lives only in prose (open suggestion R2-5 discusses it), so a future commit wrapping the parse in try/catch — or the opposite contract change — flips public behavior with zero failing tests. Sweep: 41 `JSONPath.contains` test call sites, 0 cover malformed input or assert a throw.

Suggested fix — pin the current contract in `Issue7771`:
```java
assertThrows(JSONException.class, () -> JSONPath.contains("{invalid json", "$.status"));
```
(or, if the intended contract is return-`false`-on-parse-failure, implement and pin that instead)

**[建议]** R3-3：本重载的错误路径——非法 JSON 使 `JSON.parse` 抛出 `JSONException`——没有任何测试钉住。具体代价：`contains("{invalid json", "$.status")` 在 HEAD 上抛异常（已探针实测；纯空白输入同样抛，即便路径为 `"$"`，因为解析先于 `"$"` 短路）。抛异常还是返回 `false` 的契约只存在于文字描述中（未决建议 R2-5 已讨论该问题），未来任何给解析包 try/catch（或反向的契约变更）的提交都会在没有测试失败的情况下翻转公开行为。扫描：41 处 `JSONPath.contains` 测试调用，0 处覆盖非法输入或断言抛异常。建议修复——在 `Issue7771` 中钉住当前契约（如上代码）；若预期契约是解析失败返回 `false`，则实现并钉住该语义。

</details>

wenshao (changes requested):
3 Suggestion-level finding(s) this review confirmed are already reported on this PR and are not repeated:

- javadoc exception/trailing-content divergences (R2-5) — already reported ([comment 3780275045](https://github.com/alibaba/fastjson2/pull/7778#discussion_r3780275045))
- array-root first-element semantics vs eval/extract (R2-6) — already reported ([comment 3780275047](https://github.com/alibaba/fastjson2/pull/7778#discussion_r3780275047))
- full-document materialization vs streaming — already reported ([comment 3769064093](https://github.com/alibaba/fastjson2/pull/7778#discussion_r3769064093))

<details>
<summary>中文说明</summary>

本轮确认的 3 条建议级发现已在 PR 上报告过，不再重复发布（列表见上方英文部分）。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
kalayciburak: yep, contains is presence now. present null counts, empty match sets dont. $ is the same on both overloads.

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -1531,7 +1554,7 @@ public boolean isRef() {
 
         @Override
         public boolean contains(Object object) {
-            return false;
+            return object != null;
```
wenshao, line 1557: **[Critical]** The `RootPath.contains` change (`return false` → `return object != null` — the round-3 R3-2 fix) breaks the existing, unmodified test `JSONPathTypedMultiTest.test7`, so `mvn test` is red on this PR. Full core suite at HEAD: 7984 tests, 1 failure; A/B measured against merge base `af56f06ee`: the base passes `JSONPathTypedMultiTest` 10/10 while this head fails test7 — net-new by measurement, and the CI matrix is failing for the same reason (29 of 30 checks red).

`JSONPathTypedMulti.contains` returns true if ANY sub-path is contained; for paths `{"$", "$.values[0].name", "$.values[0].date"}` against a non-null JSONObject with no `values` member, the `"$"` sub-path now reports contained, flipping the asserted answer to `true` (`expected: <false> but was: <true>`).

Witness:
```
mvn -pl core test → Tests run: 7984, Failures: 1 (JSONPathTypedMultiTest.test7)
base af56f06ee, -Dtest=JSONPathTypedMultiTest → 10/10 PASS; PR head → test7 FAILS
```

Reconcile the root-path change with the existing multi-path contract before merge: either update `test7` to the new intended semantics (if "$ is present for every non-null document" is meant to flip multi-path `contains`), or scope the change so `JSONPathTypedMulti.contains` keeps its behavior (e.g. answer `"$"` in the text overload only, without changing `RootPath.contains`).

**[严重]** `RootPath.contains` 的改动（`return false` → `return object != null`，即第 3 轮 R3-2 的修复）打破了既有且未被本 PR 修改的测试 `JSONPathTypedMultiTest.test7`，本 PR 上 `mvn test` 为红。HEAD 上 core 全量：7984 个测试、1 个失败；与 merge base `af56f06ee` 的 A/B 实测：base 上 `JSONPathTypedMultiTest` 10/10 通过，本 head 失败——按测量为新增失败，CI 矩阵也因此 29/30 红。

`JSONPathTypedMulti.contains` 在任一子路径存在时返回 true；对路径集 `{"$", "$.values[0].name", "$.values[0].date"}` 与不含 `values` 成员的非 null JSONObject，`"$"` 子路径现在判定存在，使断言的 `false` 翻转为 `true`。

**见证：** 见上方英文部分的测试输出。

**建议：** 合入前先协调根路径改动与既有多路径契约：或按新语义更新 `test7`（若“根路径在每个非 null 文档中都存在”本就应翻转多路径 `contains`），或限定改动范围（如仅在文本重载中特判 `"$"`，不改 `RootPath.contains`），使 `JSONPathTypedMulti.contains` 行为不变。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPath.java
```diff
@@ -226,6 +226,29 @@ public static boolean contains(Object rootObject, String path) {
...
+     * @param path the JSONPath expression
+     * @return true if the path exists, false otherwise
+     */
+    public static boolean contains(String json, String path) {
+        if (json == null) {
+            return false;
+        }
+
+        Object parsed = JSON.parse(json, JSONReader.Feature.DisableReferenceDetect);
+        return contains(parsed, path);
```
wenshao, line 249: **[Critical]** R3-1: The presence contract this PR establishes (javadoc `@return true if the path exists`, tests pinning present-null as present and empty match sets as false) is **still** broken across the containment machinery the new overload delegates to — probe-verified again at this head, and this round's audit found the new entrances **in the round-3 fix code itself**, so the class finding stands. Eleven verified entrances:

1. `JSONPathSegmentIndex.contains`'s Map branch drops `eval(Map)`'s positional/Long-key fallbacks: `contains("{\"a\":1,\"b\":2}", "$[1]")` = `false` while `eval` returns `2` (`JSONObject` extends `LinkedHashMap`); single-entry maps and `Map<Long,?>` keys likewise. Pre-PR the inherited default returned `true` — a regression on ordinary parsed JSON, directly reachable via this overload.
2. `MultiNameSegment.contains`'s non-Map fallback: `contains("[{\"z\":1}]", "$['x','y']")` = `true` although no element has `x`/`y` (Collection pass-through; bean roots answer true via the null-padded eval result).
3. `JSONPathSegmentIndex.contains` bounds-checks EVERY `Collection`: `contains(HashSet{1,2,3}, "$[1]")` = `true` while `eval` throws `JSONException("jsonpath not support operate")`.
4. `ValuesSegment`/`KeysSegment`/`EntrySetSegment` were not overridden: `contains("{}", "$.values()")` = `true` while `contains("{}", "$.*")` = `false` — same value set, opposite answers.
5. `RangeIndexSegment.contains` throws `JSONException("TODO")` for non-List/non-array roots (`contains("{}", "$[0:1]")`) — eval throws identically, the weakest entrance, an inherited defect re-exposed on the new public API.
6. Chained MultiName padding: `contains("{}", "$['x','y'][0]")` = `true` while `eval` = `null` — A/B: base `false` → PR `true` regression (`JSONPathTwoSegment.contains` chains via `first.eval` and the new bounds check trusts the padded `[null,null]`).
7. `NameExistsFilter.eval` (value semantics: `containsKey && get != null`) vs `.apply` (presence: `containsKey`) split, newly exposed by the `hasMatch` override: on `[{"a":null,"b":1}]`, `contains("$[?(@.a)]")` = `false` but the compound `contains("$[?(@.b==1 && exists(@.a))]")` = `true` (base: `true`/`true`).
8. `GroupFilter` extends `JSONPathSegment` directly and bypasses the new `JSONPathFilter.contains` override: `contains("[]", "$[?(@.a>9 && @.b<5)]")` = `true` with `eval` = `[]` — the empty-match contract this PR's own test pins for simple filters is broken for compound predicates within the same PR (flip-verified: the identical override on `GroupFilter` fixes it).
9. `JSONPathSegmentName.contains`'s Sequence branch never descends into Collection items: `contains("{\"a\":[{\"x\":1}],\"b\":[{\"x\":2}]}", "$.*.x")` = `false` while `eval`/`extract` return `[1,2]` (flip-verified; the branch is reworked by this diff and the text overload makes it user-reachable).
10. Flattening hazard: `hasMatch` runs on the FLATTENED eval result — `contains("[[],[1]]", "$[0,5]")` = `false` although index 0 exists (`contains("$[0]")` = `true` on the same document; `extract` returns `[[]]`), a monotonicity violation (`$[0]` ⊂ `$[0,5]` yet true→false when a nonexistent index is added).
11. Same mechanism: `contains("{\"a\":[]}", "$[*]")` = `false` while the twin spelling `contains("{\"a\":[]}", "$.*")` = `true`; both flip pre-PR `true`→`false`.

Witness (all probed at HEAD, unmodified classes):
```
contains("{\"a\":1,\"b\":2}","$[1]")=false | eval=2
contains("[{\"z\":1}]","$['x','y']")=true   (no x/y in any element)
contains(HashSet{1,2,3},"$[1]")=true       | eval throws
contains("{}","$.values()")=true           | contains("{}","$.*")=false
contains("{}","$['x','y'][0]")=true        | eval=null  (BASE: false)
contains("[]","$[?(@.a>9 && @.b<5)]")=true | eval=[]
contains("{\"a\":[{\"x\":1}],\"b\":[{\"x\":2}]}","$.*.x")=false | eval=[1,2]
contains("[[],[1]]","$[0,5]")=false        | contains("$[0]")=true | extract=[[]]
contains("{\"a\":[]}","$[*]")=false          | contains("$.*")=true
flip: reverting the SegmentIndex Map branch to eval-delegation restores true
```

Four rounds have now each found new entrances in the previous round's fix — the surface regenerates entrances faster than per-segment patches close them. Please do one coherent semantics pass rather than entrance-by-entrance patching: mirror `eval(Map)`'s lookup chain in `SegmentIndex` (positional + Long-key); give `MultiName` a real membership check for non-Map roots (and stop padding absent names when chained); restrict Collection bounds to eval-supported types; override `Values`/`Keys`/`EntrySet` and `GroupFilter` with the `hasMatch` semantics; make `SegmentName`'s Sequence branch recurse like `SegmentIndex`'s; and for the flattening hazard distinguish "matched an empty collection" from "matched nothing" (count positional/member matches before flattening). Alternatively, lift `hasMatch` into the base default deliberately — that closes (4) and removes the four copy-pasted override bodies at once. Pin each fix with true-side tests (see the coverage comment on `Issue7771`). If the machinery semantics should instead stay as-is, scope the javadoc/tests to what `contains` actually guarantees and say so on the record.

**[严重]** R3-1：本 PR 确立的存在性契约（javadoc `@return true if the path exists`、测试钉住“值为 null 仍视为存在、空匹配集为 false”）在其委托的整条 contains 机制上**仍然**被打破——已在本 head 上再次探针实测，且本轮审计在**第 3 轮修复代码本身**中发现了新入口，故该类级问题继续成立。共 11 个已验证入口：

1. `JSONPathSegmentIndex.contains` 的 Map 分支丢掉了 `eval(Map)` 的位置回退与 Long 键匹配：`contains("{\"a\":1,\"b\":2}", "$[1]")` = `false` 而 `eval` 返回 `2`（`JSONObject` 继承 `LinkedHashMap`）；单元素 Map 与 `Map<Long,?>` 同样。PR 前继承的默认实现返回 `true`——对普通已解析 JSON 构成回归，且可经本重载直达。
2. `MultiNameSegment.contains` 的非 Map 回退：`contains("[{\"z\":1}]", "$['x','y']")` = `true`，而任一元素都不含 `x`/`y`（Collection 直通；bean 根因 null 填充的 eval 结果同样为 true）。
3. `JSONPathSegmentIndex.contains` 对**所有** `Collection` 做边界判断：`contains(HashSet{1,2,3}, "$[1]")` = `true` 而 `eval` 抛 `JSONException`。
4. `ValuesSegment`/`KeysSegment`/`EntrySetSegment` 未覆写：`contains("{}", "$.values()")` = `true` 而 `contains("{}", "$.*")` = `false`——同一值集，答案相反。
5. `RangeIndexSegment.contains` 对非 List/非数组根抛 `JSONException("TODO")`（`contains("{}", "$[0:1]")`）——eval 抛同样异常，是最弱的入口，属既有缺陷在新公开 API 上的再暴露。
6. 链式 MultiName 填充：`contains("{}", "$['x','y'][0]")` = `true` 而 `eval` = `null`——A/B：base `false` → PR `true` 回归（`JSONPathTwoSegment.contains` 经 `first.eval` 链式，新的边界判断信任了填充出的 `[null,null]`）。
7. `NameExistsFilter.eval`（值语义：`containsKey && get != null`）与 `.apply`（存在语义：`containsKey`）的分裂被新的 `hasMatch` 覆写暴露：对 `[{"a":null,"b":1}]`，`contains("$[?(@.a)]")` = `false` 而复合 `contains("$[?(@.b==1 && exists(@.a))]")` = `true`（base 上为 `true`/`true`）。
8. `GroupFilter` 直接继承 `JSONPathSegment`，绕过了新的 `JSONPathFilter.contains` 覆写：`contains("[]", "$[?(@.a>9 && @.b<5)]")` = `true` 而 `eval` = `[]`——本 PR 自己的测试为简单过滤器钉住的空匹配契约，对复合谓词在同一 PR 内被打破（flip 验证：给 `GroupFilter` 加同样覆写即修复）。
9. `JSONPathSegmentName.contains` 的 Sequence 分支从不下探 Collection 元素：`contains("{\"a\":[{\"x\":1}],\"b\":[{\"x\":2}]}", "$.*.x")` = `false` 而 `eval`/`extract` 返回 `[1,2]`（flip 验证；该分支被本 diff 改写，且文本重载使其对用户可达）。
10. 扁平化陷阱：`hasMatch` 作用在**扁平化后**的 eval 结果上——`contains("[[],[1]]", "$[0,5]")` = `false`，而索引 0 明明存在（同一文档上 `contains("$[0]")` = `true`；`extract` 返回 `[[]]`），违反单调性（`$[0]` ⊂ `$[0,5]`，加入一个不存在的索引反而 true→false）。
11. 同一机制：`contains("{\"a\":[]}", "$[*]")` = `false` 而孪生写法 `contains("{\"a\":[]}", "$.*")` = `true`；两者均由 PR 前的 `true` 翻转为 `false`。

**见证：** 全部在 HEAD 未改动产物上探针实测（见上方英文部分输出）；flip 测试：将 SegmentIndex Map 分支还原为 eval 委托后恢复 true。

**建议：** 四轮审计每轮都在上一轮的修复里发现新入口——该表面产生入口的速度快于逐段补丁关闭它们的速度。请做一次统一的语义修正，而不是逐入口打补丁：在 `SegmentIndex` 中对齐 `eval(Map)` 的查找链（位置 + Long 键）；为 `MultiName` 的非 Map 根实现真正的成员判断（并在链式场景停止为缺失名填充）；把 Collection 边界判断限定到 eval 支持的类型；为 `Values`/`Keys`/`EntrySet` 与 `GroupFilter` 补上 `hasMatch` 语义；让 `SegmentName` 的 Sequence 分支像 `SegmentIndex` 一样递归；对扁平化陷阱区分“命中了空集合”与“没有命中”（在扁平化之前统计位置/成员命中）。或者，有意地把 `hasMatch` 提升到基类默认实现——可一并关闭 (4) 并消除四处复制的覆写体。每项修复请补 true 侧测试（见 `Issue7771` 上的覆盖评论）。若决定保留现机制语义，则在 javadoc/测试中限定 `contains` 实际保证的范围，并留下书面决定。

</details>

core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7771.java
```diff
@@ -0,0 +1,97 @@
...
+    @Test
+    public void containsTreatsPresentNullAsPresent() {
+        assertTrue(JSONPath.contains("{\"a\":[{\"status\":null}]}", "$.a.status"));
+        assertTrue(JSONPath.contains("{\"a\":{\"status\":null}}", "$.a.status"));
+        assertTrue(JSONPath.contains("[{\"status\":null}]", "$.status"));
+        assertTrue(JSONPath.contains("{\"a\":[{\"status\":null}]}", "$.a[0].status"));
+    }
+
+    @Test
+    public void containsOnEmptyMatchSetIsFalse() {
```
wenshao, line 77: **[Suggestion]** The new containment machinery is pinned one-sidedly: five new overrides — `RangeIndexSegment.contains`, `MultiIndexSegment.contains`, `MultiNameSegment.contains` (incl. its new Map `containsKey` fast path), `AllSegment.contains`, and `JSONPathFilter.contains` — are exercised ONLY by the `assertFalse` cases in this test. A one-line `return false;` body mutation in any of them compiles and leaves every test green (I grepped every `JSONPath.contains(` call site in core and fastjson1-compatible — no true-side assertion exists anywhere). Also untested: `JSONPathSegmentIndex.contains`' negative-index normalization (`$[-1]`), Map-key branch, scalar fallback `index == 0`, Sequence recursion, and `CycleNameSegment`'s `$..*` (MapRecursive) branch.

The cost is concrete: each surviving mutant is exactly a contains/eval divergence — the defect class this PR exists to fix, and the R3-1 comment demonstrates it live in this very code.

```suggestion
    @Test
    public void containsOnEmptyMatchSetIsFalse() {
        assertFalse(JSONPath.contains("{}", "$.*"));
        assertFalse(JSONPath.contains("{}", "$..a"));
        assertFalse(JSONPath.contains("[1,2]", "$[5,6]"));
        assertFalse(JSONPath.contains("[1,2]", "$[?(@>9)]"));
        assertFalse(JSONPath.contains("[1,2]", "$[5:6]"));
        assertFalse(JSONPath.contains("[1,2]", "$[0:0]"));
        assertFalse(JSONPath.contains("{}", "$['x','y']"));
    }

    @Test
    public void containsOnNonEmptyMatchSetIsTrue() {
        assertTrue(JSONPath.contains("[1,2]", "$[0:2]"));
        assertTrue(JSONPath.contains("[1,2]", "$[0,5]"));
        assertTrue(JSONPath.contains("{\"x\":1}", "$['x','y']"));
        assertTrue(JSONPath.contains("{\"a\":1}", "$.*"));
        assertTrue(JSONPath.contains("[1,2]", "$[?(@>0)]"));
        assertTrue(JSONPath.contains("{\"a\":1}", "$..*"));
        assertTrue(JSONPath.contains("[1,2,3]", "$[-1]"));
        assertFalse(JSONPath.contains("[1,2,3]", "$[-5]"));
        assertTrue(JSONPath.contains("{\"0\":\"a\"}", "$[0]"));
        assertTrue(JSONPath.contains("123", "$[0]"));
        assertFalse(JSONPath.contains("123", "$[1]"));
    }
```

**[建议]** 新的 contains 机制只被单侧钉住：`RangeIndexSegment.contains`、`MultiIndexSegment.contains`、`MultiNameSegment.contains`（含新增的 Map `containsKey` 快速路径）、`AllSegment.contains`、`JSONPathFilter.contains` 五个新覆写仅由本测试中的 `assertFalse` 用例覆盖。任一覆写体改成一行 `return false;` 的变异都能编译通过且全套测试仍绿（已 grep core 与 fastjson1-compatible 的全部 `JSONPath.contains(` 调用点，无任何 true 侧断言）。此外未覆盖：`JSONPathSegmentIndex.contains` 的负索引归一化（`$[-1]`）、Map 键分支、标量回退 `index == 0`、Sequence 递归，以及 `CycleNameSegment` 的 `$..*`（MapRecursive）分支。

代价是具体的：每个存活的变异恰是一个 contains/eval 分歧——正是本 PR 要修复的缺陷类，R3-1 评论已在这些代码中实证。

**建议修复：** 见上方 suggestion 代码块——在现有 false 断言旁补 true 侧镜像测试及 SegmentIndex 分支测试。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPathSegment.java
```diff
@@ -1390,6 +1437,22 @@ public void eval(JSONPath.Context context) {
...
+                    ? context.root
+                    : context.parent.value;
+            List values = new JSONArray();
+            Consumer action;
+            if (shouldRecursive()) {
+                action = new MapRecursive(context, values, 0);
+            } else {
+                action = new MapLoop(context, values);
+            }
+            action.accept(object);
```
wenshao, line 1452: **[Suggestion]** Three of the new `contains` overrides answer a boolean by materializing the full result set first. `CycleNameSegment.contains` (here) re-runs `MapLoop`/`MapRecursive` to collect EVERY match into a JSONArray before `isEmpty()` — no early exit; for `$..*`, `MapRecursive.recursive` does `values.addAll(collection)` at every level, so the list itself grows O(document). `AllSegment.contains` (~line 1102) allocates an N-element JSONArray and copies every value just to test non-emptiness, where `!map.isEmpty()` suffices. `RangeIndexSegment.contains` (~line 283) iterates ALL list elements even for a tiny or out-of-range slice (`$[5:6]` on a 10M-element list = 10M iterations), where begin/end-vs-size arithmetic is O(1).

This is a missed short-circuit opportunity in newly added code rather than a regression (the pre-PR default paid the same eval cost), but an existence probe becoming O(document) work and GC pressure per call is exactly the wrong cost model for an API positioned for per-request existence checks.

```suggestion
            action.accept(object);
            return !values.isEmpty();
```
(the shape to move away from: give `CycleNameSegment` a short-circuiting visitor — a `found` flag the loop consumers check on entry, returning at first match with no collection; and direct size/bounds checks for `AllSegment` and `RangeIndexSegment`, the same O(1) treatment this PR already gave `JSONPathSegmentIndex`)

**[建议]** 三个新的 `contains` 覆写以“先物化完整结果集”的方式回答布尔问题。`CycleNameSegment.contains`（此处）复用 `MapLoop`/`MapRecursive` 把**所有**匹配收集进 JSONArray 后才 `isEmpty()`——无提前退出；对 `$..*`，`MapRecursive.recursive` 每层执行 `values.addAll(collection)`，列表本身即 O(document) 增长。`AllSegment.contains`（约 1102 行）为判断非空分配 N 元素 JSONArray 并逐个拷贝（`!map.isEmpty()` 即可）。`RangeIndexSegment.contains`（约 283 行）即便对极小/越界切片也遍历全部元素（10 百万元素列表上的 `$[5:6]` = 一千百万次迭代），而 begin/end 与 size 的比较是 O(1)。

这不是回归（PR 前的默认实现同样付出 eval 代价），而是新增代码中错失的短路机会：存在性探测变成每次调用 O(document) 的工作量与 GC 压力，与“按请求做存在性检查”的 API 定位不符。

**建议修复：** 为 `CycleNameSegment` 提供短路访问器（循环消费者入口处检查 `found` 标志，首个匹配即返回、不收集）；为 `AllSegment`、`RangeIndexSegment` 改用直接的大小/边界判断——即本 PR 已在 `JSONPathSegmentIndex` 上给出的 O(1) 处理。

</details>

wenshao (changes requested):
2 Suggestion-level finding(s) this review confirmed are already reported on this PR and are not repeated:

- array-root first-element-only name descent (contains true where eval null; order-dependent) — already reported ([comment 3780275047](https://github.com/alibaba/fastjson2/pull/7778#discussion_r3780275047))
- full-document materialization vs streaming eval/extract — already reported ([comment 3769064093](https://github.com/alibaba/fastjson2/pull/7778#discussion_r3769064093))

Not explored to full depth (tool budget reached): `"agent 1b"`: `none — no check was cut short by the tool budget.`.

<details>
<summary>中文说明</summary>

本轮确认的 2 条建议级发现已在 PR 上报告过，不再重复发布（列表见上方英文部分）。

未探索到全部深度（达到工具调用预算）：`"agent 1b"`：`none — no check was cut short by the tool budget.`。

</details>

core/src/test/java/com/alibaba/fastjson2/JSONPathTypedMultiTest.java
```diff
@@ -200,6 +200,7 @@ public void test7() {
                 new String[]{"$", "$.values[0].name", "$.values[0].date"},
                 new Type[]{JSONObject.class, String.class, Date.class}
         );
+        assertTrue(JSONPath.contains(object, "$"));
```
wenshao, line 203: **[Suggestion]** The new RootPath skip is pinned only in the negative direction: test7 pairs "$" with all-absent subpaths, and the assertion added here routes through the static single-path overload (`JSONPath.of("$")` → `RootPath`) and never touches `JSONPathTypedMulti`. A one-line mutant `continue;` → `return false;` in `JSONPathTypedMulti.contains` therefore survives the whole suite: the loop hits the RootPath at index 0 and returns false, test7's remaining paths are all absent so the assertFalse stays green, and no other test builds a multi-path containing "$" (repo-wide sweep). Concretely, `JSONPath.of(new String[]{"$", "$.id"}, new Type[]{Object.class, Object.class}).contains(object)` must be true (`$.id` exists; "$" is legitimately skipped), but under the mutant it silently returns false. Verified with a full mutation cycle in a scratch tree: pristine returns true for the mixed path; the mutant flips it to false while `JSONPathTypedMultiTest` stays 10/10 green (also `JSONPathTypedMultiTest3` 14/14, `JSONPath_10_contains` 1/1, `JSONPathTypedTest` 2/2 green under the mutant); the proposed assertion fails under the mutant (`expected: <true> but was: <false>`) and passes on pristine code. Add the mixed positive case:

```suggestion
        assertTrue(JSONPath.contains(object, "$"));
        JSONPath jsonPath2 = JSONPath.of(
                new String[]{"$", "$.id"},
                new Type[]{JSONObject.class, Long.class}
        );
        assertTrue(jsonPath2.contains(object));
```

After applying, please confirm the pin bites: replacing `continue;` with `return false;` in `JSONPathTypedMulti.contains` and re-running `JSONPathTypedMultiTest` must turn the added `assertTrue(jsonPath2.contains(object))` red.

**问题：** 新增的 RootPath 跳过逻辑只被负向钉住：test7 中 "$" 与全部缺失的子路径组合，而此处新增的断言走的是静态单路径重载（`JSONPath.of("$")` → `RootPath`），根本不经过 `JSONPathTypedMulti`。因此 `JSONPathTypedMulti.contains` 中的单行变异 `continue;` → `return false;` 可以逃过整个测试套件：循环在索引 0 处遇到 RootPath 直接返回 false，test7 其余路径恰好都缺失，assertFalse 仍为绿；全仓库范围内也没有其他测试构造包含 "$" 的多路径。具体而言，`JSONPath.of(new String[]{"$", "$.id"}, new Type[]{Object.class, Object.class}).contains(object)` 应为 true（`$.id` 存在；"$" 被合理跳过），变异体却会悄悄返回 false。已在临时树中完成完整变异循环验证：原始代码下混合路径返回 true；变异体下翻转为 false 且 `JSONPathTypedMultiTest` 仍 10/10 全绿（`JSONPathTypedMultiTest3` 14/14、`JSONPath_10_contains` 1/1、`JSONPathTypedTest` 2/2 在变异体下同样全绿）；建议新增的断言在变异体下失败（`expected: <true> but was: <false>`），在原始代码下通过。请补充混合正例（见上方 suggestion 代码块）。

**修复验证：** 应用修复后请确认钉住效果：将 `JSONPathTypedMulti.contains` 中的 `continue;` 替换为 `return false;` 并重跑 `JSONPathTypedMultiTest`，新增的 `assertTrue(jsonPath2.contains(object))` 必须变红。

</details>

wenshao (changes requested):
**[Critical]** R3-1: still stands at head 292718f6c — the presence contract this PR establishes (javadoc `@return true if the path exists`, tests pinning present-null as present and empty match sets as false) remains broken across the containment machinery the new `contains(String, String)` overload delegates to. This round's delta (6a2a74eec..292718f6c) touches only `JSONPathTypedMulti.java` (+3) and `JSONPathTypedMultiTest.java` (+1); every machinery file the round-4 entrances cite (`JSONPathSegment`, `JSONPathSegmentIndex`, `JSONPathSegmentName`, `JSONPathFilter`, `JSONPath`) is byte-identical to the round-4 head where all 11 entrances were probe-verified (comment 3818931823). Representative entrances re-probed at THIS head and still firing: `contains("{\"a\":1,\"b\":2}", "$[1]")` = false (SegmentIndex positional lookup dropped; eval over the parsed object returns 2); `contains("{}", "$.values()")` = true while `contains("{}", "$.*")` = false (same value set, opposite answers); `contains("[]", "$[?(@.a>9 && @.b<5)]")` = true (empty match set reports present — GroupFilter bypasses the JSONPathFilter override); `contains("{\"a\":[{\"x\":1}],\"b\":[{\"x\":2}]}", "$.*.x")` = false (Sequence branch never descends into Collection items; eval returns [1,2]). This anchor sits outside this round's incremental hunks, so the finding rides in the body; the round-4 comment remains the full statement of the class finding and of the coherent-semantics-pass remedy it proposes. <details><summary>中文说明</summary> **[严重]** R3-1：在 head 292718f6c 上仍然成立——本 PR 确立的存在性契约（javadoc `@return true if the path exists`、测试钉住“值为 null 仍视为存在、空匹配集为 false”）在其委托的整条 contains 机制上仍然被打破。本轮增量（6a2a74eec..292718f6c）仅改动 `JSONPathTypedMulti.java`（+3）与 `JSONPathTypedMultiTest.java`（+1）；第 4 轮各入口引用的机制文件（`JSONPathSegment`、`JSONPathSegmentIndex`、`JSONPathSegmentName`、`JSONPathFilter`、`JSONPath`）与第 4 轮 head 逐字节一致，而当时全部 11 个入口均已探针实测（评论 3818931823）。在本 head 上重新探针实测的代表性入口仍在触发：`contains("{\"a\":1,\"b\":2}", "$[1]")` = false（SegmentIndex 丢失位置查找；对解析对象求值返回 2）；`contains("{}", "$.values()")` = true 而 `contains("{}", "$.*")` = false（同一值集，答案相反）；`contains("[]", "$[?(@.a>9 && @.b<5)]")` = true（空匹配集判为存在——GroupFilter 绕过 JSONPathFilter 覆写）；`contains("{\"a\":[{\"x\":1}],\"b\":[{\"x\":2}]}", "$.*.x")` = false（Sequence 分支从不下探 Collection 元素；eval 返回 [1,2]）。本锚点位于本轮增量 hunk 之外，故该发现置于综述正文；第 4 轮评论仍是该类级问题及其“统一语义修正”建议的完整陈述。 </details>

PR #7788, opened 2026-08-18
fix: avoid D/F suffix for non-finite numbers

annoymous3:
## Description

When `JSONWriter.Feature.WriteClassName` is enabled, non-finite
Double/Float values such as `NaN` and `Infinity` are serialized as
`null`, but `ObjectWriterImplDouble` and `ObjectWriterImplFloat`
still append the `D`/`F` suffix.

This can produce invalid JSON such as:

```json
{"value":nullD}
````

or:

```json
{"value":nullF}
```

## Fix

Only append the `D`/`F` type suffix when the value is finite.

Finite Double/Float serialization behavior remains unchanged.

## Tests

Added regression tests covering:

* `Double.NaN`
* `Double.POSITIVE_INFINITY`
* `Float.NaN`
* `Float.NEGATIVE_INFINITY`
* finite Double/Float values with `WriteClassName`

All targeted tests pass.

### DIFF
--- a/core/src/test/java/com/alibaba/fastjson2/issues_3400/Issue3449.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_3400/Issue3449.java
@@ -5,6 +5,7 @@ import com.alibaba.fastjson2.JSONWriter;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @Tag("regression")
@@ -44,5 +45,26 @@ public class Issue3449 {
 
         assertEquals("1.0D", JSON.toJSONString(1D, JSONWriter.Feature.WriteClassName));
         assertEquals("1.0F", JSON.toJSONString(1F, JSONWriter.Feature.WriteClassName));
+    }
+
+    @Test
+    public void testWriteClassNameInObjectGraph() {
+        Bean bean = new Bean();
+        bean.d = Double.NaN;
+        bean.f = Float.POSITIVE_INFINITY;
+        bean.finite = 1.5D;
+
+        String json = JSON.toJSONString(bean, JSONWriter.Feature.WriteClassName);
+        assertEquals(
+                "{\"@type\":\"com.alibaba.fastjson2.issues_3400.Issue3449$Bean\",\"d\":null,\"f\":null,\"finite\":1.5D}",
+                json
+        );
+        assertDoesNotThrow(() -> JSON.parse(json));
+    }
+
+    public static class Bean {
+        public Object d;
+        public Object f;
+        public Object finite;
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/alibaba/fastjson2/issues_3400/Issue3449.java
```diff
@@ -33,4 +34,15 @@ public void testFloatNegativeNan() {
         assertEquals("null", JSON.toJSONString(newNan));
         assertEquals("null", new String(JSON.toJSONBytes(newNan)));
     }
+
+    @Test
+    public void testWriteClassName() {
+        assertEquals("null", JSON.toJSONString(Double.NaN, JSONWriter.Feature.WriteClassName));
+        assertEquals("null", JSON.toJSONString(Double.POSITIVE_INFINITY, JSONWriter.Feature.WriteClassName));
+        assertEquals("null", JSON.toJSONString(Float.NaN, JSONWriter.Feature.WriteClassName));
+        assertEquals("null", JSON.toJSONString(Float.NEGATIVE_INFINITY, JSONWriter.Feature.WriteClassName));
```
wenshao, line 44: **[Suggestion]** The new test only pins bare root-level values, but the production shape of this bug is a non-finite value inside an object graph — a bean field or map entry of declared type `Object`/`Number`, serialized with `WriteClassName`. I verified by probe that the fix does cover that path (it routes through the same single predicate; `fieldType = Object.class` still satisfies `fieldType != Double.class`), and a sweep of `core/src/test` found no existing test pinning this shape, so a future change that reroutes field/map value writes or re-adds the suffix on that path would ship with the suite green. A pre-fix probe arm shows the untested path genuinely produced the invalid output:

```
PR arm     : {"@type":"Probe7788$Bean","d":null,"f":null,"finite":1.5D}
pre-fix    : {"@type":"Probe7788$Bean","d":nullD,"f":nullF,"finite":1.5D}
round-trip : JSONException: offset 35, character D
```

Consider adding one graph-level assertion, e.g. a bean with an `Object` field or a `Map<String, Object>` entry holding `Double.NaN` and `Float.POSITIVE_INFINITY`, serialized with `WriteClassName`, asserting the entries render as `null` with no `D`/`F` suffix and the result round-trips through `JSON.parse`.

wenshao (commented):

annoymous3:
Thanks for the suggestion. Addressed in [`2b087ff6`](https://github.com/alibaba/fastjson2/commit/2b087ff6252ea99051615e29591fe34abea36e69).

I added a graph-level regression test using a bean with `Object`-typed fields. It verifies that `Double.NaN` and `Float.POSITIVE_INFINITY` are serialized as `null` without `D`/`F` suffixes, while a finite `Double` still retains the `D` suffix. The generated output is also parsed with `JSON.parse` to verify that it is valid.

The targeted test passes.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running. Reviewed.

annoymous3:
Hi maintainers,

Just a gentle ping on this PR.

I submitted my signed Individual Contributor License Agreement (CLA) to opensource@alibaba-inc.com on August 26th (almost a month ago), but I haven't received a confirmation yet.

Could someone please check if it was received? I'm hoping we can get this sorted out so the PR can move forward. Let me know if there's anything else I need to do from my side.

Thanks for your time!

PR #7791, opened 2026-08-18
fix: handle truncated escape in JSONReaderUTF16

annoymous3:
### 这个 PR 做了什么 / 为什么需要这个修改？

当 JSON 字符串以被截断的反斜杠转义结尾时，`JSONReaderUTF16.readString()` 会读取字符缓冲区末尾之外。

例如，在 JDK 8 上执行 `JSON.parse("\"\\")` 会抛出 `ArrayIndexOutOfBoundsException`，而不是 `JSONException`。由于 `ArrayIndexOutOfBoundsException` 不是 `JSONException` 的子类，调用方无法通过 `catch (JSONException)` 处理该非法输入。

本次修改在读取反斜杠后的字符之前检查是否已经到达输入末尾，并抛出 `JSONException`，使 UTF16 reader 的行为与 UTF8 reader 及其他非法 JSON 输入保持一致。

### 修改摘要

- 在 `JSONReaderUTF16.readString()` 的反斜杠处理分支增加输入结束检查。
- 增加根字符串、对象属性值和数组元素中截断转义的回归测试。
- 覆盖 `String`、`char[]` 和 UTF8 字节数组解析入口。
- 普通无转义字符串的处理路径保持不变。

已执行以下验证：

```text
mvn -pl core -DskipTests compile
mvn -pl core -Dtest=JSONReaderUTF16Test surefire:test
```

上述命令均执行成功。

#7790

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -3284,7 +3284,10 @@ final class JSONReaderUTF16
 
                 char c = chars[offset];
                 if (c == '\\') {
-                    c = chars[++offset];
+                    if (++offset == end) {
+                        throw error("invalid escape character EOI");
+                    }
+                    c = chars[offset];
                     if (c == 'u') {
                         c = (char) hexDigit4(chars, offset + 1, end);
                         offset += 4;
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -3350,7 +3350,7 @@ final class JSONReaderUTF16
     }
 
     private void checkEscapeEnd(int offset) {
-        if (offset > end) {
+        if (offset >= end) {
             throw error("invalid escape character EOI");
         }
     }
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -969,7 +969,7 @@ final class JSONReaderUTF16
 
             if (ch == '\\') {
                 nameEscape = true;
-                ch = chars[offset++];
+                ch = charAfterEscape(offset++);
                 switch (ch) {
                     case 'u': {
                         ch = (char) hexDigit4(chars, offset, end);
@@ -977,7 +977,7 @@ final class JSONReaderUTF16
                         break;
                     }
                     case 'x': {
-                        ch = char2(chars[offset], chars[offset + 1]);
+                        ch = char2(chars, offset, end);
                         offset += 2;
                         break;
                     }
@@ -1022,7 +1022,7 @@ final class JSONReaderUTF16
             for (int i = 0; ; ++i) {
                 if (ch == '\\') {
                     nameEscape = true;
-                    ch = chars[offset++];
+                    ch = charAfterEscape(offset++);
                     switch (ch) {
                         case 'u': {
                             ch = (char) hexDigit4(chars, offset, end);
@@ -1030,7 +1030,7 @@ final class JSONReaderUTF16
                             break;
                         }
                         case 'x': {
-                            ch = char2(chars[offset], chars[offset + 1]);
+                            ch = char2(chars, offset, end);
                             offset += 2;
                             break;
                         }
@@ -1484,7 +1484,7 @@ final class JSONReaderUTF16
 
                 if (c == '\\') {
                     nameEscape = true;
-                    c = chars[++offset];
+                    c = charAfterEscape(++offset);
                     switch (c) {
                         case 'u': {
                             c = (char) hexDigit4(chars, offset + 1, end);
@@ -1492,7 +1492,7 @@ final class JSONReaderUTF16
                             break;
                         }
                         case 'x': {
-                            c = char2(chars[offset + 1], chars[offset + 2]);
+                            c = char2(chars, offset + 1, end);
                             offset += 2;
                             break;
                         }
@@ -1550,7 +1550,7 @@ final class JSONReaderUTF16
                 char c = chars[offset];
                 if (c == '\\') {
                     nameEscape = true;
-                    c = chars[++offset];
+                    c = charAfterEscape(++offset);
                     switch (c) {
                         case 'u': {
                             c = (char) hexDigit4(chars, offset + 1, end);
@@ -1558,7 +1558,7 @@ final class JSONReaderUTF16
                             break;
                         }
                         case 'x': {
-                            c = char2(chars[offset + 1], chars[offset + 2]);
+                            c = char2(chars, offset + 1, end);
                             offset += 2;
                             break;
                         }
@@ -1697,7 +1697,7 @@ final class JSONReaderUTF16
 
             if (ch == '\\') {
                 nameEscape = true;
-                ch = chars[++offset];
+                ch = charAfterEscape(++offset);
                 switch (ch) {
                     case 'u': {
                         ch = (char) hexDigit4(chars, offset + 1, end);
@@ -1705,7 +1705,7 @@ final class JSONReaderUTF16
                         break;
                     }
                     case 'x': {
-                        ch = char2(chars[offset + 1], chars[offset + 2]);
+                        ch = char2(chars, offset + 1, end);
                         offset += 2;
                         break;
                     }
@@ -1734,7 +1734,7 @@ final class JSONReaderUTF16
                 char c = chars[offset];
                 if (c == '\\') {
                     nameEscape = true;
-                    c = chars[++offset];
+                    c = charAfterEscape(++offset);
                     switch (c) {
                         case 'u': {
                             c = (char) hexDigit4(chars, offset + 1, end);
@@ -1742,7 +1742,7 @@ final class JSONReaderUTF16
                             break;
                         }
                         case 'x': {
-                            c = char2(chars[offset + 1], chars[offset + 2]);
+                            c = char2(chars, offset + 1, end);
                             offset += 2;
                             break;
                         }
@@ -1817,7 +1817,7 @@ final class JSONReaderUTF16
             char c = chars[offset];
 
             if (c == '\\') {
-                c = chars[++offset];
+                c = charAfterEscape(++offset);
                 switch (c) {
                     case 'u': {
                         c = (char) hexDigit4(chars, offset + 1, end);
@@ -1825,7 +1825,7 @@ final class JSONReaderUTF16
                         break;
                     }
                     case 'x': {
-                        c = char2(chars[offset + 1], chars[offset + 2]);
+                        c = char2(chars, offset + 1, end);
                         offset += 2;
                         break;
                     }
@@ -1869,7 +1869,7 @@ final class JSONReaderUTF16
             char c = chars[offset];
 
             if (c == '\\') {
-                c = chars[++offset];
+                c = charAfterEscape(++offset);
                 switch (c) {
                     case 'u': {
                         c = (char) hexDigit4(chars, offset + 1, end);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
```diff
@@ -3284,7 +3284,10 @@ public String readString() {
 
                 char c = chars[offset];
                 if (c == '\\') {
-                    c = chars[++offset];
+                    if (++offset == end) {
+                        throw error("invalid escape character EOI");
+                    }
```
wenshao: **[Critical]** This EOI guard is added to only one of the structurally identical escape scans in this class. The sibling scans in this same file still read the character after a backslash with no end-of-input check, so after this change the PR's stated invariant — a truncated backslash escape throws `JSONException` instead of `ArrayIndexOutOfBoundsException`, catchable via `catch (JSONException)` — is still violated on public-API paths.

Unguarded sites verified one by one (all reachable through public APIs; `char[]` input routes to `JSONReaderUTF16` unconditionally): `readString0()` pre-scan (~3106) and decode loop (~3126) — reached via typed parse, e.g. `JSON.parseObject("{\"a\":\"\\", Bean.class)` through `readNumber0()`; `readFieldNameHashCode0` (~1487, ~1553); `readFieldNameHashCodeUnquote` (972, 1025, plus its `'x'` arms reading `chars[offset]`/`chars[offset + 1]` unchecked); `readFieldName` (1991); `getFieldName` (~1929); `getNameHashCodeLCase` (~1814, ~1866); `readValueHashCode0` (1700, ~1737); `getString` (~3066); `skipName` (~3355); `skipString` (3490, dispatched from `skipValue` for every skipped string value). `skipStringEscaped` (3534) has the same shape but zero call sites.

Compiled and executed at this PR's head (witness, trimmed):

```
parse(char[]) {"a\   => ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4 @ readFieldNameHashCode0:1487
parse(byte[]) {"a\   => JSONException: invalid escape character EOI   <- same input, different exception type
parseObject({"x":"\ char[], bean without field x) => AIOOBE Index 7 @ skipString:3490
parseObject({a\ char[], Map, AllowUnQuotedFieldNames) => AIOOBE Index 3 @ readFieldNameHashCodeUnquote:972
enum parse "ab\      => AIOOBE @ readValueHashCode0:1700
sub-range reader of(char[],0,8) => reads past logical end=8 into the rest of the buffer
```

On sub-range readers (public `JSONReader.of(char[], int, int)` with `end < chars.length`) the unchecked reads consume characters past the logical end — silent misparse rather than an exception.

Suggested fix: apply the same inline guard shape to each escape scan that consumes the character after a backslash (e.g. `if (offset + 1 == end) throw error("invalid escape character EOI")` in the `readString0()` pre-scan, `if (++offset == end) throw error(...)` in the decode loops). A scratch build with the siblings guarded flips every scenario above to `JSONException`. These sites are pre-existing (this PR introduces none of them), so extending the guard here or in an immediate follow-up are both reasonable — but as-is the PR's own consistency claim does not hold.

**[Critical]** 这个 EOI 守卫只加在了本类中多个结构相同的转义扫描之一上。同一文件里的兄弟扫描路径在反斜杠之后仍然无检查地读取下一个字符，因此本 PR 合入后，其声称的不变量（截断的反斜杠转义抛出 `JSONException` 而非 `ArrayIndexOutOfBoundsException`，可被 `catch (JSONException)` 捕获）在公开 API 路径上仍然被违反。

逐一核实的无守卫位置（均可经公开 API 到达；`char[]` 输入无条件路由到 `JSONReaderUTF16`）：`readString0()` 预扫描（~3106）与解码循环（~3126）——可经类型化解析到达，如 `JSON.parseObject("{\"a\":\"\\", Bean.class)` 经 `readNumber0()`；`readFieldNameHashCode0`（~1487、~1553）；`readFieldNameHashCodeUnquote`（972、1025，其 `'x'` 分支还直接无检查地读 `chars[offset]`/`chars[offset + 1]`）；`readFieldName`（1991）；`getFieldName`（~1929）；`getNameHashCodeLCase`（~1814、~1866）；`readValueHashCode0`（1700、~1737）；`getString`（~3066）；`skipName`（~3355）；`skipString`（3490，由 `skipValue` 对每个被跳过的字符串值分派）。`skipStringEscaped`（3534）形状相同但无任何调用方。

在本 PR head 上编译执行（证据，有删减）：

```
parse(char[]) {"a\   => ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4 @ readFieldNameHashCode0:1487
parse(byte[]) {"a\   => JSONException: invalid escape character EOI   <- 同样输入，异常类型不同
parseObject({"x":"\ char[], 无字段 x 的 bean) => AIOOBE Index 7 @ skipString:3490
parseObject({a\ char[], Map, AllowUnQuotedFieldNames) => AIOOBE Index 3 @ readFieldNameHashCodeUnquote:972
枚举解析 "ab\        => AIOOBE @ readValueHashCode0:1700
子区间 reader of(char[],0,8) => 越过逻辑末尾 end=8 继续读取缓冲区剩余内容
```

在子区间 reader（公开的 `JSONReader.of(char[], int, int)`，`end < chars.length`）上，无守卫的读取会越过逻辑末尾继续消费缓冲区——产生静默误解析而非异常。

建议修复：对每个消费反斜杠后字符的转义扫描应用相同形状的内联守卫（如 `readString0()` 预扫描处 `if (offset + 1 == end) throw error("invalid escape character EOI")`、解码循环处 `if (++offset == end) throw error(...)`）。在 scratch 构建中给兄弟位置加上守卫后，上述所有场景均翻转为 `JSONException`。这些位置均为既有代码（本 PR 未引入任何一处），因此在本 PR 中扩展守卫或立即提交后续 PR 修复都合理——但就当前状态而言，PR 自身声称的行为一致性并不成立。

</details>

core/src/test/java/com/alibaba/fastjson2/JSONReaderUTF16Test.java
```diff
@@ -0,0 +1,25 @@
...
+                "\"\\",
+                "{\"a\":\"\\",
+                "[\"\\",
+                "\"ab\\"
+        };
+
+        for (String input : inputs) {
+            assertThrows(JSONException.class, () -> JSON.parse(input));
+            assertThrows(JSONException.class, () -> JSON.parse(input.toCharArray()));
+            assertThrows(JSONException.class, () -> JSON.parse(input.getBytes(StandardCharsets.UTF_8)));
```
wenshao: **[Suggestion]** This byte[] assertion never reaches a byte reader: all four inputs are pure ASCII, and `JSONReaderUTF8.of(byte[], ...)` routes pure-ASCII bytes to `JSONReaderASCII` — a class this PR does not modify, which already threw `JSONException` for these exact inputs before the change. The line therefore verifies zero changed lines while giving the impression that the byte path is covered. Meanwhile the same truncated-escape defect remains for non-ASCII byte input: executed at this PR's head, `JSON.parse("\"中\\".getBytes(UTF_8))` throws `ArrayIndexOutOfBoundsException: Index 5 out of bounds for length 5` from `JSONReaderUTF8.readString` (~5295). Either drop this line (String/char[] already cover the fix) or add at least one non-ASCII byte[] input so the assertion actually exercises a byte reader — noting that such an input fails today due to the pre-existing `JSONReaderUTF8` gap, which fixing is a reasonable separate follow-up.

**[Suggestion]** 这条 byte[] 断言从未到达字节 reader：四个输入全部是纯 ASCII，而 `JSONReaderUTF8.of(byte[], ...)` 会把纯 ASCII 字节路由到 `JSONReaderASCII` —— 该类未被本 PR 修改，且在本修改之前就已经对这些输入抛 `JSONException`。因此这一行验证的是零行被改动的代码，却制造了字节路径已被覆盖的假象。同时，同类截断转义缺陷在非 ASCII 字节输入上依然存在：在本 PR head 上实际执行，`JSON.parse("\"中\\".getBytes(UTF_8))` 从 `JSONReaderUTF8.readString`（~5295）抛出 `ArrayIndexOutOfBoundsException: Index 5 out of bounds for length 5`。要么删除这一行（String/char[] 已覆盖修复），要么增加至少一个非 ASCII 字节输入使断言真正进入字节 reader —— 注意该输入目前会因 `JSONReaderUTF8` 的既有缺陷而失败，修复它可以作为独立的后续工作。

</details>

core/src/test/java/com/alibaba/fastjson2/JSONReaderUTF16Test.java
```diff
@@ -0,0 +1,25 @@
...
+    public void truncatedEscape() {
+        String[] inputs = {
+                "\"\\",
+                "{\"a\":\"\\",
+                "[\"\\",
+                "\"ab\\"
+        };
+
+        for (String input : inputs) {
+            assertThrows(JSONException.class, () -> JSON.parse(input));
```
wenshao, line 21: **[Suggestion]** The String assertion does not reach the patched code on JDK 9+: all four inputs are LATIN1, and `JSONReader.of(String)` routes LATIN1 compact strings to `JSONReaderASCII` (probed on JDK 11: all four inputs construct `JSONReaderASCII`), which already threw `JSONException` for these inputs before this PR and is untouched by the diff. Only the `char[]` line reliably exercises the fix on every JDK. A revert probe confirmed the blind spot: with the fix reverted, the String assertion still passes while the char[] one fails. Add at least one non-LATIN1 input (e.g. `"\"中\\"`) to `inputs`, so both the String and char[] assertions route to `JSONReaderUTF16` on every JDK.

**[Suggestion]** String 断言在 JDK 9+ 上不会到达被修复的代码：四个输入全部是 LATIN1，而 `JSONReader.of(String)` 会把 LATIN1 紧凑字符串路由到 `JSONReaderASCII`（在 JDK 11 上实测：四个输入全部构造出 `JSONReaderASCII`），该类在本 PR 之前就会对这些输入抛 `JSONException`，且未被本 diff 触碰。只有 `char[]` 一行在所有 JDK 上可靠地执行被修复的代码。回退验证确认了这个盲区：回退本修复后，String 断言依旧通过，而 char[] 断言失败。在 `inputs` 中增加至少一个非 LATIN1 输入（如 `"\"中\\"`），使 String 与 char[] 两条断言在每个 JDK 上都路由到 `JSONReaderUTF16`。

</details>

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
annoymous3: Thanks for the detailed review. Addressed in the latest commit.

I audited all escape scans in `JSONReaderUTF16` and applied logical `end` checks consistently across field-name scanning, value hashing, string decoding, skipped names/values, enum parsing, and object-reader paths.

The fix also validates the complete span of truncated `\u` and `\x` escapes and uses the logical reader boundary rather than the backing array length, preventing sub-range readers from consuming data beyond their configured range.

Regression tests now cover the reported public API paths, unknown-field skipping, unquoted field names, enum parsing, incomplete `\u`/`\x` escapes, and sub-range readers. The targeted reader regression suite passes with 158 tests and no failures.

core/src/test/java/com/alibaba/fastjson2/JSONReaderUTF16Test.java
annoymous3: Thanks. I added a non-LATIN1 truncated input (`"中\`) so the String assertion reliably routes to `JSONReaderUTF16` on JDK 9+, while the `char[]` assertions continue to exercise UTF16 directly.

core/src/test/java/com/alibaba/fastjson2/JSONReaderUTF16Test.java
annoymous3: Thanks. I removed the pure-ASCII `byte[]` assertion because it routes to `JSONReaderASCII` and does not exercise the changed UTF16 code.I decided to cover the non-ASCII UTF-8 path in this PR as well.

I audited all backslash escape branches in `JSONReaderUTF8` and added logical end-of-input guards for the character following a backslash and for truncated `\x` escapes. I also added a non-ASCII UTF-8 regression using `"中\`, which reliably exercises `JSONReaderUTF8`, covering root values, field names, typed parsing, skipped unknown fields, enum parsing, and sub-range readers.

The related UTF8/UTF16 reader tests pass, and `mvn -pl core validate` succeeds.

PR #7797, opened 2026-08-21
fix: resolve $ref for shared objects nested under numeric-keyed maps

### LINKED ISSUE #3516: [BUG]java对象存在map套map且内层map是同一个对象，序列化结果带ref，反序列化失败
### 问题描述
如下代码，对象map套map，内层的reversePriceMap在多个外层对象中是同一个对象，序列化出来的字符串带有ref，但是在反序列化的时候失败。在fastjson中是直接抛异常，在fastjson2中是反序列化出来的内层对象是null

### 环境信息
*请填写以下信息：*

 - OS信息：  macos arm
 - JDK信息： 1.8
 - 版本信息：2.0.57
 

### 重现步骤
*如何操作可以重现该问题：*

1. 使用 `xxx.xxx` 方法
2. 输入 `...`  数据
3. 出现 `...` 错误
```java
        List<RiskInfo> list = buildObject();
        String json = JSON.toJSONString(list);
        System.out.println(json);

        List<RiskInfo> list2 = JSON.parseObject(json, new TypeReference<List<RiskInfo>>(){});
        System.out.println(list2);
    }


    @Data
    public static class RiskInfo{
        private Map<Long, Item> itemMap;
    }

    @Data
    public static class Item {
        private Long itemId;
        private Map<String, Long> reversePriceMap;
    }


    @NotNull
    private static List<RiskInfo> buildObject() {
        RiskInfo riskInfo1 = new RiskInfo();
        Item item = new Item();
        item.setItemId(1L);
        Map<String, Long> reversePriceMap = new HashMap<>();
        reversePriceMap.put("1", 1L);
        Map<Long, Item> itemMap = new HashMap<>();
        itemMap.put(item.getItemId(), item);
        item.setReversePriceMap(reversePriceMap);
        riskInfo1.setItemMap(itemMap);

        RiskInfo riskInfo2 = new RiskInfo();
        Item item2 = new Item();
        item2.setItemId(2L);
        item2.setReversePriceMap(reversePriceMap);
        Map<Long , Item> itemMap2 = new HashMap<>();
        itemMap2.put(item2.getItemId(), item2);
        riskInfo2.setItemMap(itemMap2);
//        riskInfo2.setItemMap(itemMap);
        return Lists.newArrayList(riskInfo1, riskInfo2);
    }
```

### 期待的正确结果
[{"itemMap":{1:{"itemId":1,"reversePriceMap":{"1":1}}}},{"itemMap":{2:{"itemId":2,"reversePriceMap":{"$ref":"$[0].itemMap.reversePriceMap"}}}}]
[JsonRefTest2.RiskInfo(itemMap={1=JsonRefTest2.Item(itemId=1, reversePriceMap={1=1})}), JsonRefTest2.RiskInfo(itemMap={2=JsonRefTest2.Item(itemId=2, reversePriceMap={1=1})})]

### 实际输出
[{"itemMap":{1:{"itemId":1,"reversePriceMap":{"1":1}}}},{"itemMap":{2:{"itemId":2,"reversePriceMap":{"$ref":"$[0].itemMap.reversePriceMap"}}}}]
[JsonRefTest2.RiskInfo(itemMap={1=JsonRefTest2.Item(itemId=1, reversePriceMap={1=1})}), JsonRefTest2.RiskInfo(itemMap={2=JsonRefTest2.Item(itemId=2, reversePriceMap=null)})]

### 相关日志输出
暂无

#### 附加信息

1919chichi:
## Summary

Fixes #3516.

When a `Map<Long, V>` / `Map<Integer, V>` field was serialized with `JSONWriter.Feature.ReferenceDetection` enabled, an object shared between two map entries got a `$ref` path that silently dropped the map's key segment (e.g. `$[0].itemMap.reversePriceMap` instead of `$[0].itemMap.1.reversePriceMap`). On read, that broken path resolved to `null` instead of the shared object.

Root cause, on both ends:

- **Write side** (`ObjectWriterImplMap.writeMapKey`): for `Integer`/`Long` keys written in raw (unquoted) form, the method returned `strKey == null`. The caller uses `strKey != null` as the gate for whether to call `jsonWriter.setPath(...)` for that entry, so the key segment was never pushed onto the path stack — nested `$ref` paths ended up pointing one level too high.
- **Read side** (`JSONPathSegmentName.eval`): the numeric-key fallback matching (used when `map.get(name)` fails because `name` is a `String` but the actual key is boxed numeric) only handled `Long` keys, not `Integer`. So even once the write side produces a correct path with an `Integer` key segment, resolution would still fail.

Both are fixed:
- `writeMapKey` now sets `strKey = key.toString()` for `Integer`/`Long` keys (the actual bytes written to the JSON key are unchanged — still unquoted).
- `JSONPathSegmentName.eval` gets an `Integer` fallback branch mirroring the existing `Long` one.

## Test plan

- [x] Added `RefTest8` covering: the exact issue #3516 repro (`Map<Long,Item>` with a shared inner `Map<String,Long>`, asserting the exact `$ref` path string and reference identity after round-trip), a direct-share case for `Long` keys, a direct-share case for `Integer` keys, and a regression guard proving output is byte-for-byte unchanged when there's no shared object to reference.
- [x] Verified all four new tests fail without the fix (reverted the fix locally, reran) and pass with it.
- [x] `mvn test -pl core` (full suite) passes.
- [x] `mvn validate -pl core` (checkstyle) passes.

<details>
<summary>中文说明</summary>

修复 #3516。

当 `Map<Long, V>` / `Map<Integer, V>` 字段在开启 `JSONWriter.Feature.ReferenceDetection` 的情况下序列化时，被两个 map entry 共享的对象生成的 `$ref` 路径会漏掉 map key 这一段（例如 `$[0].itemMap.reversePriceMap`，而不是正确的 `$[0].itemMap.1.reversePriceMap`）。反序列化时这个残缺路径解析不出对象，结果被静默置为 `null`。

根因分两处：

- **写端**（`ObjectWriterImplMap.writeMapKey`）：对于以裸数字形式写出的 `Integer`/`Long` key，该方法返回的 `strKey` 始终是 `null`。调用方用 `strKey != null` 作为"是否要为这个 entry 调用 `jsonWriter.setPath(...)`"的开关，导致 key 这一层路径从未被压入路径栈，后续嵌套对象的 `$ref` 路径因此少了一层。
- **读端**（`JSONPathSegmentName.eval`）：当 `map.get(name)` 因为 `name` 是 `String` 而实际 key 是装箱数字类型而失败时，有一个数字兜底匹配逻辑，但只处理了 `Long`，没处理 `Integer`。所以即便写端修好、生成了带 `Integer` key 的正确路径，读端依然解析不出来。

两处都已修复：
- `writeMapKey` 对 `Integer`/`Long` key 补上 `strKey = key.toString()`（实际写入 JSON 的字节不变，依然是不带引号的数字）。
- `JSONPathSegmentName.eval` 参照现有的 `Long` 分支，新增了 `Integer` 分支。

**测试**：新增 `RefTest8`，覆盖 issue #3516 的原始复现场景、`Long`/`Integer` key 直接共享场景，以及一个回归测试（证明没有共享对象时输出的 JSON 文本完全不变）。已验证：去掉修复后这些测试会失败，恢复修复后全部通过；`mvn test -pl core` 全量测试通过；`mvn validate -pl core` checkstyle 检查通过。

</details>

---

## Review follow-up — round 1

The first review round found that the write-side fix opened up a `$ref` path shape the read side had never been exercised with: a segment that is a bare number pointing at a map whose keys are not Strings. Four read-side gaps and one hot-path allocation were reported; all five are addressed in `5425ac9`, `f3fe7d2` and `ca816ee`. Reproductions and details are in the individual review threads — summary below.

### Read side: one shared lookup instead of two copies

The root cause of R1-1 was structural. `JSONPathSegmentName` and `JSONPathSingleName` each carried a near-identical copy of the "String segment to typed map key" fallback, so a fix in one was invisible to the other — and single-segment paths (`$.1`), which is exactly what a root-level map produces, are evaluated by the copy that was never updated.

Both now delegate to one `JSONPathSegmentName#getMapValue`, and the three remaining defects were fixed inside it:

| Finding | Problem | Fix |
| --- | --- | --- |
| R1-1 | root `Map<Integer, V>` reference `$.1` resolved to null | `JSONPathSingleName#eval` delegates to the shared helper; the duplicate is deleted |
| R1-4 | `map.get(name)` throws `ClassCastException` on a comparator-based map (declared `SortedMap`/`TreeMap`) before the fallback can run | the exact lookup is guarded, catching only `ClassCastException`, then resolution continues with the typed-key scan |
| R1-3 | an exact String key mapped to `null` was treated as absent and shadowed by a numeric key | fall back only when `map.containsKey(name)` is false |
| R1-2 | `Integer.parseInt` aborted the whole evaluation for a segment outside the 32-bit range | parse once as `long` outside the loop and derive the `Integer` candidate only when it is in range; a segment beyond `long` matches nothing instead of throwing |

R1-3, and part of R1-2, were **pre-existing on `main`** for `Long`/`Enum` keys rather than introduced by this PR — for example `JSONPath.eval(map, "$.99999999999999999999")` throws on `main` today when the map has a `Long` key. Since the change touches this exact code, they are fixed here rather than left behind.

Side effect: `JSONPathSegmentName#eval` shrank from 949 to 726 bytes of bytecode, since the entrySet scan no longer sits inside the hot method.

### Write side: derive the ref path lazily (R1-5)

`strKey` was computed at the top of the loop body, ahead of the primitive-value fast paths that all `continue` without reading it — so a `Map<Long, Long>` allocated one immediately-garbage String per entry even with reference detection disabled.

`writeMapKey` is now **byte-for-byte identical to `main`**, and the path String is derived inside the reference-detection branch, matching what the JSONB branch in this same file already does. With reference detection off the cost is two `instanceof` checks and zero allocation; with it on, `toString()` runs only for the entries that actually get a path pushed onto the stack.

### Verification

- `mvn clean test -pl core` — **7984 tests, 0 failures, 0 errors**
- `mvn validate -pl core` — checkstyle passes
- `RefTest8` grew from 4 to **10** tests: `SortedMap` round trips for Long and Integer keys, root `Map<Integer,V>` / `Map<Long,V>` round trips asserting the exact `$ref` path and `assertSame`, exact-null String key semantics on both evaluators, and ordered mixed Integer/Long keys including a segment beyond `long`
- Each guard was removed one at a time and the matching test confirmed to turn red (4/4)

第一轮评审指出：写端的修复放开了一类读端从未被压测过的 `$ref` 路径形态——路径段是裸数字、且指向一个 key 不是 String 的 Map。共报告 4 处读端缺口和 1 处热路径分配，全部已在 `5425ac9`、`f3fe7d2`、`ca816ee` 中处理。复现细节见各条评审线程，摘要如下。

### 读端：用一份共享查找替换两份拷贝

R1-1 的根因是结构性的。`JSONPathSegmentName` 和 `JSONPathSingleName` 各自持有一份几乎相同的「String 段名 → 类型化 map key」兜底逻辑，改一份对另一份不可见——而单段路径（`$.1`）恰恰由那份从未更新的拷贝求值，根级 map 产生的正是单段路径。

现在两者都委托给同一个 `JSONPathSegmentName#getMapValue`，其余三个缺陷在其中一并修复：

| 问题 | 现象 | 修法 |
| --- | --- | --- |
| R1-1 | 根级 `Map<Integer, V>` 的引用 `$.1` 解析为 null | `JSONPathSingleName#eval` 委托给共享 helper，删除重复实现 |
| R1-4 | 比较型 Map（声明为 `SortedMap`/`TreeMap`）上 `map.get(name)` 抛 `ClassCastException`，兜底逻辑根本跑不到 | 保护精确查询，只捕获 `ClassCastException`，之后继续执行类型化 key 扫描 |
| R1-3 | 精确 String key 映射到 `null` 时被当作不存在，被数值 key 遮蔽 | 仅当 `map.containsKey(name)` 为 false 时才降级 |
| R1-2 | 超出 32 位范围的路径段上 `Integer.parseInt` 中断整个求值 | 循环外用 `long` 解析一次，只在 int 范围内构造 `Integer` 候选；超出 `long` 范围的段不匹配任何 key 而非抛异常 |

其中 R1-3 与 R1-2 的一部分在 `main` 上对 `Long`/`Enum` key **本就存在**，并非本 PR 引入——例如 map 含 `Long` key 时，`JSONPath.eval(map, "$.99999999999999999999")` 在 `main` 上今天就会抛异常。既然改动正好落在这段代码上，就在这里一并修掉而不是留在原地。

附带效果：`JSONPathSegmentName#eval` 的字节码从 949 字节降到 726 字节，entrySet 扫描不再位于这个热方法内部。

### 写端：路径字符串改为惰性求值（R1-5）

`strKey` 原先在循环体开头求值，位于那些全部 `continue` 掉、从不读取它的 primitive value 快速路径之前——因此 `Map<Long, Long>` 即使关闭引用检测，每个 entry 也会分配一个立即变成垃圾的 String。

现在 `writeMapKey` 与 `main` **逐字节一致**，路径字符串改在引用检测分支内部求值，与本文件 JSONB 分支已有的写法保持一致。关闭引用检测时的代价是两次 `instanceof`、零分配；开启时 `toString()` 只对真正会被压入路径栈的 entry 执行。

### 验证

- `mvn clean test -pl core` —— **7984 个测试，0 失败 0 错误**
- `mvn validate -pl core` —— checkstyle 通过
- `RefTest8` 从 4 个测试扩充到 **10** 个：Long/Integer key 的 `SortedMap` 往返、根级 `Map<Integer,V>` / `Map<Long,V>` 往返（断言精确 `$ref` 路径与 `assertSame`）、两个 evaluator 上精确 null String key 的语义、以及含超出 `long` 范围段的有序 Integer/Long 混合 key
- 每个保护逻辑均单独移除并确认对应测试变红（4/4）

</details>

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
@@ -202,6 +202,7 @@ class JSONPathSegmentName
             if (value == null) {
                 boolean isNum = IOUtils.isNumber(this.name);
                 Long longValue = null;
+                Integer intValue = null;
 
                 for (Object o : map.entrySet()) {
                     Map.Entry entry = (Map.Entry) o;
@@ -214,6 +215,14 @@ class JSONPathSegmentName
                             longValue = Long.parseLong(this.name);
                         }
                         if (entryKey.equals(longValue)) {
+                            value = entry.getValue();
+                            break;
+                        }
+                    } else if (entryKey instanceof Integer) {
+                        if (intValue == null && isNum) {
+                            intValue = Integer.parseInt(this.name);
+                        }
+                        if (entryKey.equals(intValue)) {
                             value = entry.getValue();
                             break;
                         }
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplMap.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplMap.java
@@ -436,8 +436,10 @@ public final class ObjectWriterImplMap
             } else {
                 if (key instanceof Integer) {
                     jsonWriter.writeName((Integer) key);
+                    strKey = key.toString();
                 } else if (key instanceof Long) {
                     jsonWriter.writeName((Long) key);
+                    strKey = key.toString();
                 } else {
                     jsonWriter.writeNameAny(key);
                 }
--- a/core/src/test/java/com/alibaba/fastjson2/codec/RefTest8.java
+++ b/core/src/test/java/com/alibaba/fastjson2/codec/RefTest8.java
@@ -6,6 +6,8 @@ import com.alibaba.fastjson2.TypeReference;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -40,7 +42,7 @@ public class RefTest8 {
         riskInfo2.itemMap = new HashMap<>();
         riskInfo2.itemMap.put(item2.itemId, item2);
 
-        List<RiskInfo> list = List.of(riskInfo1, riskInfo2);
+        List<RiskInfo> list = Arrays.asList(riskInfo1, riskInfo2);
 
         String json = JSON.toJSONString(list, JSONWriter.Feature.ReferenceDetection);
         assertEquals(
@@ -53,7 +55,7 @@ public class RefTest8 {
         });
         Item item1Parsed = list2.get(0).itemMap.get(1L);
         Item item2Parsed = list2.get(1).itemMap.get(2L);
-        assertEquals(Map.of("1", 1L), item1Parsed.reversePriceMap);
+        assertEquals(Collections.singletonMap("1", 1L), item1Parsed.reversePriceMap);
         assertSame(item1Parsed.reversePriceMap, item2Parsed.reversePriceMap);
     }
 
--- a/core/src/test/java/com/alibaba/fastjson2/codec/RefTest8.java
+++ b/core/src/test/java/com/alibaba/fastjson2/codec/RefTest8.java
@@ -1,0 +1,136 @@
+package com.alibaba.fastjson2.codec;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONWriter;
+import com.alibaba.fastjson2.TypeReference;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+
+/**
+ * https://github.com/alibaba/fastjson2/issues/3516
+ * A shared object nested one level below a Map keyed by Long/Integer produced a $ref path
+ * that skipped the map key segment, so the reference could not be resolved on read and the
+ * field was silently set to null instead of pointing back at the shared object.
+ */
+@Tag("codec")
+public class RefTest8 {
+    @Test
+    public void test_issue3516_longKeyedMap() {
+        Map<String, Long> reversePriceMap = new HashMap<>();
+        reversePriceMap.put("1", 1L);
+
+        Item item1 = new Item();
+        item1.itemId = 1L;
+        item1.reversePriceMap = reversePriceMap;
+        RiskInfo riskInfo1 = new RiskInfo();
+        riskInfo1.itemMap = new HashMap<>();
+        riskInfo1.itemMap.put(item1.itemId, item1);
+
+        Item item2 = new Item();
+        item2.itemId = 2L;
+        item2.reversePriceMap = reversePriceMap;
+        RiskInfo riskInfo2 = new RiskInfo();
+        riskInfo2.itemMap = new HashMap<>();
+        riskInfo2.itemMap.put(item2.itemId, item2);
+
+        List<RiskInfo> list = List.of(riskInfo1, riskInfo2);
+
+        String json = JSON.toJSONString(list, JSONWriter.Feature.ReferenceDetection);
+        assertEquals(
+                "[{\"itemMap\":{1:{\"itemId\":1,\"reversePriceMap\":{\"1\":1}}}},"
+                        + "{\"itemMap\":{2:{\"itemId\":2,\"reversePriceMap\":{\"$ref\":\"$[0].itemMap.1.reversePriceMap\"}}}}]",
+                json
+        );
+
+        List<RiskInfo> list2 = JSON.parseObject(json, new TypeReference<List<RiskInfo>>() {
+        });
+        Item item1Parsed = list2.get(0).itemMap.get(1L);
+        Item item2Parsed = list2.get(1).itemMap.get(2L);
+        assertEquals(Map.of("1", 1L), item1Parsed.reversePriceMap);
+        assertSame(item1Parsed.reversePriceMap, item2Parsed.reversePriceMap);
+    }
+
+    @Test
+    public void test_longKeyedMap_directShare() {
+        Item shared = new Item();
+        shared.itemId = 100L;
+
+        Bean bean = new Bean();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1L, shared);
+        bean.map2 = new HashMap<>();
+        bean.map2.put(2L, shared);
+
+        String json = JSON.toJSONString(bean, JSONWriter.Feature.ReferenceDetection);
+        assertEquals(
+                "{\"map1\":{1:{\"itemId\":100}},\"map2\":{2:{\"$ref\":\"$.map1.1\"}}}",
+                json
+        );
+
+        Bean bean2 = JSON.parseObject(json, Bean.class);
+        assertSame(bean2.map1.get(1L), bean2.map2.get(2L));
+        assertEquals(100L, bean2.map1.get(1L).itemId);
+    }
+
+    @Test
+    public void test_integerKeyedMap_directShare() {
+        Item shared = new Item();
+        shared.itemId = 200L;
+
+        BeanInt bean = new BeanInt();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1, shared);
+        bean.map2 = new HashMap<>();
+        bean.map2.put(2, shared);
+
+        String json = JSON.toJSONString(bean, JSONWriter.Feature.ReferenceDetection);
+        assertEquals(
+                "{\"map1\":{1:{\"itemId\":200}},\"map2\":{2:{\"$ref\":\"$.map1.1\"}}}",
+                json
+        );
+
+        BeanInt bean2 = JSON.parseObject(json, BeanInt.class);
+        assertSame(bean2.map1.get(1), bean2.map2.get(2));
+        assertEquals(200L, bean2.map1.get(1).itemId);
+    }
+
+    @Test
+    public void test_longKeyedMap_noSharing_outputUnchanged() {
+        Bean bean = new Bean();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1L, new Item());
+        bean.map2 = new HashMap<>();
+        bean.map2.put(2L, new Item());
+
+        assertEquals(
+                "{\"map1\":{1:{}},\"map2\":{2:{}}}",
+                JSON.toJSONString(bean, JSONWriter.Feature.ReferenceDetection)
+        );
+    }
+
+    public static class RiskInfo {
+        public Map<Long, Item> itemMap;
+    }
+
+    public static class Item {
+        public Long itemId;
+        public Map<String, Long> reversePriceMap;
+    }
+
+    public static class Bean {
+        public Map<Long, Item> map1;
+        public Map<Long, Item> map2;
+    }
+
+    public static class BeanInt {
+        public Map<Integer, Item> map1;
+        public Map<Integer, Item> map2;
+    }
+}


core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
```diff
@@ -202,6 +202,7 @@ public void eval(JSONPath.Context context) {
             if (value == null) {
                 boolean isNum = IOUtils.isNumber(this.name);
                 Long longValue = null;
+                Integer intValue = null;
```
wenshao: **[Critical]** R1-4: `TreeMap<Integer, ...>` and `TreeMap<Long, ...>` cannot reach this typed-key fallback. A path such as `$.map1.1` first calls `TreeMap.get("1")`, which compares a `String` with the numeric key and throws `ClassCastException`, so round-tripping a shared value in a declared `SortedMap` fails before the new logic runs. Catch only this incompatible-key lookup case, then continue with typed-key resolution while preserving exact String-key precedence.

Witness:
```text
Integer: ClassCastException at TreeMap.get -> JSONPathSegmentName.eval
Long: ClassCastException at TreeMap.get -> JSONPathSegmentName.eval
With a ClassCastException-safe lookup: Tests run: 2, Errors: 0, BUILD SUCCESS
```

Please add Integer and Long `TreeMap` shared-value round-trip tests and verify they fail when the guard is removed.

**问题：** `TreeMap<Integer, ...>` 和 `TreeMap<Long, ...>` 无法进入这里的类型化 key 兜底逻辑。`$.map1.1` 会先执行 `TreeMap.get("1")`，由于 `String` 与数值 key 不可比较而抛出 `ClassCastException`，因此声明为 `SortedMap` 的共享对象往返解析会在新逻辑运行前失败。应只处理这类不兼容 key 查询异常，然后继续执行类型化 key 解析，同时保留精确 String key 的优先级。

请增加 Integer/Long `TreeMap` 共享对象往返测试，并验证移除该保护后测试会失败。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
```diff
@@ -217,6 +218,14 @@ public void eval(JSONPath.Context context) {
                             value = entry.getValue();
                             break;
                         }
+                    } else if (entryKey instanceof Integer) {
+                        if (intValue == null && isNum) {
+                            intValue = Integer.parseInt(this.name);
+                        }
```
wenshao: **[Critical]** R1-2: This `Integer.parseInt` aborts lookup for numeric segments outside the 32-bit range. If an `Integer` key is encountered before a matching `2147483648L` key, evaluating `$.nested.2147483648` throws `NumberFormatException` and never reaches the valid `Long` entry; an attacker-controlled `$ref` can trigger the same failure. Parse into a wider value once and compare as `Integer` only when it is in range.

Witness:
```text
B direct=long-match
B nestedException=java.lang.NumberFormatException: For input string: "2147483648"
With range-checked conversion: B nested=long-match
```

Please add ordered mixed Integer/Long JSONPath tests and verify removing the range guard makes them fail.

**问题：** 这里的 `Integer.parseInt` 会让超出 32 位范围的数值路径直接中断查询。如果遍历先遇到 Integer key，之后才有匹配的 `2147483648L`，求值 `$.nested.2147483648` 会抛 `NumberFormatException`，无法继续匹配合法的 Long key；攻击者控制的 `$ref` 也能触发同样的失败。应先解析为更宽的数值，只在 int 范围内才构造 Integer 候选。

请增加有序的 Integer/Long 混合 key JSONPath 测试，并验证移除范围保护后测试会失败。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
```diff
@@ -217,6 +218,14 @@ public void eval(JSONPath.Context context) {
                             value = entry.getValue();
                             break;
                         }
+                    } else if (entryKey instanceof Integer) {
+                        if (intValue == null && isNum) {
```
wenshao: **[Critical]** R1-3: The fallback treats an exact String key mapped to `null` as if it were absent. For a map containing both `"1" -> null` and `1 -> "integer-value"`, evaluating `$.nested.1` returns the Integer-key value instead of the exact String-key mapping. Only run Enum/numeric fallback when `map.containsKey(name)` is false.

Witness:
```text
PR code: expected <null> but was <integer-value>; Tests run: 1, Failures: 1
With `value == null && !map.containsKey(name)`: Tests run: 1, Failures: 0, BUILD SUCCESS
```

Please add a mixed-key test that pins the exact null String-key result and verify it turns red when the `containsKey` guard is removed.

**问题：** 精确 String key 的值为 `null` 时，这段逻辑会把它误判为 key 不存在。对于同时包含 `"1" -> null` 和 `1 -> "integer-value"` 的 Map，求值 `$.nested.1` 会错误返回 Integer key 的值，而不是精确 String key 对应的 `null`。只有在 `map.containsKey(name)` 为 false 时才应执行 Enum/数值 key 兜底。

请增加混合 key 测试固定精确 String key 为 null 的语义，并验证移除 `containsKey` 保护后测试会失败。

</details>

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplMap.java
```diff
@@ -436,8 +436,10 @@ String writeMapKey(Object key, JSONWriter jsonWriter, long features) {
             } else {
                 if (key instanceof Integer) {
                     jsonWriter.writeName((Integer) key);
+                    strKey = key.toString();
```
wenshao: **[Critical]** R1-1: This change starts emitting root-level Integer-key references such as `$.1`, but those paths are evaluated by `JSONPathSingleName`, whose Map fallback does not support Integer keys. A root `Map<Integer, Item>` with two keys sharing one value serializes with a `$ref`, then deserializes the second value as `null`. Add safe Integer-key handling to the single-name evaluator as well.

Witness:
```text
A json={1:{"id":7},2:{"$ref":"$.1"}} key1=Item@... key2=null same=false
With bounded Integer-key handling: key1=Item@... key2=Item@... same=true
```

Please add a root `Map<Integer, Item>` round-trip test that asserts the `$.1` path and `assertSame`; removing the single-name Integer branch must make it fail.

**问题：** 该改动会开始生成 `$.1` 这样的根级 Integer key 引用，但此路径由 `JSONPathSingleName` 求值，其 Map 兜底逻辑不支持 Integer key。根级 `Map<Integer, Item>` 中两个 key 共享一个对象时，序列化会产生 `$ref`，反序列化后第二个值却变成 `null`。需要同步为单段 JSONPath evaluator 增加安全的 Integer key 处理。

请增加根级 `Map<Integer, Item>` 往返测试，断言 `$.1` 路径和 `assertSame`，并验证删除单段 Integer 分支后测试会失败。

</details>

wenshao: **[Suggestion]** R1-5: `key.toString()` now runs for every Integer/Long map entry even when reference detection is disabled, before the primitive-value fast paths. Serializing a million-entry numeric-key map therefore creates roughly a million unnecessary temporary Strings, adding allocation and GC pressure to this hot path. Keep numeric name writing allocation-free and derive the path String lazily only on branches that consume it.

**建议：** 即使未启用引用检测，这里的 `key.toString()` 也会对每个 Integer/Long map entry 执行，而且发生在 primitive value 快速路径之前。序列化百万级数值 key Map 时会产生约百万个不必要的临时 String，增加核心热路径的分配与 GC 压力。应保持数值 name 写出无分配，仅在实际消费路径字符串的分支中延迟构造。

</details>

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
1919chichi: Fixed in 5425ac9.

**Reproduced first.** With the field declared as `SortedMap<Long, Item>` the reader instantiates a `TreeMap`, and `map.get("1")` throws `ClassCastException` on the very first line, so the typed-key scan never runs. Worth recording that the declared type is what triggers it, not the data: with the field declared as a plain `Map` the reader builds a `HashMap`, `get` returns null, and everything works — that variant does not reproduce.

**Why this PR has to own it.** On `main` the write side never emitted a numeric key segment (`strKey` was null, so `setPath` was skipped), so the same `TreeMap` simply serialized the shared object twice: identity lost, but no exception. This PR turns that into a hard failure, so it is a regression introduced here rather than a pre-existing one.

**What changed.** Rather than adding a third copy of the fallback, the whole "String segment to typed key" lookup moved into a shared `JSONPathSegmentName#getMapValue`, with the exact lookup guarded there:

```java
static Object getMapValue(Map map, String name) {
    try {
        Object value = map.get(name);
        if (value != null || map.containsKey(name)) {
            return value;
        }
    } catch (ClassCastException ignored) {
        // the map orders keys by a comparator that cannot accept a String, fall through
    }

    return getTypedKeyValue(map, name);
}
```

Only `ClassCastException` is caught, and only around the exact lookup — `entrySet()` iteration is safe on a `TreeMap`, so resolution continues normally afterwards. Exact String-key precedence is preserved because the guarded block returns before the scan is reached.

**Tests.** `RefTest8#test_sortedMap_longKey_directShare` and `#test_sortedMap_integerKey_directShare`, both asserting the `$ref` path text and `assertSame` after a round trip. Verified both turn red with `ClassCastException` when the try/catch is removed.

已在 5425ac9 修复。

**先复现。** 字段声明为 `SortedMap<Long, Item>` 时 reader 实例化的是 `TreeMap`，`map.get("1")` 在第一行就抛 `ClassCastException`，类型化 key 扫描一行都跑不到。这里值得记录的是：触发条件是**声明类型**而不是数据——字段声明为普通 `Map` 时 reader 建的是 `HashMap`，`get` 返回 null，一切正常，那种写法复现不出来。

**为什么这个问题该由本 PR 负责。** 在 `main` 上写端根本不会生成数值 key 路径段（`strKey` 为 null，`setPath` 被跳过），所以同样的 `TreeMap` 只是把共享对象重复序列化两遍：identity 丢失，但不抛异常。本 PR 把它变成了硬失败，所以这是本 PR 引入的回归，不是既有缺陷。

**改法。** 没有再抄第三份兜底逻辑，而是把整个「String 段名 → 类型化 key」的查找抽到共享的 `JSONPathSegmentName#getMapValue`，并在其中保护精确查询（代码见上）。只捕获 `ClassCastException`，且只包住精确查询——`entrySet()` 遍历对 `TreeMap` 是安全的，之后解析照常进行。精确 String key 的优先级得以保留，因为被保护的代码块会在进入扫描前就返回。

**测试。** `RefTest8#test_sortedMap_longKey_directShare` 与 `#test_sortedMap_integerKey_directShare`，均断言 `$ref` 路径文本以及往返后的 `assertSame`。已验证移除 try/catch 后两条都会以 `ClassCastException` 变红。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
1919chichi: Fixed in 5425ac9.

**Reproduced exactly as described.** With `{1 -> "int-value", 2147483648L -> "long-match"}` in a `LinkedHashMap`, evaluating `$.nested.2147483648` hits the `Integer` entry first and throws `NumberFormatException` before the matching `Long` key is ever compared.

**The scope turned out wider than the new branch.** `main`'s `Long` branch has the same hazard for a segment beyond `Long.MAX_VALUE`: `JSONPath.eval(map, "$.99999999999999999999")` throws today on `main` when the map has a `Long` key. So this is not only a defect this PR introduced — the pattern was already there, and the same change fixes both.

Two things were wrong with the lazy-parse-inside-the-loop shape:
- which parse runs depends on **which key type the iteration happens to see first**, so the outcome varies with hash order for the same data;
- a parse failure aborts the whole evaluation instead of meaning "this entry does not match", and for `$ref` resolution the segment text comes from the input document.

**What changed.** Parse once, outside the loop, into the widest type, and derive the `Integer` candidate only when it is in range:

```java
Long longValue = null;
Integer intValue = null;
if (IOUtils.isNumber(name)) {
    try {
        long num = Long.parseLong(name);
        longValue = num;
        if (num >= Integer.MIN_VALUE && num <= Integer.MAX_VALUE) {
            intValue = (int) num;
        }
    } catch (NumberFormatException ignored) {
        // out of the long range, no boxed numeric key can match
    }
}
```

A candidate left as `null` simply never matches, since `entryKey.equals(null)` is `false` — no extra null checks needed in the loop, and the order dependency is gone.

**Tests.** `RefTest8#test_jsonpath_numericKeyOutOfIntRange` covers the ordered mixed Integer/Long case in both the two-segment (`$.nested.2147483648`) and single-segment (`$.2147483648`) form, plus an out-of-long-range segment that must resolve to null rather than throw. Verified it turns red with `NumberFormatException` when the parse is moved back inside the loop.

已在 5425ac9 修复。

**完全按描述复现。** `LinkedHashMap` 中放入 `{1 -> "int-value", 2147483648L -> "long-match"}`，求值 `$.nested.2147483648` 会先遇到 `Integer` entry 并抛出 `NumberFormatException`，那个匹配的 `Long` key 根本没机会参与比较。

**实际影响面比新增分支更广。** `main` 上的 `Long` 分支对超过 `Long.MAX_VALUE` 的路径段有同样隐患：当 map 含 `Long` key 时，`JSONPath.eval(map, "$.99999999999999999999")` 在 `main` 上今天就会抛异常。所以这不只是本 PR 引入的缺陷——原有代码就存在同样的模式，本次改动一并修掉。

「循环内惰性解析」这种写法有两个问题：
- 执行哪种解析取决于**迭代恰好先遇到哪种类型的 key**，同一份数据的结果会随 hash 顺序漂移；
- 解析失败会中断整个求值，而不是表示「这一项不匹配」；而在 `$ref` 解析场景下，路径段文本来自输入文档。

**改法。** 在循环外用最宽类型解析一次，只有落在 int 范围内才构造 `Integer` 候选（代码见上）。候选保持为 `null` 时自然永不命中，因为 `entryKey.equals(null)` 恒为 `false`——循环里不需要额外判空，顺序依赖也随之消失。

**测试。** `RefTest8#test_jsonpath_numericKeyOutOfIntRange` 覆盖有序的 Integer/Long 混合 key 场景，包含双段（`$.nested.2147483648`）与单段（`$.2147483648`）两种形式，以及一个超出 long 范围、必须解析为 null 而非抛异常的路径段。已验证把解析移回循环内后该测试以 `NumberFormatException` 变红。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONPathSegmentName.java
1919chichi: Fixed in 5425ac9.

**Confirmed.** A map holding both `"1" -> null` and `1 -> "integer-value"`, evaluated at `$.nested.1`, returned `integer-value`. The cause is that `value == null` has always been used as the "key is absent" test, which conflates "no such key" with "the key maps to null".

**For the record, this is not new in this PR.** `main` behaves the same way for a `Long` key (`{"1" -> null, 1L -> ...}`) — the Enum and Long branches sit behind the same `value == null` check. This PR only widened it to `Integer`. Since the change touches this exact code, it seemed better to fix the semantics than to leave a known-wrong branch behind, so the fix covers all three key kinds.

**What changed.** The guard is exactly as suggested — the fallback runs only when the key is genuinely absent:

```java
Object value = map.get(name);
if (value != null || map.containsKey(name)) {
    return value;
}
```

One implementation note: the `containsKey` call has to sit inside the same `try` as `get`. A comparator-based map rejects the String segment on *either* call (see R1-4), so guarding only `get` would just move the `ClassCastException` one line down.

**Tests.** `RefTest8#test_jsonpath_exactStringKeyMappedToNull_notShadowed` pins null for both `$.nested.1` (`JSONPathSegmentName`) and `$.1` (`JSONPathSingleName`), so the semantics is fixed in both evaluators. Verified it turns red (`expected <null> but was <integer-value>`) when `|| map.containsKey(name)` is dropped.

已在 5425ac9 修复。

**已确认。** 同时包含 `"1" -> null` 和 `1 -> "integer-value"` 的 Map，求值 `$.nested.1` 返回了 `integer-value`。根因是一直用 `value == null` 作为「key 不存在」的判据，把「没有这个 key」和「key 映射到 null」混为一谈了。

**需要说明：这不是本 PR 新引入的。** `main` 上对 `Long` key（`{"1" -> null, 1L -> ...}`）行为完全相同——Enum 和 Long 分支都在同一个 `value == null` 判断之后。本 PR 只是把它扩展到了 `Integer`。既然改动正好落在这段代码上，与其留下一个已知错误的分支，不如把语义修对，所以修复覆盖了全部三类 key。

**改法。** 保护逻辑与建议完全一致——只有 key 确实不存在时才走兜底（代码见上）。

一个实现细节：`containsKey` 必须和 `get` 放在同一个 `try` 里。比较型 Map 对这**两个**调用都会拒收 String 段名（见 R1-4），只保护 `get` 只会把 `ClassCastException` 往下挪一行。

**测试。** `RefTest8#test_jsonpath_exactStringKeyMappedToNull_notShadowed` 同时固定 `$.nested.1`（`JSONPathSegmentName`）与 `$.1`（`JSONPathSingleName`）都返回 null，两个 evaluator 的语义都已修正。已验证去掉 `|| map.containsKey(name)` 后该测试以 `expected <null> but was <integer-value>` 变红。

</details>

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterImplMap.java
1919chichi: Fixed in f3fe7d2.

**Confirmed.** A root `Map<Integer, Item>` with two keys sharing one value serialized to `{1:{"itemId":400},2:{"$ref":"$.1"}}`, and `parsed.get(2)` came back null after the round trip. Baseline for comparison: on `main` no `$ref` was emitted at all for that map (`strKey` was null, so `setPath` was skipped) and the object was written twice — identity lost but data intact. The null is introduced by this PR.

**The cause is structural, not a missing branch.** `JSONPath.of("$.1")` produces a single segment and is evaluated by `JSONPathSingleName`, which carries its own near-identical copy of the map fallback. Any fix made in `JSONPathSegmentName` is invisible to it, and a single-segment path is exactly what a root-level container produces. Adding an `Integer` branch to the copy would fix this instance and leave the next one — the two copies would drift again the first time a `Short` or `BigInteger` key is supported.

**What changed.** The duplicate is deleted. `JSONPathSingleName#eval` now delegates to the shared helper:

```java
Object value;
if (root instanceof Map) {
    value = JSONPathSegmentName.getMapValue((Map) root, name);
} else {
```

So `Integer` support, the `ClassCastException` guard (R1-4), the exact-null semantics (R1-3) and the range-checked numeric parse (R1-2) all reach single-segment paths at once. It also removed the last use of `IOUtils` in that file.

As a side effect `JSONPathSegmentName#eval` shrank from 949 to 726 bytes of bytecode — the entrySet scan no longer occupies the hot method's inlining budget.

**Tests.** `RefTest8#test_rootMap_integerKey_directShare` asserts the exact `$.1` path text and `assertSame` after the round trip; `#test_rootMap_longKey_directShare` guards the Long case that already worked, so the delegation cannot regress it. Verified the Integer one turns red (`expected: not <null>`) when `JSONPathSingleName` is reverted to its own loop.

已在 f3fe7d2 修复。

**已确认。** 根级 `Map<Integer, Item>` 两个 key 共享一个对象时，序列化为 `{1:{"itemId":400},2:{"$ref":"$.1"}}`，往返后 `parsed.get(2)` 为 null。对照基线：`main` 上这个 map 根本不会生成 `$ref`（`strKey` 为 null，`setPath` 被跳过），对象被写了两遍——identity 丢失但数据完整。这个 null 是本 PR 引入的。

**根因是结构性的，不是漏了一个分支。** `JSONPath.of("$.1")` 只解析出一个段，由 `JSONPathSingleName` 求值，而它自带一份几乎逐行相同的 map 兜底拷贝。在 `JSONPathSegmentName` 里做的任何修复对它都不可见，而根级容器产生的恰好就是单段路径。只给这份拷贝补一个 `Integer` 分支，只能修好这一例、留下下一例——将来一旦支持 `Short` 或 `BigInteger` key，两份拷贝会再次分叉。

**改法。** 删掉重复实现，`JSONPathSingleName#eval` 改为委托给共享 helper（代码见上）。这样 `Integer` 支持、`ClassCastException` 保护（R1-4）、精确 null 语义（R1-3）、带范围检查的数值解析（R1-2）一次性全部覆盖到单段路径。同时该文件中 `IOUtils` 的最后一处引用也被移除。

附带效果：`JSONPathSegmentName#eval` 的字节码从 949 字节降到 726 字节——entrySet 扫描不再占用这个热方法的内联预算。

**测试。** `RefTest8#test_rootMap_integerKey_directShare` 断言精确的 `$.1` 路径文本以及往返后的 `assertSame`；`#test_rootMap_longKey_directShare` 守住本来就能工作的 Long 场景，确保委托改造不会让它回退。已验证把 `JSONPathSingleName` 改回自有循环后，Integer 那条以 `expected: not <null>` 变红。

</details>

1919chichi: Fixed in ca816ee.

**The analysis holds.** `strKey` was computed at the top of the loop body, ahead of the `String` / `Integer` / `Long` / `Boolean` / `BigDecimal` value fast paths that all `continue` without ever reading it. A `Map<Long, Long>` therefore allocated one immediately-garbage String per entry even with reference detection off, and the only consumer was a feature that is disabled by default.

The intended shape is already in this same file: the JSONB branch calls `jsonWriter.setPath(entryKey.toString(), value)` *inside* the reference-detection branch. The text branch now matches it.

**What changed.** `writeMapKey` is reverted to its `main` form and the path String is derived where it is consumed:

```java
boolean valueRefDetect = refDetect && !isPrimitiveOrEnum
        && (strKey != null || key instanceof Integer || key instanceof Long);
if (valueRefDetect) {
    if (value == object) {
        jsonWriter.writeReference("..");
        continue;
    }

    if (strKey == null) {
        strKey = key.toString();
    }

    String refPath = jsonWriter.setPath(strKey, value);
```

Effects:

- **Reference detection off:** two `instanceof` checks behind a short-circuited `refDetect &&`, zero allocation. `Map<Long, Long>` never reaches this line at all — it returns at the primitive fast path above.
- **Reference detection on:** `toString()` runs only for entries whose value is a non-primitive object, i.e. exactly the entries that get a path pushed onto the stack.
- `git diff main` on this file is now this single hunk — `writeMapKey` is byte-for-byte identical to `main`. That also means the `fieldName` handed to `valueWriter.write(...)` at the `Long` + `userDefineMask` fast path stays `null` for numeric keys just as on `main`, so no incidental behavior change slips in there.

**Verification.** `mvn test -pl core` passes in full (7984 tests), and `RefTest8#test_longKeyedMap_noSharing_outputUnchanged` pins that the emitted JSON text is unchanged when there is nothing to reference.

已在 ca816ee 修复。

**分析成立。** `strKey` 在循环体开头就被求值，位于 `String` / `Integer` / `Long` / `Boolean` / `BigDecimal` 这些 value 快速路径之前，而这些分支全部 `continue` 掉、从不读取它。因此 `Map<Long, Long>` 即使关闭引用检测，每个 entry 也会分配一个立即变成垃圾的 String，而唯一的消费方是一个默认关闭的特性。

正确的写法在本文件中已有先例：JSONB 分支是在引用检测分支**内部**调用 `jsonWriter.setPath(entryKey.toString(), value)` 的。现在文本分支与之对齐。

**改法。** `writeMapKey` 恢复为 `main` 的原样，路径字符串改在实际消费处求值（代码见上）。

效果：

- **关闭引用检测：** 在短路的 `refDetect &&` 之后只多两次 `instanceof`，零分配。`Map<Long, Long>` 甚至根本走不到这一行——它在上方的 primitive 快速路径就返回了。
- **开启引用检测：** `toString()` 只对 value 是非 primitive 对象的 entry 执行，也就是恰好会被压入路径栈的那些 entry。
- 该文件相对 `main` 的 `git diff` 现在只剩这一处 hunk——`writeMapKey` 与 `main` 逐字节一致。这同时意味着 `Long` + `userDefineMask` 快速路径传给 `valueWriter.write(...)` 的 `fieldName` 对数值 key 仍与 `main` 一样是 `null`，不会在那里混入意外的行为变化。

**验证。** `mvn test -pl core` 全量通过（7984 个测试），`RefTest8#test_longKeyedMap_noSharing_outputUnchanged` 固定了「没有共享对象时输出 JSON 文本完全不变」。

</details>

1919chichi:
### Question: should `$ref` support map keys beyond String/Integer/Long/Enum?

While verifying the review fixes I checked whether the same class of defect exists for other key types. It does, but it is **pre-existing on `main`**, not introduced by this PR — I confirmed by reverting the three main-source files to `main` and re-running the same probe, which produced identical results. So I would rather ask before widening this PR.

**What I measured** — one shared object referenced from two map entries, round-tripped, checking whether the `$ref` resolves back to the same instance:

| key type | JSON (default) | JSON + WriteNonStringKeyAsString | JSONB |
| --- | --- | --- | --- |
| `Integer` / `Long` | OK | OK | OK |
| `Enum` | no `$ref` emitted | OK | OK |
| `Short` / `Byte` | no `$ref` emitted | **resolves to null** | **resolves to null** |
| `Character` | no `$ref` emitted | **resolves to null** | **resolves to null** |
| `BigInteger` / `BigDecimal` | no `$ref` emitted | **resolves to null** | **resolves to null** |
| `Boolean` / `Double` | no `$ref` emitted | **resolves to null** | **resolves to null** |

The cause is that the two ends use different criteria. The write side decides "can this key become a path segment?" with `ObjectWriterProvider.isPrimitiveOrEnum` — in the JSONB branch (`ObjectWriterImplMap:331`) and in `isWriteAsString` (`:394`). The read-side fallback only knows `Enum` / `Long` / `Integer`. Everything in between produces a path the reader cannot resolve, and the failure is a silent null rather than an exception.

**One detail worth flagging:** the "no `$ref` emitted" column is not harmless. It only looks safe when the *map value itself* is the shared object. In the shape this PR is actually about — the shared object nested one level below, i.e. issue #3516 — the JSON default path emits a path that is missing the key segment:

```
Map<Short, Item> sharing an inner map:
  [{"itemMap":{1:{"reversePriceMap":{"1":1}}}},
   {"itemMap":{2:{"reversePriceMap":{"$ref":"$[0].itemMap.reversePriceMap"}}}}]
                                              ^^^ key segment missing -> resolves to null
```

That is exactly the #3516 symptom, with `Short` instead of `Long`. It cannot be fixed on the read side alone, because the segment was never pushed onto the path stack.

**Options as I see them:**

1. **Leave as is.** Non-String/Integer/Long/Enum keys keep losing shared-object identity, silently.
2. **Read side only** — extend the fallback to `isPrimitiveOrEnum`. Fixes the WriteNonStringKeyAsString and JSONB columns; the JSON default column stays broken for the nested shape above.
3. **Both sides, symmetric on `isPrimitiveOrEnum`.** Fixes everything. The write side change is a behavior change: a `Map<Short, V>` with reference detection on starts emitting `$ref` where it previously duplicated the object.

I prototyped option 3 locally to check feasibility — 6 lines of main source, all 30 combinations (10 key types × 3 write paths) pass, custom-object keys still emit no `$ref` so the boundary is not widened, and the full `mvn test -pl core` suite (7992 tests) is green. I have **not** included it here.

**Would you like this addressed, and if so, which option?** My suggestion is option 3 in a separate PR, keeping this one focused on #3516. Happy to open an issue with the probe code and the matrix above, or to fold it in here if you prefer.

### 提问：`$ref` 是否需要支持 String/Integer/Long/Enum 之外的 map key？

在验证本轮 review 修复时，我顺带检查了同类缺陷是否存在于其它 key 类型。结论是存在，但**在 `main` 上就已存在**，并非本 PR 引入——我把三个主代码文件还原成 `main` 后重跑同一份探测，结果完全一致。所以想先征求意见，再决定是否扩大本 PR 的范围。

**实测数据**：一个共享对象被两个 map entry 引用，往返后检查 `$ref` 能否解析回同一实例（矩阵见上）。

根因是两端判据不一致。写端用 `ObjectWriterProvider.isPrimitiveOrEnum` 决定「这个 key 能否作为路径段」——JSONB 分支（`ObjectWriterImplMap:331`）和 `isWriteAsString`（`:394`）都是如此；而读端兜底只认 `Enum` / `Long` / `Integer`。两者的差集会生成读端解析不了的路径，且失败形式是**静默变 null**，不抛异常。

**一个需要特别说明的点**：表中「不生成 `$ref`」并非无害。只有当**共享对象就是 map value 本身**时才看似安全。在本 PR 真正针对的形态下——共享对象嵌套在下一层，也就是 issue #3516 的场景——JSON 默认通路会生成一个缺少 key 段的路径（示例见上），正是 #3516 的症状，只是 key 类型从 `Long` 换成了 `Short`。这一条**无法只靠读端修复**，因为那一层从未被压入路径栈。

**我认为的几个选项：**

1. **维持现状**：非 String/Integer/Long/Enum 的 key 继续静默丢失共享对象的 identity。
2. **只改读端**：把兜底判据扩展为 `isPrimitiveOrEnum`。可修复 WriteNonStringKeyAsString 与 JSONB 两列；JSON 默认列在上述嵌套形态下仍然是坏的。
3. **两端对称，统一用 `isPrimitiveOrEnum`**：全部修复。写端属于行为变更——开启引用检测的 `Map<Short, V>` 会开始生成 `$ref`，而此前是重复写出对象。

我在本地对选项 3 做了原型验证：主代码 6 行，30 个组合（10 种 key 类型 × 3 条写通路）全部通过，自定义对象 key 仍不生成 `$ref`（边界未被放宽），`mvn test -pl core` 全量 7992 个测试通过。这部分**没有**包含在本 PR 中。

**是否需要处理？如果需要，倾向哪个选项？** 我的建议是按选项 3 单开一个 PR，让本 PR 保持聚焦于 #3516。也可以先开一个 issue 附上探测代码和上面这张矩阵，或者按您的意见直接并入本 PR。

</details>

PR #7798, opened 2026-08-21
fix: use BigDecimal.valueOf for Double and Float values in getBigDecimal

### LINKED ISSUE #7794: [BUG]fastjson2 2.0.64 使用getBigDecimal存在精度丢失问题
### 问题描述
fastjson2 2.0.64 使用getBigDecimal存在精度丢失问题

### 环境信息
 - OS信息：  Mac Tahoe 26.1
 - JDK信息：OpenJDK 21.0.7
 - 版本信息：2.0.64
 

### 重现步骤
        JSONObject value = new JSONObject().fluentPut("value", 0.00054797);
        System.out.println(value.getBigDecimal("value"));
}
输出：0.0005479699999999999，与预期0.00054797不符。

waterWang:
Closes #7794.

## Problem
`JSONObject().fluentPut("value", 0.00054797).getBigDecimal("value")` returns `0.0005479699999999999` instead of `0.00054797`.

## Root cause
`JSONObject.getBigDecimal(String)` and `JSONArray.getBigDecimal(int)` call `TypeUtils.toBigDecimal(double)` for Double values, which uses the custom `NumberUtils.writeDouble` → `parseBigDecimal` path. The custom `writeDouble` implementation does not guarantee the same rounding as the JDK's shortest-representation algorithm, so `0.00054797` becomes `0.0005479699999999999`.

## Fix
- **Double values**: `return BigDecimal.valueOf(doubleValue)` — `BigDecimal.valueOf(double)` internally uses `Double.toString(double)`, which the JDK guarantees to produce the shortest representation that can reconstruct the original value.
- **Float values**: `return new BigDecimal(Float.toString(floatValue))` — same rationale; `BigDecimal.valueOf(float)` would widen to `double` and lose the float's own short representation (verified: `BigDecimal.valueOf(0.00054797f)` yields `0.0005479700048454106`).

## Verification
```java
JSONObject value = new JSONObject().fluentPut("value", 0.00054797);
System.out.println(value.getBigDecimal("value"));
// Before: 0.0005479699999999999
// After:  0.00054797
```

Added regression tests `testGetBigDecimalPrecision` and `testGetBigDecimalPrecisionFloat` in `JSONObjectTest`. Both pass with the fix and fail without it.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
@@ -748,7 +748,7 @@ public class JSONArray
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return BigDecimal.valueOf(floatValue);
+                return new BigDecimal(Float.toString(floatValue));
             }
 
             if (value instanceof Double) {
--- a/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
@@ -748,12 +748,12 @@ public class JSONArray
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return toBigDecimal(floatValue);
+                return BigDecimal.valueOf(floatValue);
             }
 
             if (value instanceof Double) {
                 double doubleValue = (Double) value;
-                return toBigDecimal(doubleValue);
+                return BigDecimal.valueOf(doubleValue);
             }
 
             long longValue = ((Number) value).longValue();
--- a/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
@@ -914,12 +914,12 @@ public class JSONObject
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return toBigDecimal(floatValue);
+                return BigDecimal.valueOf(floatValue);
             }
 
             if (value instanceof Double) {
                 double doubleValue = (Double) value;
-                return toBigDecimal(doubleValue);
+                return BigDecimal.valueOf(doubleValue);
             }
 
             long longValue = ((Number) value).longValue();
--- a/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
@@ -914,7 +914,7 @@ public class JSONObject
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return BigDecimal.valueOf(floatValue);
+                return new BigDecimal(Float.toString(floatValue));
             }
 
             if (value instanceof Double) {
--- a/core/src/test/java/com/alibaba/fastjson2/types/JSONObjectTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/types/JSONObjectTest.java
@@ -1508,4 +1508,18 @@ public class JSONObjectTest {
         LocalDateTime resultWithDefault = jsonObject.getLocalDateTime("nullValue", defaultDateTime);
         assertEquals(defaultDateTime, resultWithDefault);
     }
+
+    @Test
+    public void testGetBigDecimalPrecision() {
+        JSONObject value = new JSONObject().fluentPut("value", 0.00054797);
+        java.math.BigDecimal result = value.getBigDecimal("value");
+        assertEquals(new java.math.BigDecimal("0.00054797"), result);
+    }
+    
+    @Test
+    public void testGetBigDecimalPrecisionFloat() {
+        JSONObject value = new JSONObject().fluentPut("value", 0.00054797f);
+        java.math.BigDecimal result = value.getBigDecimal("value");
+        assertEquals(new java.math.BigDecimal("0.00054797"), result);
+    }
 }


core/src/main/java/com/alibaba/fastjson2/JSONObject.java
```diff
@@ -914,12 +914,12 @@ public BigDecimal getBigDecimal(String key) {
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return toBigDecimal(floatValue);
+                return new BigDecimal(Float.toString(floatValue));
```
wenshao, line 917: **[Critical]** This fixes the precision case only in the two core containers, while the shared `TypeUtils.toBigDecimal(float/double)` path still returns the old value. For the same Double `0.00054797`, core returns `0.00054797`, but fastjson1-compatible `JSONObject` and `JSONArray` still return `0.0005479699999999999`, so applications using different entry points get inconsistent exact values. Please fix the shared conversion policy centrally and keep all callers delegated to it.

```text
UNMODIFIED PR:
TypeUtils.double=0.0005479699999999999
compat.JSONObject.double=0.0005479699999999999
compat.JSONArray.double=0.0005479699999999999

CENTRAL TypeUtils FIX:
TypeUtils.double=0.00054797
compat.JSONObject.double=0.00054797
compat.JSONArray.double=0.00054797
Tests run: 1, Failures: 0
```

Please add TypeUtils plus core and fastjson1-compatible container tests and verify that removing the central fix makes them fail.

wenshao, line 917: **[Critical]** This exact conversion now depends on the JDK's shortest-decimal algorithm. The same raw Float produces `11884683100000` on JDK 8/11/17 and `11884683000000` on JDK 21/25; persisted values, equality, hashes, or signatures can therefore change after a runtime upgrade. Use the library's corrected deterministic formatter rather than delegating exact decimal selection to `Float.toString`.

```text
JDK 8/11/17: floatBigDecimal=11884683100000
JDK 21/25:   floatBigDecimal=11884683000000
```

Please add the raw-bit Float/Double cases to both container tests and verify the supported JDK matrix disagrees when the deterministic fix is removed.

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
```diff
@@ -914,12 +914,12 @@ public BigDecimal getBigDecimal(String key) {
...
             }
 
             if (value instanceof Double) {
                 double doubleValue = (Double) value;
-                return toBigDecimal(doubleValue);
+                return BigDecimal.valueOf(doubleValue);
```
wenshao, line 922: **[Critical]** `BigDecimal.valueOf(double)` also delegates exact decimal selection to the executing JDK. For `Double.longBitsToDouble(0xc37781a83875ea81L)`, this path returns `-105863743457372176` on JDK 8/11/17 but `-105863743457372180` on JDK 21/25. Use the corrected deterministic shared formatter so a runtime upgrade cannot change the exact `BigDecimal`.

```text
JDK 8/11/17: doubleBigDecimal=-105863743457372176
JDK 21/25:   doubleBigDecimal=-105863743457372180
```

Please add the raw-bit Float/Double cases to both container tests and verify the supported JDK matrix disagrees when the deterministic fix is removed.

core/src/main/java/com/alibaba/fastjson2/JSONArray.java
```diff
@@ -748,12 +748,12 @@ public BigDecimal getBigDecimal(int index) {
 
             if (value instanceof Float) {
                 float floatValue = (Float) value;
-                return toBigDecimal(floatValue);
+                return new BigDecimal(Float.toString(floatValue));
```
wenshao, line 751: **[Critical]** This exact conversion now depends on the JDK's shortest-decimal algorithm. The same raw Float produces `11884683100000` on JDK 8/11/17 and `11884683000000` on JDK 21/25; the array API therefore changes exact values across supported runtimes. Route this branch through the corrected deterministic shared formatter.

```text
JDK 8/11/17: floatBigDecimal=11884683100000
JDK 21/25:   floatBigDecimal=11884683000000
```

Please add the raw-bit Float/Double cases to both container tests and verify the supported JDK matrix disagrees when the deterministic fix is removed.

core/src/main/java/com/alibaba/fastjson2/JSONArray.java
```diff
@@ -748,12 +748,12 @@ public BigDecimal getBigDecimal(int index) {
...
             }
 
             if (value instanceof Double) {
                 double doubleValue = (Double) value;
-                return toBigDecimal(doubleValue);
+                return BigDecimal.valueOf(doubleValue);
```
wenshao, line 756: **[Critical]** `BigDecimal.valueOf(double)` makes this array conversion JDK-dependent as well. `Double.longBitsToDouble(0xc37781a83875ea81L)` yields `-105863743457372176` on JDK 8/11/17 and `-105863743457372180` on JDK 21/25. Use the corrected deterministic shared formatter so exact results stay stable across the supported runtime matrix.

```text
JDK 8/11/17: doubleBigDecimal=-105863743457372176
JDK 21/25:   doubleBigDecimal=-105863743457372180
```

Please add the raw-bit Float/Double cases to both container tests and verify the supported JDK matrix disagrees when the deterministic fix is removed.

core/src/test/java/com/alibaba/fastjson2/types/JSONObjectTest.java
```diff
@@ -1508,4 +1508,18 @@ public void testGetLocalDateTimeWithNullValue() {
         LocalDateTime resultWithDefault = jsonObject.getLocalDateTime("nullValue", defaultDateTime);
         assertEquals(defaultDateTime, resultWithDefault);
     }
+
+    @Test
+    public void testGetBigDecimalPrecision() {
+        JSONObject value = new JSONObject().fluentPut("value", 0.00054797);
+        java.math.BigDecimal result = value.getBigDecimal("value");
+        assertEquals(new java.math.BigDecimal("0.00054797"), result);
+    }
```
wenshao, line 1517: **[Critical]** The blank line immediately after this test contains trailing whitespace, so Checkstyle stops every Maven package, compile, and test invocation before compilation or tests begin. Remove the spaces so the line is completely empty.

```text
JSONObjectTest.java:1518 RegexpMultiline: Line has trailing whitespace
./mvnw -q -DskipTests package: exit 1
./mvnw compile: exit 1
./mvnw test -q: exit 1
```

core/src/main/java/com/alibaba/fastjson2/JSONArray.java
wenshao, line 751: **[Suggestion]** The PR changes both Float and Double branches in `JSONArray.getBigDecimal`, but the new precision-sensitive tests exercise only `JSONObject`. Either array branch can regress while every test added here stays green. Add symmetric `0.00054797f` and `0.00054797d` cases to `JSONArrayTest`.

Please verify each new array test turns red when its corresponding production branch is restored to the previous conversion.

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
wenshao, line 917: **[Suggestion]** This Float path now materializes an intermediate `String` and its backing storage before `BigDecimal` parses it. In record-processing loops, repeated Float-backed `getBigDecimal` calls add allocation and GC pressure compared with the previous fixed-byte-buffer formatter. Correct the allocation-light shared formatter and keep this branch delegated to it.

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
wenshao, line 922: **[Suggestion]** `BigDecimal.valueOf(double)` calls `Double.toString` and then parses that intermediate `String`. Repeated Double-backed `getBigDecimal` calls therefore add allocation and GC pressure compared with the previous fixed-byte-buffer formatter. Correct the allocation-light shared formatter and keep this branch delegated to it.

core/src/main/java/com/alibaba/fastjson2/JSONArray.java
wenshao, line 751: **[Suggestion]** This Float path now materializes an intermediate `String` and its backing storage before `BigDecimal` parses it. High-volume array extraction adds allocation and GC pressure compared with the previous fixed-byte-buffer formatter. Correct the allocation-light shared formatter and keep this branch delegated to it.

core/src/main/java/com/alibaba/fastjson2/JSONArray.java
wenshao, line 756: **[Suggestion]** `BigDecimal.valueOf(double)` calls `Double.toString` and parses the resulting intermediate `String`. High-volume array extraction therefore adds allocation and GC pressure compared with the previous fixed-byte-buffer formatter. Correct the allocation-light shared formatter and keep this branch delegated to it.

wenshao (changes requested):

PR #7800, opened 2026-08-21
fix: preserve declared numeric key type when reading typed Maps from JSONB

### LINKED ISSUE #7799: [BUG] JSONB Map<Long/Integer,V> field loses key type on round-trip (default features)
### 问题描述 / Description

When a field is declared `Map<Long, V>` (or `Map<Integer, V>`) and the enclosing object is serialized to **JSONB** (binary format) with the default `JSONWriter` features — i.e. **without** `JSONWriter.Feature.WriteClassName` — the map's numeric keys come back as `String` instead of `Long`/`Integer` after `JSONB.parseObject(...)`. No shared/circular references or `ReferenceDetection` are involved; a single, unshared entry is enough to reproduce it.

Because the actual stored key is a `String` while the field is declared `Map<Long, Item>`, the two realistic access patterns fail differently:
- `map.get(1L)` silently returns `null` (`Map.get(Object)` doesn't check the declared key type against the argument), leading to a downstream `NullPointerException`.
- `for (Long key : map.keySet())` throws `ClassCastException: class java.lang.String cannot be cast to class java.lang.Long`, because javac inserts an implicit `checkcast` to `Long` for the generically-typed iteration — this can crash ordinary application code with no fastjson2 frames anywhere in the stack trace, making it very hard to diagnose from the caller's side.

The equivalent **text-JSON** round-trip (`JSON.toJSONString` / `JSON.parseObject`) does **not** have this problem, and JSONB **with** `JSONWriter.Feature.WriteClassName` enabled also works correctly — so this is specific to the JSONB codec's default (no `@type`) write path.

### 环境信息 / Environment

- 版本信息 / Version: current `main` branch (also present in released versions; not yet bisected to a specific release)
- JDK: any (reproduced on JDK 17)

### 重现步骤 / Steps to reproduce

```java
import com.alibaba.fastjson2.JSONB;
import java.util.HashMap;
import java.util.Map;

public class Repro {
    public static class Item {
        public Long itemId;
    }
    public static class Bean {
        public Map<Long, Item> map1;
    }

        Item item = new Item();
        item.itemId = 300L;

        Bean bean = new Bean();
        bean.map1 = new HashMap<>();
        bean.map1.put(1L, item);

        byte[] bytes = JSONB.toBytes(bean); // default features, no WriteClassName
        Bean bean2 = JSONB.parseObject(bytes, Bean.class);

        System.out.println(bean2.map1.get(1L)); // expected: Item, actual: null
        for (Long key : bean2.map1.keySet()) {
            System.out.println(key); // throws ClassCastException before printing anything
        }
    }
}
```

### 期待的正确结果 / Expected

`bean2.map1.get(1L)` returns the `Item`, and iterating `map1.keySet()` with the declared `Long` type does not throw — matching the behavior of the text-JSON round-trip.

### 实际输出 / Actual

`bean2.map1.get(1L)` returns `null`; iterating `for (Long key : bean2.map1.keySet())` throws:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Long (java.lang.String and java.lang.Long are in module java.base of loader 'bootstrap')
```

### 补充说明 / Additional notes

This was found while investigating #3516 — it's a separate, unrelated bug (no reference detection or sharing involved) discovered in the same debugging session. I have a local reproduction test and a candidate fix ready; will follow up with a PR.

1919chichi:
## Summary

Fixes #7799.

A field declared `Map<Long, V>` / `Map<Integer, V>` lost its key type when round-tripped through **JSONB** with the default `JSONWriter` features (i.e. without `JSONWriter.Feature.WriteClassName`): the keys came back as `String` instead of `Long`/`Integer`. No shared references or `ReferenceDetection` are involved — a single, unshared entry reproduces it.

Root cause: `ObjectReaderImplMapTyped.readJSONBObject()` decides how to parse a map entry's key purely from whether the current JSONB token happens to be a string token (`jsonReader.isString()`), without ever consulting the field's declared `keyType`:

```java
Object name;
if (keyType == String.class || jsonReader.isString()) {
    name = jsonReader.readFieldName();   // no cast back to keyType
} else {
    ...
}
```

`ObjectWriterImplMap.writeJSONB()` collapses non-`String` keys (e.g. `Long`/`Integer`) to raw string tokens whenever `WriteClassName` is off (the default), so every map key read back through the branch above stayed a `String`, even for a declared `Map<Long, V>` field.

This silently broke `map.get(1L)` (returns `null`, since `Map.get(Object)` never checks the declared key type against its argument) and threw `ClassCastException` from `for (Long key : map.keySet())`, since javac inserts an implicit `checkcast` to the declared generic type on typed iteration — with no fastjson2 frames anywhere in the resulting stack trace.

Two sibling methods in the same class, `readObject()` (the text-JSON path) and `createInstance(Map, ...)`, already handle exactly this situation by casting the parsed key back via `TypeUtils.cast(name, keyType)`. `readJSONBObject()` was missing the equivalent call; this PR adds it.

## Additional fix 1: `JSONReader.read(Map, keyType, valueType, features)` broken for JSONB

While auditing this fix for related gaps, I found a second, unrelated pre-existing bug in a different code path: the public `JSONReader.read(Map object, Type keyType, Type valueType, long features)` API (used when callers read directly into an existing `Map`, not via a bean field) never worked against a JSONB reader at all — it throws `JSONException: UnsupportedOperation`.

The method assumed text-JSON semantics unconditionally:

- it gated entries on `!comma` (a field only ever set by the character-based text readers — always `false` for JSONB, so any map with **more than one entry** threw, regardless of key type),
- it called `nextIfMatch(':')` after reading a non-`String` key (JSONB has no colon token — `JSONReaderJSONB.nextIfMatch()` is `throw new JSONException("UnsupportedOperation")`),
- and it called `nextIfComma()` unconditionally at the end (same problem).

Fix: guard each of these four text-only operations with `!jsonb`, the same convention `ObjectReaderImplMapTyped` already relies on — `nextIfObjectStart()` / `nextIfObjectEnd()` / `readFieldName()` already work correctly for both formats, so JSONB simply skips the comma/colon bookkeeping that text-JSON needs.

## Additional fix 2: parameterized Map value types never reach the JSONB path

Enabling JSONB on the API above exposed the next layer down. `ObjectReaderImplMapTyped.readObject()` lacked the `if (jsonReader.jsonb) return readJSONBObject(...)` dispatch that all three sibling readers already have (`ObjectReaderImplMap`, `ObjectReaderImplMapString`, `ObjectReaderImplList`), so a parameterized `valueType` such as `Map<String, Integer>` resolved to this reader and fell into its text-only path:

```
com.alibaba.fastjson2.JSONException: UnsupportedOperation
	at JSONReaderJSONB.nextIfComma(JSONReaderJSONB.java:271)
	at ObjectReaderImplMapTyped.readObject(ObjectReaderImplMapTyped.java:437)
	at JSONReader.read(JSONReader.java:3654)
```

`ObjectReaderImplMapString` extends this class and already carried its own copy of the dispatch; adding it to the base class makes all four consistent.

## Additional fix 3: `NullPointerException` on a JSONB map with a null key (read side)

A `HashMap` may legitimately hold a null key, and JSONB encodes it as `BC_NULL`, which `readFieldName()` / `readInt64()` return as `null`. `JSONReader.read(Map, ...)` then dereferenced it in `name.equals("@type")` and threw `NullPointerException` whenever `SupportAutoType` was set. For non-`String` key types this was previously masked because the unguarded `nextIfMatch(':')` threw first, so fix 1 above is what makes the line reachable. Flipped to the null-safe `"@type".equals(name)`.

## Additional fix 4: the JSONB **writer** throws `NullPointerException` on null map keys

Serialization and deserialization are one feature, so the write side of the same null-key case is fixed here rather than deferred. `ObjectWriterImplMap.writeJSONB()` dereferenced the entry key in two places without a null check, while the text path has handled it all along (`writeMapKey()` checks `key == null` first):

- **`entryKey.getClass()`** on the `value == null && WriteNulls` path.
- **`entryKey.toString()`** on the main path. This one is a branch-ordering bug: an `entryKey == null` branch already existed, but it sat *third*, behind a catch-all whose second half is unconditionally true while `WriteClassName` is off (the default), making the null branch dead code:

```java
if (entryKey instanceof String || (contextFeatures & Feature.WriteClassName.mask) == 0) {
    ...
    key = entryKey.toString();   // NPE
} else if (entryKey == null) {
    jsonWriter.writeNull();      // correct, but unreachable by default
}
```

Measured before the fix:

```
[OK   ] text  JSON.toJSONString({null:"v"})            = {"null":"v"}
[OK   ] text  + WriteNulls  {null:null}                = {"null":null}
[THROW] jsonb JSONB.toBytes({null:"v"})  default        -> NPE @ ObjectWriterImplMap.writeJSONB:268
[OK   ] jsonb + WriteClassName            {null:"v"}
[THROW] jsonb + WriteClassName+WriteNulls {null:null}   -> NPE @ ObjectWriterImplMap.writeJSONB:238
[OK   ] jsonb + WriteClassName            {1L:"v"}      (control, non-null key)
```

Fix: hoist the `entryKey == null` check ahead of the catch-all and add the same check on the `WriteNulls` path. The `else if (entryKey == null)` branch that used to sit third is deleted because hoisting the identical check makes it unreachable, not because null stopped being handled — the `WriteClassName` case still executes the same `jsonWriter.writeNull()`, just from the first branch instead of the third. Verified byte-for-byte on `HashMap{null: "v"}`:

| features | before | after |
| --- | --- | --- |
| `WriteClassName` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5` (identical) |
| `WriteClassName` + `ReferenceDetection` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5` (identical) |
| `WriteClassName`, `@type` emitted | `92 4a 4d 00 a6 af 4a 76 a5` | `92 4a 4d 00 a6 af 4a 76 a5` (identical) |
| default (no `WriteClassName`) | NPE | `a6 af 4a 76 a5` |
| `WriteClassName` + `WriteNulls`, null value | NPE | `a6 af af a5` |

(`a6` = `BC_OBJECT`, `af` = `BC_NULL`, `4a 76` = 1-char ASCII string `"v"`, `a5` = `BC_OBJECT_END`.) Only the rows that used to throw changed. Both sites now emit `BC_NULL`, which reads back as a genuine `null` key — strictly better than the text path, which writes the string `"null"`. No compatibility risk: these two paths previously threw, so no existing payload depends on the old behaviour. `writeJSONB()` is ~1160 bytes of bytecode and already far past the 325-byte C2 inlining threshold, so the two added branches do not change its inlining behaviour.

## Test plan

- [x] Added `MapLongKeyTypeTest` covering: `Map<Long,V>` and `Map<Integer,V>` round-tripped through JSONB with default features (both `map.get(key)` and typed `keySet()` iteration), plus two control cases showing the equivalent text-JSON round-trip and JSONB-with-`WriteClassName` round-trip were already correct.
- [x] Added `JSONReaderTest.readMap_jsonb_stringKey` / `readMap_jsonb_nonStringKey` covering `JSONReader.read(Map, keyType, valueType, features)` against a JSONB reader for both String and non-String key types with multiple entries.
- [x] Added `JSONReaderTest.readMap_jsonb_parameterizedValueType` (fix 2) — `Map<Long, Map<String, Integer>>` read via `read(map, Long.class, parameterizedValueType, 0L)`; asserting through `map.get(123L)` pins the outer key type (`Long`, not `String`) and the inner value type (`Integer`, not `String`) in one assertion.
- [x] Added `JSONReaderTest.readMap_jsonb_nullKey_autoType` (fix 3) — null key read with `SupportAutoType`, asserting `map.get(null)`.
- [x] Added `MapNullKeyTest` (fix 4) — six JSONB round-trips of a null-keyed map: default features, `WriteClassName`, `WriteNulls` with a null value, `WriteNulls` + `WriteClassName`, a typed `Map<Long, String>` bean field, and a non-null-key control. Four go RED without the writer fix (two per NPE site), the other two are controls that already passed.
- [x] Verified every new test fails without its corresponding fix and passes with it, reverting each fix individually.
- [x] `mvn test` across the whole reactor: 9686 tests, 0 failures, 0 errors (7988 of them in `core`).
- [x] `mvn validate` (checkstyle) passes for all modules.

<details>
<summary>中文说明</summary>

修复 #7799。

字段声明为 `Map<Long, V>` / `Map<Integer, V>`，用默认配置（未开启 `JSONWriter.Feature.WriteClassName`）通过 **JSONB** 序列化再反序列化后，key 会变成 `String` 而不是 `Long`/`Integer`。不涉及共享引用或 `ReferenceDetection`，一个不共享的单条 entry 就能复现。

根因：`ObjectReaderImplMapTyped.readJSONBObject()` 判断怎么解析 key，只看当前 JSONB token 是不是字符串 token（`jsonReader.isString()`），完全没用到字段声明的 `keyType`。而 `ObjectWriterImplMap.writeJSONB()` 在 `WriteClassName` 关闭（默认）时会把非 `String` 类型的 key（比如 `Long`/`Integer`）一律降级写成裸字符串 token，导致读端走进上面这条分支时，即使字段声明是 `Map<Long, V>`，读出来的 key 也一直是 `String`。

这会导致 `map.get(1L)` 静默返回 `null`（`Map.get(Object)` 不检查声明的 key 类型），以及 `for (Long key : map.keySet())` 直接抛 `ClassCastException`（javac 对泛型类型迭代插入的隐式 `checkcast` 失败）——异常堆栈里完全看不到 fastjson2 的痕迹，很难从调用方定位。

同一个类里另外两处方法——`readObject()`（文本 JSON 路径）和 `createInstance(Map, ...)`——已经通过 `TypeUtils.cast(name, keyType)` 处理了同样的情况，唯独 `readJSONBObject()` 少了这一步。本次改动补上这个调用。

**追加修复 1**：`JSONReader.read(Map, keyType, valueType, features)` 对 JSONB 完全不可用。这是复查这个问题时顺带发现的另一个独立、预先存在的 bug，出在一条完全不同的代码路径上——公开 API `JSONReader.read(Map object, Type keyType, Type valueType, long features)`（供调用方直接读进一个已有的 `Map`，不经过 bean 字段）。这个方法从头到尾按文本 JSON 的语义写的：用 `comma` 字段判断 entry 之间有没有分隔符（这个字段只有基于字符的文本 reader 会设置，JSONB reader 上永远是 `false`，所以只要 map 有 **两个以上 entry** 就会抛异常，跟 key 类型无关）；读完非 `String` key 后无条件调用 `nextIfMatch(':')`（JSONB 没有冒号这个概念，`JSONReaderJSONB.nextIfMatch()` 直接 `throw new JSONException("UnsupportedOperation")`）；循环结束后又无条件调用一次 `nextIfComma()`（同样的问题）。修复方式是给这四处纯文本语义的操作加上 `!jsonb` 判断——`nextIfObjectStart()` / `nextIfObjectEnd()` / `readFieldName()` 本身对两种格式都是对的，JSONB 只是不需要逗号/冒号这套记账逻辑，直接跳过即可。

**追加修复 2**：参数化 Map 的 value 类型走不到 JSONB 分支。上面那个 API 打通 JSONB 之后暴露了下一层问题：`ObjectReaderImplMapTyped.readObject()` 缺少三个兄弟 reader（`ObjectReaderImplMap`、`ObjectReaderImplMapString`、`ObjectReaderImplList`）都有的 `if (jsonReader.jsonb) return readJSONBObject(...)` 分发，因此 `Map<String, Integer>` 这类参数化 `valueType` 解析到本 reader 后会掉进文本专用路径，在 `nextIfComma()` 上抛 `UnsupportedOperation`。`ObjectReaderImplMapString` 继承本类且自带一份同样的分发，补到基类后四个类保持一致。

**追加修复 3**：JSONB map 带 null key 时读端 NPE。`HashMap` 允许 null key，JSONB 会把它编码成 `BC_NULL`，`readFieldName()` / `readInt64()` 读回来就是 `null`；`JSONReader.read(Map, ...)` 随后在 `name.equals("@type")` 上解引用，只要开启 `SupportAutoType` 就抛 `NullPointerException`。非 `String` key 类型此前被未加守卫的 `nextIfMatch(':')` 先行抛异常掩盖，正是上面的追加修复 1 让这一行变得可达。改为 null 安全的 `"@type".equals(name)`。

**追加修复 4**：JSONB **写端** 对 null key 抛 `NullPointerException`。序列化和反序列化是一体的，因此同一个 null key 场景的写端一并在本 PR 修掉，不另行推迟。`ObjectWriterImplMap.writeJSONB()` 有两处未加判空就解引用 entry key，而文本路径一直是处理了的（`writeMapKey()` 第一句就判 `key == null`）：

- **`entryKey.getClass()`**，位于 `value == null && WriteNulls` 路径；
- **`entryKey.toString()`**，位于主路径。这一处是分支顺序 bug：`entryKey == null` 分支其实已经写了，但它排在**第三位**，前面挡着一个兜底条件，而该条件的后半段在 `WriteClassName` 关闭（默认）时恒为 true，导致 null 分支成为死代码：

```java
if (entryKey instanceof String || (contextFeatures & Feature.WriteClassName.mask) == 0) {
    ...
    key = entryKey.toString();   // NPE
} else if (entryKey == null) {
    jsonWriter.writeNull();      // 逻辑正确，但默认配置下不可达
}
```

修复前实测：

```
[OK   ] text  JSON.toJSONString({null:"v"})            = {"null":"v"}
[OK   ] text  + WriteNulls  {null:null}                = {"null":null}
[THROW] jsonb JSONB.toBytes({null:"v"})  默认 feature     -> NPE @ ObjectWriterImplMap.writeJSONB:268
[OK   ] jsonb + WriteClassName            {null:"v"}
[THROW] jsonb + WriteClassName+WriteNulls {null:null}   -> NPE @ ObjectWriterImplMap.writeJSONB:238
[OK   ] jsonb + WriteClassName            {1L:"v"}      （对照组，非 null key）
```

修复方式：把 `entryKey == null` 判断提到兜底条件之前，`WriteNulls` 路径上补同样的判断。原先排在第三位的 `else if (entryKey == null)` 之所以删掉，是因为把同样的判断提前之后它已不可达，而不是不再处理 null——`WriteClassName` 场景执行的仍是同一句 `jsonWriter.writeNull()`，只是从第三分支变成第一分支。已对 `HashMap{null: "v"}` 做字节级比对：

| feature 组合 | 修复前 | 修复后 |
| --- | --- | --- |
| `WriteClassName` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5`（一致） |
| `WriteClassName` + `ReferenceDetection` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5`（一致） |
| `WriteClassName`，写出 `@type` | `92 4a 4d 00 a6 af 4a 76 a5` | `92 4a 4d 00 a6 af 4a 76 a5`（一致） |
| 默认（未开 `WriteClassName`） | NPE | `a6 af 4a 76 a5` |
| `WriteClassName` + `WriteNulls`，值为 null | NPE | `a6 af af a5` |

（`a6` = `BC_OBJECT`，`af` = `BC_NULL`，`4a 76` = 长度 1 的 ASCII 串 `"v"`，`a5` = `BC_OBJECT_END`。）只有原本抛异常的两行发生了变化。两处现在都写出 `BC_NULL`，读回来是真正的 `null` key——严格优于文本路径（文本写的是字符串 `"null"`）。无兼容性风险：这两条路径原本就是抛异常，不存在依赖旧行为的历史数据。`writeJSONB()` 的字节码约 1160 字节，本就远超 C2 的 325 字节内联阈值，新增的两个分支不改变其内联行为。

**测试**：新增 `MapLongKeyTypeTest`，覆盖 `Map<Long,V>`/`Map<Integer,V>` 在默认配置下的 JSONB 往返（`get()` 和带类型的 `keySet()` 遍历两种访问方式），并用两个对照组（文本 JSON、开启 `WriteClassName` 的 JSONB）证明问题只出在这一条代码路径上。新增 `JSONReaderTest.readMap_jsonb_stringKey` / `readMap_jsonb_nonStringKey`，覆盖 `JSONReader.read(Map, keyType, valueType, features)` 在 JSONB 场景下、String key 和非 String key、多条 entry 的情况。新增 `JSONReaderTest.readMap_jsonb_parameterizedValueType`（对应追加修复 2）和 `readMap_jsonb_nullKey_autoType`（对应追加修复 3）。新增 `MapNullKeyTest`（对应追加修复 4），覆盖 null key 的六种 JSONB 往返组合：默认配置、`WriteClassName`、`WriteNulls` + null value、`WriteNulls` + `WriteClassName`、类型化的 `Map<Long, String>` bean 字段，以及非 null key 对照组；去掉写端修复后其中四个变红（两个 NPE 点各两个），另外两个是本来就通过的对照组。已逐个回退每处修复，确认对应测试都会变红、加回后变绿。整个 reactor `mvn test` 通过：9686 个测试，0 失败 0 错误（其中 `core` 7988 个）；`mvn validate` checkstyle 全模块通过。

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Aq75UPgBr25T6xuRX6XW8B

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -3628,7 +3628,7 @@ public abstract class JSONReader
         long contextFeatures = features | context.getFeatures();
 
         for (int i = 0; ; ++i) {
-            if (ch == '/') {
+            if (!jsonb && ch == '/') {
                 skipComment();
             }
 
@@ -3636,7 +3636,7 @@ public abstract class JSONReader
                 break;
             }
 
-            if (i != 0 && !comma) {
+            if (!jsonb && i != 0 && !comma) {
                 throw new JSONException(info());
             }
 
@@ -3646,7 +3646,9 @@ public abstract class JSONReader
                 name = readFieldName();
             } else {
                 name = keyReader.readObject(this, null, null, 0L);
-                nextIfMatch(':');
+                if (!jsonb) {
+                    nextIfMatch(':');
+                }
             }
 
             Object value = valueReader.readObject(this, null, null, 0L);
@@ -3676,7 +3678,9 @@ public abstract class JSONReader
             }
         }
 
-        nextIfComma();
+        if (!jsonb) {
+            nextIfComma();
+        }
     }
 
     public <T> T read(Class<T> type) {
--- a/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderImplMapTyped.java
+++ b/core/src/main/java/com/alibaba/fastjson2/reader/ObjectReaderImplMapTyped.java
@@ -193,6 +193,9 @@ class ObjectReaderImplMapTyped
             Object name;
             if (keyType == String.class || jsonReader.isString()) {
                 name = jsonReader.readFieldName();
+                if (keyType != null && keyType != String.class) {
+                    name = TypeUtils.cast(name, keyType);
+                }
             } else {
                 if (jsonReader.isReference()) {
                     String reference = jsonReader.readReference();
--- a/core/src/test/java/com/alibaba/fastjson2/JSONReaderTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/JSONReaderTest.java
@@ -1069,6 +1069,34 @@ public class JSONReaderTest {
     }
 
     @Test
+    public void readMap_jsonb_stringKey() {
+        Map<String, String> src = new LinkedHashMap<>();
+        src.put("a", "1");
+        src.put("b", "2");
+        byte[] bytes = JSONB.toBytes(src);
+
+        JSONReader jsonReader = JSONReader.ofJSONB(bytes);
+        Map map = new HashMap();
+        jsonReader.read(map, String.class, String.class, 0L);
+        assertEquals("1", map.get("a"));
+        assertEquals("2", map.get("b"));
+    }
+
+    @Test
+    public void readMap_jsonb_nonStringKey() {
+        Map<Long, String> src = new LinkedHashMap<>();
+        src.put(123L, "456");
+        src.put(234L, "567");
+        byte[] bytes = JSONB.toBytes(src);
+
+        JSONReader jsonReader = JSONReader.ofJSONB(bytes);
+        Map map = new HashMap();
+        jsonReader.read(map, Long.class, String.class, 0L);
+        assertEquals("456", map.get(123L));
+        assertEquals("567", map.get(234L));
+    }
+
+    @Test
     public void readArray() {
         String str = "[\"123\",\"456\"]";
         JSONReader jsonReader = JSONReader.of(JSONFactory.createReadContext(), str.getBytes());
--- a/core/src/test/java/com/alibaba/fastjson2/jsonb/basic/MapLongKeyTypeTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/jsonb/basic/MapLongKeyTypeTest.java
@@ -1,0 +1,116 @@
+package com.alibaba.fastjson2.jsonb.basic;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.JSONB;
+import com.alibaba.fastjson2.JSONReader;
+import com.alibaba.fastjson2.JSONWriter;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * A Map field typed as Map&lt;Long, V&gt; / Map&lt;Integer, V&gt; used to round-trip its keys back as
+ * String instead of Long/Integer when serialized to JSONB with the default JSONWriter features (i.e.
+ * without JSONWriter.Feature.WriteClassName). No shared/circular references are involved; this
+ * reproduced even for a single, unshared entry.
+ *
+ * Because the actual stored key was a String while the field is declared Map&lt;Long, Item&gt;, the two
+ * realistic access patterns failed differently:
+ *  - map.get(1L) silently returned null (Map.get(Object) doesn't check the declared key type), and
+ *  - `for (Long key : map.keySet())` threw ClassCastException, because javac inserts an implicit
+ *    checkcast to Long for the generically-typed iteration.
+ */
+@Tag("jsonb")
+public class MapLongKeyTypeTest {
+    public static class Item {
+        public Long itemId;
+    }
+
+    public static class Bean {
+        public Map<Long, Item> map1;
+    }
+
+    public static class BeanInt {
+        public Map<Integer, Item> map1;
+    }
+
+    @Test
+    public void test_longKey_jsonb_default() {
+        Item item = new Item();
+        item.itemId = 300L;
+
+        Bean bean = new Bean();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1L, item);
+
+        byte[] bytes = JSONB.toBytes(bean);
+        Bean bean2 = JSONB.parseObject(bytes, Bean.class);
+
+        assertEquals(300L, bean2.map1.get(1L).itemId);
+        for (Long key : bean2.map1.keySet()) {
+            assertEquals(1L, key);
+        }
+    }
+
+    @Test
+    public void test_integerKey_jsonb_default() {
+        Item item = new Item();
+        item.itemId = 200L;
+
+        BeanInt bean = new BeanInt();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1, item);
+
+        byte[] bytes = JSONB.toBytes(bean);
+        BeanInt bean2 = JSONB.parseObject(bytes, BeanInt.class);
+
+        assertEquals(200L, bean2.map1.get(1).itemId);
+        for (Integer key : bean2.map1.keySet()) {
+            assertEquals(1, key);
+        }
+    }
+
+    @Test
+    public void test_longKey_textJson_control() {
+        // Control case: the equivalent text-JSON round-trip preserves the Long key correctly,
+        // showing this was specific to the JSONB codec rather than Map<Long,V> type handling in general.
+        Item item = new Item();
+        item.itemId = 300L;
+
+        Bean bean = new Bean();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1L, item);
+
+        String json = JSON.toJSONString(bean);
+        Bean bean2 = JSON.parseObject(json, Bean.class);
+
+        assertEquals(300L, bean2.map1.get(1L).itemId);
+        for (Long key : bean2.map1.keySet()) {
+            assertEquals(1L, key);
+        }
+    }
+
+    @Test
+    public void test_longKey_jsonb_writeClassName_control() {
+        // Control case: turning on WriteClassName also preserved the Long key correctly,
+        // showing the default (WriteClassName off) JSONB write path was where the type got lost.
+        Item item = new Item();
+        item.itemId = 300L;
+
+        Bean bean = new Bean();
+        bean.map1 = new HashMap<>();
+        bean.map1.put(1L, item);
+
+        byte[] bytes = JSONB.toBytes(bean, JSONWriter.Feature.WriteClassName);
+        Bean bean2 = JSONB.parseObject(bytes, Bean.class, JSONReader.Feature.SupportAutoType);
+
+        assertEquals(300L, bean2.map1.get(1L).itemId);
+        for (Long key : bean2.map1.keySet()) {
+            assertEquals(1L, key);
+        }
+    }
+}


core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3646,7 +3646,9 @@ public final void read(Map object, Type keyType, Type valueType, long features)
                 name = readFieldName();
             } else {
                 name = keyReader.readObject(this, null, null, 0L);
-                nextIfMatch(':');
+                if (!jsonb) {
+                    nextIfMatch(':');
+                }
             }
 
             Object value = valueReader.readObject(this, null, null, 0L);
```
wenshao, line 3654: **[Critical]** The JSONB enablement of `read(Map, keyType, valueType, features)` is incomplete for parameterized-Map key/value types: `ObjectReaderImplMapTyped.readObject` lacks the `if (jsonReader.jsonb) return readJSONBObject(...)` dispatch its sibling readers have (`ObjectReaderImplMap.java:428`, `ObjectReaderImplMapString.java:21`, `ObjectReaderImplList.java:598`). A `valueType` such as `Map<String, Integer>` resolves to `ObjectReaderImplMapTyped`, whose text-only primitives — the trailing `nextIfComma()` (`ObjectReaderImplMapTyped.java:437`) and `nextIfMatch(':')` for non-String inner keys — both throw `JSONException("UnsupportedOperation")` on the JSONB reader. The two new tests only cover simple String/Long key + String value, so the API appears JSONB-complete while nested typed maps fail. Executed against this commit: JSONB of `Map<Long, Map<String, Integer>>` read via `read(map, Long.class, <Map<String,Integer>>, 0L)` throws `UnsupportedOperation` with stack `JSONReaderJSONB.nextIfComma -> ObjectReaderImplMapTyped.readObject:437 -> JSONReader.read:3654`; adding the one-line dispatch flips all nested cases to OK (verified with the patch applied). Note: pre-PR this API threw for ALL JSONB input, so this is not a regression — but the newly enabled functionality ships incomplete for a commonplace shape.

Suggested fix, at the top of `ObjectReaderImplMapTyped.readObject`:

```java
if (jsonReader.jsonb) {
    return readJSONBObject(jsonReader, fieldType, fieldName, features);
}
```

Fix witness: a `JSONReaderTest` case serializing `Map<Long, Map<String, Integer>>` via `JSONB.toBytes`, read back with `jsonReader.read(map, Long.class, parameterizedValueType, 0L)` asserting the nested entries — it goes RED (`JSONException: UnsupportedOperation`) without the dispatch; please remove the guard again after adding it to confirm the test pins the fix.

本 PR 为 `read(Map, keyType, valueType, features)` 启用的 JSONB 支持，对参数化 Map 类型的 key/value 不完整：`ObjectReaderImplMapTyped.readObject` 缺少兄弟 reader 都有的 `if (jsonReader.jsonb) return readJSONBObject(...)` 分发（`ObjectReaderImplMap.java:428`、`ObjectReaderImplMapString.java:21`、`ObjectReaderImplList.java:598`）。当 `valueType` 是 `Map<String, Integer>` 这类参数化 Map 时会走到 `ObjectReaderImplMapTyped`，其文本专用操作——末尾的 `nextIfComma()`（`ObjectReaderImplMapTyped.java:437`）、非 String 内层 key 的 `nextIfMatch(':')`——在 JSONB reader 上都会抛 `JSONException("UnsupportedOperation")`。新增的两个测试只覆盖简单类型（String/Long key + String value），该 API 看似已完整支持 JSONB，但嵌套类型化 Map 会失败。已在当前提交上运行验证：`Map<Long, Map<String, Integer>>` 的 JSONB 经 `read(map, Long.class, <Map<String,Integer>>, 0L)` 读取时抛 `UnsupportedOperation`（堆栈 `JSONReaderJSONB.nextIfComma -> ObjectReaderImplMapTyped.readObject:437 -> JSONReader.read:3654`）；补上单行分发后所有嵌套场景均通过（已打补丁复验）。说明：本 PR 之前该 API 对所有 JSONB 输入都抛异常，因此不是回归——但新启用的功能对常见数据形态仍不完整。

修复建议（`ObjectReaderImplMapTyped.readObject` 开头）：

```java
if (jsonReader.jsonb) {
    return readJSONBObject(jsonReader, fieldType, fieldName, features);
}
```

修复验证：在 `JSONReaderTest` 新增用例，将 `Map<Long, Map<String, Integer>>` 用 `JSONB.toBytes` 序列化后经 `jsonReader.read(map, Long.class, parameterizedValueType, 0L)` 读回并断言嵌套 entry——缺少该分发时测试以 `JSONException: UnsupportedOperation` 失败；补上后请再次移除该分发确认测试能钉住此修复。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3646,7 +3646,9 @@ public final void read(Map object, Type keyType, Type valueType, long features)
                 name = readFieldName();
             } else {
                 name = keyReader.readObject(this, null, null, 0L);
-                nextIfMatch(':');
+                if (!jsonb) {
+                    nextIfMatch(':');
+                }
```
wenshao, line 3651: **[Suggestion]** With this `!jsonb` guard removing the old barrier, the newly JSONB-enabled path reaches the pre-existing `name.equals("@type")` check (~line 3661) with `name == null` when a JSONB map carries a `BC_NULL` key and `SupportAutoType` is set: `JSONReaderJSONB.readFieldName()`/`readInt64()` return null on `BC_NULL`, and the unguarded dereference throws `NullPointerException`. For non-String keyTypes this was previously masked because the unguarded `nextIfMatch(':')` threw first, so the guard added here is what makes it reachable (the String-keyType variant is pre-existing on that unchanged line). Executed against this commit: a `HashMap<Long, String>` with a null key, serialized with `WriteClassName` + `NotWriteHashMapArrayListClassName`, throws NPE at `JSONReader.java:3661` when read via this method with `SupportAutoType`, while the same input without `SupportAutoType` reads fine (`{null=v}`). Since JSONB can legitimately carry null keys once this API accepts JSONB input, flip the comparison to `"@type".equals(name)` (null-safe). Fix witness: a `JSONReaderTest` case serializing a HashMap with a null key, read via `read(map, Long.class, String.class, Feature.SupportAutoType.mask)` asserting `map.get(null)` — it fails with NPE while `name.equals("@type")` stands.

这个 `!jsonb` 守卫移除了旧屏障之后，新启用的 JSONB 路径会在 JSONB map 带 `BC_NULL` key 且开启 `SupportAutoType` 时，以 `name == null` 走到既有的 `name.equals("@type")` 检查（约 3661 行）：`JSONReaderJSONB.readFieldName()`/`readInt64()` 对 `BC_NULL` 返回 null，未加防护的解引用抛 `NullPointerException`。非 String keyType 场景此前被未加守卫的 `nextIfMatch(':')` 先行抛异常所掩盖，正是本次新增的守卫使其变得可达（String keyType 变体在该未改动行上本就存在）。已在当前提交上运行验证：带 null key 的 `HashMap<Long, String>`，用 `WriteClassName` + `NotWriteHashMapArrayListClassName` 序列化，经本方法以 `SupportAutoType` 读取时在 `JSONReader.java:3661` 抛 NPE；同样输入不开 `SupportAutoType` 则正常读出 `{null=v}`。JSONB 在本 API 接受 JSONB 输入后可以合法携带 null key，建议将比较改为 `"@type".equals(name)`（null 安全）。修复验证：在 `JSONReaderTest` 新增用例，序列化带 null key 的 HashMap，经 `read(map, Long.class, String.class, Feature.SupportAutoType.mask)` 读取并断言 `map.get(null)`——保留 `name.equals("@type")` 时该测试以 NPE 失败。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3628,15 +3628,15 @@ public final void read(Map object, Type keyType, Type valueType, long features)
         long contextFeatures = features | context.getFeatures();
 
         for (int i = 0; ; ++i) {
-            if (ch == '/') {
+            if (!jsonb && ch == '/') {
```
wenshao, line 3631: **[Suggestion]** This guard hunk is defensive-only and ships with nothing gating it: reverting this hunk alone leaves all 56 affected tests green (measured). `JSONReaderJSONB` never assigns `ch`, so the branch is unreachable on JSONB today, and `skipComment()` throws `UnsupportedOperation` — no code change is required, but be aware no test pins this guard: if `JSONReaderJSONB` ever starts maintaining `ch`, a regression removing it would pass the whole suite silently and JSONB map reads would fail at runtime. The other three guard hunks are each pinned — reverting any of them individually turns the new JSONB tests RED.

这个守卫 hunk 纯属防御性，且没有任何测试钉住它：单独回退此 hunk 后全部 56 个相关测试仍为绿（已实测）。`JSONReaderJSONB` 从不给 `ch` 赋值，因此该分支在 JSONB 上目前不可达，且 `skipComment()` 会抛 `UnsupportedOperation`——无需改代码，但请注意没有测试钉住这个守卫：若将来 `JSONReaderJSONB` 开始维护 `ch`，删除该守卫的回归将静默通过全部测试，JSONB map 读取会在运行期失败。其余三个守卫 hunk 各有测试钉住——分别单独回退都会让新增的 JSONB 测试变红。

</details>

wenshao (changes requested):

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
1919chichi: Confirmed and fixed in 33c4cfd.

Reproduced exactly as described on `9dc64b4`, same stack:

```
com.alibaba.fastjson2.JSONException: UnsupportedOperation
	at JSONReaderJSONB.nextIfComma(JSONReaderJSONB.java:271)
	at ObjectReaderImplMapTyped.readObject(ObjectReaderImplMapTyped.java:437)
	at JSONReader.read(JSONReader.java:3654)
```

Applied the suggested dispatch at the top of `ObjectReaderImplMapTyped.readObject`. Since `ObjectReaderImplMapString` extends this class and already carried its own copy of the dispatch, this also makes the base class consistent with all three siblings.

Fix witness added: `JSONReaderTest.readMap_jsonb_parameterizedValueType` — `Map<Long, Map<String, Integer>>` via `JSONB.toBytes`, read back with `read(map, Long.class, parameterizedValueType, 0L)`. Asserting through `map.get(123L)` pins both the outer key type (`Long`, not `String`) and the inner value type (`Integer`, not `String`) in one assertion. Verified RED without the dispatch (`JSONException: UnsupportedOperation`) and GREEN with it.

Regression: `mvn -o test` across the whole reactor — 9686 tests, 0 failures, 0 errors (7988 of them in `core`). `mvn -o validate` clean. `fastjson1-compatible` in particular exercises this reader heavily and is unaffected.

已确认并在 33c4cfd 修复。

在 `9dc64b4` 上完整复现，堆栈与描述一致：

```
com.alibaba.fastjson2.JSONException: UnsupportedOperation
	at JSONReaderJSONB.nextIfComma(JSONReaderJSONB.java:271)
	at ObjectReaderImplMapTyped.readObject(ObjectReaderImplMapTyped.java:437)
	at JSONReader.read(JSONReader.java:3654)
```

已按建议在 `ObjectReaderImplMapTyped.readObject` 开头加上分发。`ObjectReaderImplMapString` 继承本类且自己已经带了一份同样的分发，补到基类后与三个兄弟类全部一致。

新增 fix witness：`JSONReaderTest.readMap_jsonb_parameterizedValueType`——`Map<Long, Map<String, Integer>>` 经 `JSONB.toBytes` 序列化后用 `read(map, Long.class, parameterizedValueType, 0L)` 读回。断言走 `map.get(123L)`，一条断言同时钉住外层 key 类型（`Long` 而非 `String`）和内层 value 类型（`Integer` 而非 `String`）。已验证：去掉分发该测试变红（`JSONException: UnsupportedOperation`），加回后变绿。

回归：`mvn -o test` 覆盖整个 reactor，9686 个测试，0 失败 0 错误（其中 `core` 7988 个）；`mvn -o validate` 通过。其中 `fastjson1-compatible` 大量使用该 reader，未受影响。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
1919chichi: Confirmed and fixed in e46d82e — flipped to `"@type".equals(name)`.

Reproduced on `9dc64b4` with the exact recipe given (`HashMap<Long, String>` with a null key, `WriteClassName` + `NotWriteHashMapArrayListClassName`), and the causal chain holds: with a non-String `keyType` the unguarded `nextIfMatch(':')` used to throw first, so the `!jsonb` guard added by this PR is what makes line 3661 reachable.

Fix witness added: `JSONReaderTest.readMap_jsonb_nullKey_autoType`, asserting `map.get(null)`. Verified RED (`NullPointerException ... "name" is null`) without the fix, GREEN with it.

One note on why the test has to use those two writer features rather than a plain `JSONB.toBytes(src)` — the write side has a related pre-existing defect. `ObjectWriterImplMap.writeJSONB` does have an `entryKey == null` branch, but it sits *third*, behind a catch-all:

```java
if (entryKey instanceof String || (contextFeatures & Feature.WriteClassName.mask) == 0) {
    ...
    key = entryKey.toString();   // line 268 — NPE
} else if (entryKey == null) {
    jsonWriter.writeNull();      // line 281 — correct, but unreachable by default
}
```

With `WriteClassName` off (the default) the second half of the first condition is unconditionally true, so every key is swallowed by the first branch and the null branch is dead code. Measured:

```
[OK   ] text  JSON.toJSONString({null:"v"})            = {"null":"v"}
[OK   ] text  + WriteNulls  {null:null}                = {"null":null}
[THROW] jsonb JSONB.toBytes({null:"v"})  default        -> NPE @ ObjectWriterImplMap.writeJSONB:268
[OK   ] jsonb + WriteClassName            {null:"v"}
[THROW] jsonb + WriteClassName+WriteNulls {null:null}   -> NPE @ ObjectWriterImplMap.writeJSONB:238
[OK   ] jsonb + WriteClassName            {1L:"v"}      (control, non-null key)
```

There is a second, independent site at line 238 (`entryKey.getClass()` on the `value == null && writeNulls` path), which `WriteClassName` does not rescue. The text path is fine — `writeMapKey()` at line 427 checks `key == null` first.

That is a writer-side bug in a different file, unrelated to #7799, and fixing it would change actual JSONB output, so I have kept it out of this PR. Happy to open a separate issue for it — let me know if you'd prefer it folded in here instead.

已确认并在 e46d82e 修复，改为 `"@type".equals(name)`。

在 `9dc64b4` 上按给出的配方（带 null key 的 `HashMap<Long, String>`，`WriteClassName` + `NotWriteHashMapArrayListClassName`）完整复现，因果链也成立：非 String `keyType` 时原先是未加守卫的 `nextIfMatch(':')` 先抛异常，正是本 PR 新增的 `!jsonb` 守卫让 3661 行变得可达。

新增 fix witness：`JSONReaderTest.readMap_jsonb_nullKey_autoType`，断言 `map.get(null)`。已验证去掉修复变红（`NullPointerException ... "name" is null`），加回变绿。

补充说明测试为什么必须用这两个 writer feature、而不能直接 `JSONB.toBytes(src)`——写端有一个相关的既有缺陷。`ObjectWriterImplMap.writeJSONB` 其实写了 `entryKey == null` 分支，但它排在**第三位**，前面挡着一个兜底条件：

```java
if (entryKey instanceof String || (contextFeatures & Feature.WriteClassName.mask) == 0) {
    ...
    key = entryKey.toString();   // 268 行，NPE
} else if (entryKey == null) {
    jsonWriter.writeNull();      // 281 行，逻辑正确，但默认配置下不可达
}
```

`WriteClassName` 关闭（默认）时第一个条件的后半段恒为 true，所有 key 都被第一个分支吞掉，null 分支成为死代码。实测：

```
[OK   ] text  JSON.toJSONString({null:"v"})            = {"null":"v"}
[OK   ] text  + WriteNulls  {null:null}                = {"null":null}
[THROW] jsonb JSONB.toBytes({null:"v"})  默认 feature     -> NPE @ ObjectWriterImplMap.writeJSONB:268
[OK   ] jsonb + WriteClassName            {null:"v"}
[THROW] jsonb + WriteClassName+WriteNulls {null:null}   -> NPE @ ObjectWriterImplMap.writeJSONB:238
[OK   ] jsonb + WriteClassName            {1L:"v"}      （对照组，非 null key）
```

另有一处独立的 238 行（`value == null && writeNulls` 路径上的 `entryKey.getClass()`），开 `WriteClassName` 也躲不掉。文本路径没有这个问题——`writeMapKey()`（427 行）第一句就判了 `key == null`。

这是另一个文件里的写端 bug，与 #7799 无关，且修复会改变实际的 JSONB 输出字节，因此未纳入本 PR。可以另开 issue 跟踪；如果您希望一并在这个 PR 里修，也请告知。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
1919chichi: Agreed on the conclusion — no code change, keeping the guard. One correction to the premise, which I think actually strengthens the case for keeping it.

`JSONReaderJSONB` *does* assign `ch`: `JSONReaderJSONB.java:1380`, inside `startArray()`:

```java
if (type >= BC_ARRAY_FIX_MIN && type <= BC_ARRAY_FIX_MAX) {
    ch = (char) -type;
    return type - BC_ARRAY_FIX_MIN;
}
```

The branch is nevertheless unreachable, but for a narrower reason than "never assigns `ch`": `BC_ARRAY_FIX_0 = -108` and `ARRAY_FIX_LEN = 15`, so `-type` only ever lands in 93..108, and `'/'` is 47. It never collides.

So `ch` on a JSONB reader is not permanently 0 — it holds a stale array-length byte after any fixed-length array is read. That is precisely the condition under which an unguarded `ch == '/'` would be fragile if the encoding ever grew a byte in the ASCII range, so I would rather keep the guard than remove an unpinned-but-correct check. Agreed there is no sensible way to pin it with a test today without faking a `ch` value that the format cannot produce.

结论同意——不改代码，保留守卫。但论据有一处需要更正，而且这处更正反而更支持保留它。

`JSONReaderJSONB` **是会**给 `ch` 赋值的：`JSONReaderJSONB.java:1380`，在 `startArray()` 里：

```java
if (type >= BC_ARRAY_FIX_MIN && type <= BC_ARRAY_FIX_MAX) {
    ch = (char) -type;
    return type - BC_ARRAY_FIX_MIN;
}
```

该分支确实不可达，但原因比"从不给 `ch` 赋值"更窄：`BC_ARRAY_FIX_0 = -108`、`ARRAY_FIX_LEN = 15`，因此 `-type` 只可能落在 93..108，而 `'/'` 是 47，永远撞不上。

也就是说，JSONB reader 上的 `ch` 并非恒为 0——读过任意定长数组之后，它会残留一个数组长度字节。恰恰是这种情况会让未加守卫的 `ch == '/'` 变得脆弱（一旦编码将来引入落在 ASCII 区间的字节）。因此相比删掉一个"没有测试钉住但正确"的检查，我倾向于保留它。同意目前没有合理手段用测试钉住它——除非伪造一个该格式根本产生不出来的 `ch` 值。

</details>

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
1919chichi: Follow-up, superseding the last paragraph of my previous comment: the writer-side defect is now fixed in this PR (d8e6a7a) rather than deferred to a separate issue. Serialization and deserialization are two halves of one feature, and splitting them would leave a release where a null-keyed map reads back correctly but still throws on write, so folding it in is the right call.

What changed in `ObjectWriterImplMap.writeJSONB()`:

- The `entryKey == null` check is hoisted ahead of the catch-all condition. The `else if (entryKey == null)` branch that used to sit third is deleted because the hoist makes it unreachable — not because null stopped being handled. With `WriteClassName` on, the exact same `jsonWriter.writeNull()` still runs, just from the first branch instead of the third.
- The same check is added on the `value == null && WriteNulls` path, which `WriteClassName` never rescued.

Verified byte-for-byte on `HashMap{null: "v"}`, so the previously working configurations are provably untouched:

| features | before | after |
| --- | --- | --- |
| `WriteClassName` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5` (identical) |
| `WriteClassName` + `ReferenceDetection` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5` (identical) |
| `WriteClassName`, `@type` emitted | `92 4a 4d 00 a6 af 4a 76 a5` | `92 4a 4d 00 a6 af 4a 76 a5` (identical) |
| default (no `WriteClassName`) | NPE | `a6 af 4a 76 a5` |
| `WriteClassName` + `WriteNulls`, null value | NPE | `a6 af af a5` |

(`a6` = `BC_OBJECT`, `af` = `BC_NULL`, `4a 76` = 1-char ASCII string `"v"`, `a5` = `BC_OBJECT_END`.) Only the two rows that used to throw changed, so there is no compatibility risk.

Tests: `MapNullKeyTest` covers six JSONB round-trips of a null-keyed map — default features, `WriteClassName`, `WriteNulls` with a null value, `WriteNulls` + `WriteClassName`, a typed `Map<Long, String>` bean field, and a non-null-key control. Reverting the writer fix turns four of them RED (two per NPE site); the `WriteClassName` case and the non-null-key case are controls that pass either way, pinning exactly the behaviour that must not change. `javap` puts `writeJSONB()` at ~1160 bytes, already well past the 325-byte C2 inlining threshold, so the added branches do not affect inlining.

Whole-reactor `mvn test`: 9686 tests, 0 failures, 0 errors. `mvn validate` clean. The PR description now carries all of this as "Additional fix 4".

补充说明，覆盖我上一条回复的最后一段：写端缺陷已在本 PR 中一并修复（d8e6a7a），不再另开 issue 推迟。序列化和反序列化是同一个功能的两半，拆开修会让某个版本处于「带 null key 的 map 能正确读回、但写出去仍然抛异常」的不一致状态，因此并入本 PR 更合适。

`ObjectWriterImplMap.writeJSONB()` 的改动：

- 把 `entryKey == null` 判断提到兜底条件之前。原先排在第三位的 `else if (entryKey == null)` 之所以删掉，是因为提前之后它已不可达，而不是不再处理 null——`WriteClassName` 开启时执行的仍是同一句 `jsonWriter.writeNull()`，只是从第三分支变成第一分支。
- 在 `value == null && WriteNulls` 路径补上同样的判断，这一处开 `WriteClassName` 也躲不掉。

已对 `HashMap{null: "v"}` 做字节级比对，证明原本正常的配置分毫未动：

| feature 组合 | 修复前 | 修复后 |
| --- | --- | --- |
| `WriteClassName` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5`（一致） |
| `WriteClassName` + `ReferenceDetection` | `a6 af 4a 76 a5` | `a6 af 4a 76 a5`（一致） |
| `WriteClassName`，写出 `@type` | `92 4a 4d 00 a6 af 4a 76 a5` | `92 4a 4d 00 a6 af 4a 76 a5`（一致） |
| 默认（未开 `WriteClassName`） | NPE | `a6 af 4a 76 a5` |
| `WriteClassName` + `WriteNulls`，值为 null | NPE | `a6 af af a5` |

（`a6` = `BC_OBJECT`，`af` = `BC_NULL`，`4a 76` = 长度 1 的 ASCII 串 `"v"`，`a5` = `BC_OBJECT_END`。）只有原本抛异常的两行发生变化，无兼容性风险。

测试：`MapNullKeyTest` 覆盖 null key 的六种 JSONB 往返组合——默认配置、`WriteClassName`、`WriteNulls` + null value、`WriteNulls` + `WriteClassName`、类型化的 `Map<Long, String>` bean 字段，以及非 null key 对照组。回退写端修复后其中四个变红（两个 NPE 点各两个）；`WriteClassName` 那个用例和非 null key 用例两种情况下都绿，正是用来钉住「不允许变化」的那部分行为。`javap` 实测 `writeJSONB()` 约 1160 字节，本就远超 C2 的 325 字节内联阈值，新增分支不影响内联。

整个 reactor `mvn test`：9686 个测试，0 失败 0 错误；`mvn validate` 通过。以上内容已写入 PR 描述的「Additional fix 4」一节。

</details>

PR #7802, opened 2026-08-22
fix: make int overflow handling consistent across equivalent JSON numbers

### LINKED ISSUE #7801: [BUG] 等价 JSON 数值转 int 时产生拒绝／回绕／饱和三种结果，与 JLS 窄化语义不一致
### 问题描述

同一个数学值 `10¹⁰` 绑定到 Java `int` 字段时，fastjson2 的结果取决于 JSON 数值的**词法书写形式**，而不是数值本身：

- 整数写法 `10000000000`：抛出 `JSONException`；
- 普通小数写法 `10000000000.0`：得到 `1410065408`，发生模 `2^32` 回绕；
- 科学计数法 `1e10`：得到 `2147483647`，发生饱和。

这三种 JSON token 表示完全相同的数学值，却触发了三条不同的内部转换路径。

其中科学计数法的结果与 JLS §5.1.3 规定的浮点数到整数窄化转换一致：超出 `int` 正向范围时应饱和到 `Integer.MAX_VALUE`，即：

```java
(int) 1e10 == Integer.MAX_VALUE;
```

普通小数路径表现得像 `BigDecimal.intValue()`，而整数路径执行了显式溢出拒绝。当前行为把内部中间类型和转换实现泄漏到了公开 JSON 绑定结果中，导致等价数值无法获得等价结果。

### 环境信息
 - OS信息：Ubuntu 20.04.6 LTS
 - JDK信息：Eclipse Temurin 1.8.0_402
 - 版本信息：Fastjson2 2.0.64；2.0.65-SNAPSHOT

### 重现步骤
1. 使用 `JSON.parseObject(String, Class)` 将 JSON 对象绑定到包含 `int` 字段的 Bean。
2. 分别输入数学值相同、书写形式不同的 `10000000000`、`10000000000.0` 和 `1e10`。
3. 可以观察到拒绝、回绕和饱和三种不同结果。

```java

public class Repro {
    public static class Bean {
        private int count;

        public int getCount() {
            return count;
        }

        public void setCount(int count) {
            this.count = count;
        }
    }

        for (String value : new String[]{"10000000000", "10000000000.0", "1e10"}) {
            try {
                Bean bean = JSON.parseObject("{\"count\":" + value + "}", Bean.class);
                System.out.println(value + " -> " + bean.getCount());
            } catch (Exception error) {
                System.out.println(value + " -> " + error.getClass().getSimpleName());
            }
        }
    }
}
```

完整写法矩阵如下，所有输入值均为 `10¹⁰`：

| 写法 | 实际结果 | 转换表现 |
|---|---:|---|
| `10000000000` | `JSONException` | 拒绝 |
| `10000000000.0` | `1410065408` | 回绕 |
| `10000000000.00` | `1410065408` | 回绕 |
| `1e10` | `2147483647` | JLS 饱和 |
| `1E10` | `2147483647` | JLS 饱和 |
| `1.0e10` | `2147483647` | JLS 饱和 |
| `100e8` | `2147483647` | JLS 饱和 |
| `0.1e11` | `2147483647` | JLS 饱和 |

### 期待的正确结果

在相同的 `JSONReader` 配置下，表示同一数学值的不同合法 JSON 写法应产生相同结果，不应因为解析阶段选择了 `Long`、`BigDecimal` 或 `Double` 中间类型而分别出现拒绝、回绕和饱和。

如果按 Java 原始类型窄化语义处理，JLS §5.1.3 给出的参考结果是饱和到 `Integer.MAX_VALUE`。如果 fastjson2 选择维持其已记录的默认溢出策略，则应至少保证：

- 默认配置下，所有超出 `int` 范围的等价写法统一抛出 `JSONException`；
- 启用 `JSONReader.Feature.NonErrorOnNumberOverflow` 后，所有等价写法统一采用相同的非报错窄化结果。

无论选择哪种策略，都不应由数值的词法书写形式决定结果。

### 相关日志输出
```text
10000000000    -> JSONException
10000000000.0  -> 1410065408
1e10           -> 2147483647
```

参考 Java 转换结果：

```java
(int) 1e10                                // 2147483647，JLS §5.1.3 饱和
new BigDecimal("1e10").intValue()         // 1410065408，模 2^32 回绕
```

#### 附加信息

- `long` 字段不受影响，上述八种写法均得到 `10000000000L`。
- 问题在 `String`、`char[]` 和 UTF-8 `byte[]` 输入路径均可观察到。
- 这可能造成校验层和执行层在数值文本被重新格式化后，对同一个业务值产生不同结果。

annoymous3:
Fixes #7801 

  Equivalent JSON representations of the same mathematical value could previously produce different `int` results depending on their lexical form:

  - `10000000000` threw `JSONException`.
  - `10000000000.0` wrapped to `1410065408`.
  - `1e10` saturated to `2147483647`.

  This PR makes `int` overflow handling independent of the JSON number's lexical representation.

  ### Summary of your change

  - Made decimal-to-`int` conversion follow fastjson2's documented overflow policy.
  - By default, all equivalent out-of-range representations throw `JSONException`.
  - With `JSONReader.Feature.NonErrorOnNumberOverflow`, all representations use the same non-error narrowing result.
  - Preserved truncation toward zero for fractional values within the `int` range.
  - Added regression tests for String, `char[]`, and UTF-8 `byte[]` inputs.
  - Added coverage for eight equivalent representations of `10^10`.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -1955,7 +1955,9 @@ public abstract class JSONReader
                 Number number = getNumber();
                 if (number instanceof Long) {
                     long longValue = number.longValue();
-                    if (longValue < Integer.MIN_VALUE || longValue > Integer.MAX_VALUE) {
+                    if ((longValue < Integer.MIN_VALUE || longValue > Integer.MAX_VALUE)
+                            && (context.features & Feature.NonErrorOnNumberOverflow.mask) == 0
+                    ) {
                         throw new JSONException(info("integer overflow " + longValue));
                     }
                     return (int) longValue;
@@ -1972,8 +1974,17 @@ public abstract class JSONReader
                     }
                 }
                 return number.intValue();
-            case JSON_TYPE_DEC:
-                return getNumber().intValue();
+            case JSON_TYPE_DEC: {
+                BigDecimal decimal = getBigDecimal();
+                if ((context.features & Feature.NonErrorOnNumberOverflow.mask) != 0) {
+                    return decimal.intValue();
+                }
+                try {
+                    return decimal.toBigInteger().intValueExact();
+                } catch (ArithmeticException e) {
+                    throw numberError();
+                }
+            }
             case JSON_TYPE_BOOL:
                 return boolValue ? 1 : 0;
             case JSON_TYPE_NULL:
--- a/core/src/test/java/com/alibaba/fastjson2/JSONReaderInt32Test.java
+++ b/core/src/test/java/com/alibaba/fastjson2/JSONReaderInt32Test.java
@@ -1,0 +1,49 @@
+package com.alibaba.fastjson2;
+
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+import java.nio.charset.StandardCharsets;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class JSONReaderInt32Test {
+    @Test
+    public void decimalOverflow() {
+        for (String value : new String[]{
+                "10000000000",
+                "10000000000.0",
+                "10000000000.00",
+                "1e10",
+                "1E10",
+                "1.0e10",
+                "100e8",
+                "0.1e11"
+        }) {
+            String json = "{\"count\":" + value + "}";
+            assertThrows(JSONException.class, () -> JSON.parseObject(json, Bean.class), value);
+            assertThrows(JSONException.class, () -> JSON.parseObject(json.toCharArray(), Bean.class), value);
+            assertThrows(
+                    JSONException.class,
+                    () -> JSON.parseObject(json.getBytes(StandardCharsets.UTF_8), Bean.class),
+                    value
+            );
+
+            Bean bean = JSON.parseObject(json, Bean.class, JSONReader.Feature.NonErrorOnNumberOverflow);
+            assertEquals(new BigDecimal(value).intValue(), bean.count, value);
+        }
+    }
+
+    @Test
+    public void decimalFractionWithinRange() {
+        assertEquals(1, JSON.parseObject("{\"count\":1.9}", Bean.class).count);
+        assertEquals(-1, JSON.parseObject("{\"count\":-1.9}", Bean.class).count);
+        assertEquals(Integer.MAX_VALUE, JSON.parseObject("{\"count\":2147483647.9}", Bean.class).count);
+        assertEquals(Integer.MIN_VALUE, JSON.parseObject("{\"count\":-2147483648.9}", Bean.class).count);
+    }
+
+    public static class Bean {
+        public int count;
+    }
+}


core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -1972,8 +1974,17 @@ public final int getInt32Value() {
...
                 return number.intValue();
-            case JSON_TYPE_DEC:
-                return getNumber().intValue();
+            case JSON_TYPE_DEC: {
+                BigDecimal decimal = getBigDecimal();
+                if ((context.features & Feature.NonErrorOnNumberOverflow.mask) != 0) {
+                    return decimal.intValue();
+                }
+                try {
+                    return decimal.toBigInteger().intValueExact();
```
wenshao, line 1983: **[Critical]** The overflow decision here is made on the value returned by `getBigDecimal()`, but for exponent-form literals that is not the exact decimal: `getBigDecimal()`'s `JSON_TYPE_DEC` branch (around lines 4313-4317 of this file) routes every `exponent != 0` literal through `Double.parseDouble(decimalStr + "E" + exponent)` followed by `TypeUtils.toBigDecimal(double)` — a lossy 53-bit-mantissa round trip. Exponent forms and plain-decimal forms of the same value therefore get different int results — the exact lexical-form divergence this PR exists to eliminate — and this diff introduces regressions near the boundaries (pre-diff these inputs returned the correct value via `Double.intValue()` saturation).

Verified end-to-end against the PR build and the merge base:
- `{"count":2.14748364799999999e9}` — true value 2147483647.99999999, truncates to the in-range `Integer.MAX_VALUE` — but the double rounds up to 2147483648.0, so `intValueExact()` throws a spurious JSONException, while the identical plain form `2147483647.99999999` returns 2147483647 (the semantics this PR's own `decimalFractionWithinRange` pins). The mirror case `-2.14748364899999999e9` throws spuriously as well.
- With `NonErrorOnNumberOverflow` enabled, the same rounding makes the exponent form silently return `-2147483648` (sign flip) via `decimal.intValue()`, while the dotted form returns 2147483647.

```
BASE: 2.14748364799999999e9 => 2147483647
PR:   2.14748364799999999e9 => JSONException: illegal number, offset 31, char }
PR:   2147483647.99999999   => 2147483647
BASE: exp form + NonErrorOnNumberOverflow => 2147483647
PR:   exp form + NonErrorOnNumberOverflow => -2147483648 (sign flip)
```

The PR's tests cannot catch this: the exponent forms used (`1e10`, `100e8`, `0.1e11`) are far from the int boundaries, where double rounding cannot flip the decision.

Fix: make the exponent-form conversion exact — in `getBigDecimal()`'s DEC branch, build the BigDecimal with scale arithmetic (e.g. `new BigDecimal(unscaledBigInteger, scale - exponent)` or `decimal.scaleByPowerOfTen(exponent)`) instead of the double round trip. This also fixes the `1e309`+ `NumberFormatException` leak commented below. Once fixed, `assertEquals(Integer.MAX_VALUE, JSON.parseObject("{\"count\":2.14748364799999999e9}", Bean.class).count)` added to `JSONReaderInt32Test` should go green — it fails today, and it goes red again if the double round trip is reintroduced.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -1972,8 +1974,17 @@ public final int getInt32Value() {
                     }
                 }
                 return number.intValue();
-            case JSON_TYPE_DEC:
-                return getNumber().intValue();
+            case JSON_TYPE_DEC: {
+                BigDecimal decimal = getBigDecimal();
```
wenshao, line 1978: **[Critical]** For exponent-form decimals whose magnitude exceeds double range (roughly `1e309`..`1e2047`, all accepted since `readNumber0`'s `MAX_EXP = 2047`), this call throws a raw `NumberFormatException` instead of the `JSONException` every other overflow path in `getInt32Value()` produces — and it runs before the `NonErrorOnNumberOverflow` check below, so the opt-in non-error feature is bypassed as well. The chain: `Double.parseDouble("1E400")` = Infinity, then `TypeUtils.toBigDecimal(double)` routes it through `NumberUtils.writeDouble`/`writeSpecial`, which writes the literal bytes `null`, and `parseBigDecimal` falls back to `new BigDecimal("null")` — a NumberFormatException ("Character n is neither a decimal digit...") escapes `JSON.parseObject`. Callers catching `JSONException` — the contract this PR's own tests pin — miss it. Pre-diff this input silently saturated to `Integer.MAX_VALUE`, so the wrong-typed leak is introduced by this diff.

```
BASE: 1e400 => 2147483647 (silent Double saturation)
PR:   1e400 => java.lang.NumberFormatException: Character n is neither a decimal digit number...
PR:   -1e400 => java.lang.NumberFormatException
PR:   1e400 + NonErrorOnNumberOverflow => java.lang.NumberFormatException (feature bypassed)
```

Fix: the exact BigDecimal construction proposed in the other comment removes the double/Infinity path entirely; as a minimal local alternative, widen the catch below to `catch (ArithmeticException | NumberFormatException e) { throw numberError(); }`. Once fixed, `assertThrows(JSONException.class, () -> JSON.parseObject("{\"count\":1e400}", Bean.class))` added to `JSONReaderInt32Test` should pass — it fails today with the escaping NumberFormatException.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -1972,8 +1974,17 @@ public final int getInt32Value() {
...
+                if ((context.features & Feature.NonErrorOnNumberOverflow.mask) != 0) {
```
wenshao, line 1979: **[Suggestion]** This feature gate is missing from the sibling `JSON_TYPE_BIG_DEC` case of the same switch (around lines 2018-2027): it calls `getBigDecimal().intValueExact()` unconditionally. `readNumber0` classifies literals with more than 38 significant digits as BIG_DEC (JSONReaderUTF8.java:5476), so with `NonErrorOnNumberOverflow` enabled, `1e40` wraps via this new DEC branch (returns 0) while the same value written as a 41-digit literal throws JSONException — the PR description's "all representations use the same non-error narrowing result" is observably false. The BIG_DEC branch's `intValueExact()` also rejects in-range fractions: `1.5` followed by 40 zeros (mathematically identical to `1.5`) throws even in default config, while `1.5` truncates to 1. Verified against the PR build:

```
PR: 1e40 + NonErrorOnNumberOverflow => 0
PR: 41-digit literal + NonErrorOnNumberOverflow => JSONException: illegal number, offset 51, char }
PR: 1.5 + 40 trailing zeros (default) => JSONException (while 1.5 => 1)
```

Mirror this branch in the BIG_DEC case:

```java
case JSON_TYPE_BIG_DEC: {
    BigDecimal decimal = getBigDecimal();
    if ((context.features & Feature.NonErrorOnNumberOverflow.mask) != 0) {
        return decimal.intValue();
    }
    try {
        return decimal.toBigInteger().intValueExact();
    } catch (ArithmeticException e) {
        throw numberError();
    }
}
```

Once added, a >38-digit literal in `decimalOverflow`'s value list (e.g. 10^40 spelled with 41 digits) must return `new BigDecimal(value).intValue()` under `NonErrorOnNumberOverflow` — it throws today, so that assertion pins the change.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -1972,8 +1974,17 @@ public final int getInt32Value() {
...
+                    return decimal.intValue();
+                }
+                try {
+                    return decimal.toBigInteger().intValueExact();
+                } catch (ArithmeticException e) {
+                    throw numberError();
+                }
```
wenshao, line 1986: **[Suggestion]** Decimal overflow here reports `numberError()` ("illegal number, offset N, char X", quoting the innocent delimiter after the number) while the sibling Long and INT64/FLOAT/DOUBLE branches in the same method report `info("integer overflow " + value)` — for the same mathematical overflow. Verified against the PR build: `{"count":10000000000}` throws "integer overflow 10000000000, offset 21, ..." but `{"count":10000000000.0}` throws "illegal number, offset 23, char }". The JSON is perfectly valid, nothing in the message says overflow, and the quoted char is the closing brace — whoever debugs it is sent chasing a syntax error that doesn't exist, on the exact axis this PR sets out to unify.

```suggestion
                } catch (ArithmeticException e) {
                    throw new JSONException(info("integer overflow " + decimal));
                }
```

Once changed, capture the exception thrown for one decimal form in `decimalOverflow` and assert its message contains "integer overflow" — that assertion goes red if this reverts to `numberError()`, pinning the fix.

wenshao (changes requested):

PR #7804, opened 2026-08-22
fix: make numeric boolean coercion consistent

### LINKED ISSUE #7803: [BUG] 等价 JSON 数值转 boolean 时结果依赖书写形式并静默降级
### 问题描述

将 JSON 数值绑定到 Java `boolean` 字段时，fastjson2 的结果取决于数值的**词法书写形式**：整数写法 `1` 得到 `true`，而数学值相同的小数写法 `1.0` 得到 `false`。

RFC 8259 对 JSON number 使用统一的数值语法，并未定义独立的整数和小数数据类型。`1`、`1.0`、`1e0` 和 `10e-1` 表示相同的数学值，因此不应因为解析阶段进入不同的内部数值分支而得到相反的布尔值。

当前行为还存在静默降级：小数数值不会被拒绝，而是无条件得到 `false`。例如 `1.0` 不抛出任何异常，却把原本由整数写法表示为 `true` 的字段静默改为 `false`。

JLS §5.1.7 没有定义数值类型与 `boolean` 之间的装箱、拆箱或转换关系，因此允许数值绑定到 boolean 是 fastjson2 自行定义的 coercion 规则。本问题并非要求 Java 提供这种转换，而是要求库自身的规则满足一致性：同一数学值的不同合法 JSON 写法应得到相同结果。

此外，启用 `JSONReader.Feature.NonZeroNumberCastToBooleanAsTrue` 后，小数路径仍返回 `false`。例如 `0.5` 是非零数值，却没有按照该 Feature 的名称和现有整数语义得到 `true`。

### 环境信息

 - OS信息：Ubuntu 20.04.6 LTS
 - JDK信息：Eclipse Temurin 1.8.0_402
 - 版本信息：Fastjson2 2.0.64；2.0.65-SNAPSHOT

### 重现步骤

1. 使用 `JSON.parseObject(String, Class)` 将 JSON 对象绑定到包含 `boolean` 字段的 Bean。
2. 分别输入数学值相同、书写形式不同的 `1` 和 `1.0`。
3. 可以观察到 `1` 得到 `true`，`1.0` 却得到 `false`，且后者不抛异常。

```java

public class Repro {
    public static class Bean {
        private boolean enabled;

        public boolean isEnabled() {
            return enabled;
        }

        public void setEnabled(boolean enabled) {
            this.enabled = enabled;
        }
    }

        for (String value : new String[]{"1", "1.0", "-1", "2", "0"}) {
            Bean bean = JSON.parseObject("{\"enabled\":" + value + "}", Bean.class);
            System.out.println(value + " -> " + bean.isEnabled());
        }
    }
}
```

完整取值表：

| `{"enabled":X}` | 实际结果 | 说明 |
|---|---:|---|
| `1` | `true` | 整数路径 |
| `1.0` | **`false`** | 与 `1` 同值但结果相反 |
| `-1` | `true` | 当前默认规则按绝对值 1 判断 |
| `-2` | `false` | |
| `2` | `false` | |
| `0` | `false` | |
| `0.0` | `false` | |
| `0.5` | `false` | 启用 `NonZeroNumberCastToBooleanAsTrue` 后也仍为 `false` |
| `"1"` | `true` | 字符串路径 |
| `"true"` | `true` | 字符串路径 |
| `"yes"` | `JSONException` | 无法解释的字符串会被拒绝 |
| `null` | `false` | primitive boolean 默认值 |

### 期待的正确结果

在相同的 `JSONReader` 配置下，表示同一数学值的合法 JSON number 写法应产生相同的 boolean 结果，不应由整数、小数或指数的词法形式决定。

为了保持兼容性，可以沿用 fastjson2 当前默认的整数规则，并将其他数值形式与之对齐：

- 默认配置：数值绝对值等于 1 时为 `true`，否则为 `false`；因此 `1`、`1.0`、`1e0`、`10e-1` 均为 `true`，`2` 与 `2.0` 均为 `false`；
- 启用 `NonZeroNumberCastToBooleanAsTrue`：所有非零数值均为 `true`，包括 `0.5`、`-0.5` 和 `2.0`；零的各种写法均为 `false`。

另一种可接受策略是拒绝所有非整数数值并抛出 `JSONException`，但不应继续把无法按当前分支处理的小数静默降级为 `false`。

### 相关日志输出

```text
1    -> true
1.0  -> false
-1   -> true
2    -> false
0    -> false
```

启用 `JSONReader.Feature.NonZeroNumberCastToBooleanAsTrue` 后：

```text
0.5  -> false    // 预期为 true
2.0  -> false    // 预期为 true
```

#### 附加信息

- 当前整数路径的默认规则实际为 `|value| == 1`，而小数路径无条件返回 `false`，可解释上述全部结果。
- 问题在 `String`、`char[]` 和 UTF-8 `byte[]` 输入路径均可复现。
- 字符串值和 `null` 的现有行为与本问题无关，可以保持不变。
- 同类的词法形式依赖也曾出现在 JSON 数值转 `int` 的路径中：同一数学值可因写法不同而出现拒绝、回绕或饱和。

annoymous3:

  Fixes #7803 

  Equivalent JSON number representations currently produce different boolean
  values. For example, `1` is converted to `true`, while the mathematically
  equivalent `1.0` is silently converted to `false`.

  Decimal values also bypass
  `JSONReader.Feature.NonZeroNumberCastToBooleanAsTrue`, causing non-zero values
  such as `0.5` and `2.0` to be converted to `false`.

  ### Summary of your change

  - Centralize numeric-to-boolean conversion for UTF-16 and UTF-8 readers.
  - Preserve the existing default rule where values with an absolute value of
    one are converted to `true`.
  - Make integer, decimal, and exponent representations behave consistently.
  - Apply `NonZeroNumberCastToBooleanAsTrue` to decimal values.
  - Preserve the allocation-free integer path.
  - Add regression tests for String, char[], and UTF-8 byte[] inputs.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReader.java
@@ -3868,6 +3868,37 @@ public abstract class JSONReader
         return boolValue;
     }
 
+    protected final boolean getBooleanValue() {
+        return (context.features & Feature.NonZeroNumberCastToBooleanAsTrue.mask) != 0
+                ? isNonZeroNumber()
+                : isOneNumber();
+    }
+
+    private boolean isNonZeroNumber() {
+        return mag0 != 0 || mag1 != 0 || mag2 != 0 || mag3 != 0;
+    }
+
+    private boolean isOneNumber() {
+        if (valueType == JSON_TYPE_INT) {
+            return isOneInt();
+        }
+
+        return isOneDecimal();
+    }
+
+    private boolean isOneInt() {
+        return mag0 == 0
+                && mag1 == 0
+                && mag2 == 0
+                && mag3 == 1;
+    }
+
+    private boolean isOneDecimal() {
+        return getBigDecimal()
+                .abs()
+                .compareTo(BigDecimal.ONE) == 0;
+    }
+
     /**
      * Reads a boolean value from JSON data as a primitive boolean.
      *
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF16.java
@@ -5471,17 +5471,7 @@ final class JSONReaderUTF16
             val = false;
         } else if (ch == '-' || (ch >= '0' && ch <= '9')) {
             readNumber();
-            if (valueType == JSON_TYPE_INT) {
-                if ((context.features & Feature.NonZeroNumberCastToBooleanAsTrue.mask) != 0) {
-                    return mag0 != 0 || mag1 != 0 || mag2 != 0 || mag3 != 0;
-                } else {
-                    return mag0 == 0
-                            && mag1 == 0
-                            && mag2 == 0
-                            && mag3 == 1;
-                }
-            }
-            return false;
+            return getBooleanValue();
         } else if (ch == 'n' && offset + 2 < chars.length
                 && chars[offset] == 'u'
                 && chars[offset + 1] == 'l'
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderUTF8.java
@@ -7569,17 +7569,7 @@ class JSONReaderUTF8
             val = false;
         } else if (ch == '-' || (ch >= '0' && ch <= '9')) {
             readNumber();
-            if (valueType == JSON_TYPE_INT) {
-                if ((context.features & Feature.NonZeroNumberCastToBooleanAsTrue.mask) != 0) {
-                    return mag0 != 0 || mag1 != 0 || mag2 != 0 || mag3 != 0;
-                } else {
-                    return mag0 == 0
-                            && mag1 == 0
-                            && mag2 == 0
-                            && mag3 == 1;
-                }
-            }
-            return false;
+            return getBooleanValue();
         } else if (ch == 'n' && offset + 2 < bytes.length
                 && bytes[offset] == 'u'
                 && bytes[offset + 1] == 'l'
--- a/core/src/test/java/com/alibaba/fastjson2/JSONReaderBooleanTest.java
+++ b/core/src/test/java/com/alibaba/fastjson2/JSONReaderBooleanTest.java
@@ -1,0 +1,97 @@
+package com.alibaba.fastjson2;
+
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class JSONReaderBooleanTest {
+    @Test
+    public void numberNotation() {
+        assertValue(true, "1");
+        assertValue(true, "1.0");
+        assertValue(true, "1.00");
+        assertValue(true, "1e0");
+        assertValue(true, "10e-1");
+        assertValue(true, "-1");
+        assertValue(true, "-1.0");
+        assertValue(false, "2");
+        assertValue(false, "2.0");
+        assertValue(false, "-2");
+        assertValue(false, "-2.0");
+        assertValue(false, "0");
+        assertValue(false, "0.0");
+        assertValue(false, "0.5");
+    }
+
+    @Test
+    public void nonZeroNumberFeature() {
+        for (String value : new String[]{"1", "1.0", "-1", "-1.0", "2", "2.0", "0.5", "-0.5"}) {
+            assertFeatureValue(true, value);
+        }
+
+        for (String value : new String[]{"0", "0.0", "0e10"}) {
+            assertFeatureValue(false, value);
+        }
+    }
+
+    @Test
+    public void stringAndNullBehavior() {
+        assertValue(true, "\"1\"");
+        assertValue(true, "\"true\"");
+        assertValue(false, "null");
+        assertThrows(JSONException.class, () -> parse("\"yes\""));
+    }
+
+    private static void assertValue(boolean expected, String value) {
+        String json = "{\"enabled\":" + value + "}";
+        assertEquals(expected, JSON.parseObject(json, Bean.class).enabled, value);
+        assertEquals(expected, JSON.parseObject(json.toCharArray(), Bean.class).enabled, value);
+        assertEquals(
+                expected,
+                JSON.parseObject(json.getBytes(StandardCharsets.UTF_8), Bean.class).enabled,
+                value
+        );
+    }
+
+    private static boolean parse(String value) {
+        return JSON.parseObject("{\"enabled\":" + value + "}", Bean.class).enabled;
+    }
+
+    private static void assertFeatureValue(boolean expected, String value) {
+        String json = "{\"enabled\":" + value + "}";
+        assertEquals(
+                expected,
+                JSON.parseObject(
+                        json,
+                        Bean.class,
+                        JSONReader.Feature.NonZeroNumberCastToBooleanAsTrue
+                ).enabled,
+                value
+        );
+        assertEquals(
+                expected,
+                JSON.parseObject(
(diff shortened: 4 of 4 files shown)


core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3868,6 +3868,37 @@ public Boolean readBool() {
...
+        return mag0 != 0 || mag1 != 0 || mag2 != 0 || mag3 != 0;
+    }
+
+    private boolean isOneNumber() {
+        if (valueType == JSON_TYPE_INT) {
+            return isOneInt();
+        }
+
+        return isOneDecimal();
+    }
```
wenshao, line 3887: **[Critical]** Type-suffixed number literals (`1B`, `1S`, `1L`, `1F`, `1D`) in a boolean position now throw `JSONException("TODO : N")` instead of coercing to a boolean.

`readNumber0()` unconditionally assigns `JSON_TYPE_INT8`/`INT16`/`INT64`/`FLOAT`/`DOUBLE` for suffixed literals in both readers, but `isOneNumber()` only special-cases `JSON_TYPE_INT`, so the suffixed types fall into `isOneDecimal()` → `getBigDecimal()`, whose switch has no case for them and hits `default: throw new JSONException("TODO : " + valueType)`. Before this PR the removed fall-through returned `false` for these inputs, so previously-parseable lenient input now fails to parse — including fastjson2's own writer output: with `WriteClassName`, `JSON.toJSONString` emits `1L`, and re-parsing that into a boolean field throws. The behavior is also self-inconsistent: with `NonZeroNumberCastToBooleanAsTrue` enabled, the same `1L` returns `true` via the mag path instead of throwing.

Witness (probe on this PR's compiled code; String, char[] and byte[] inputs all identical):

```
1L default=THROWS JSONException: TODO : 11 nonZeroFeature=true
1B default=THROWS JSONException: TODO : 9 nonZeroFeature=true
1S default=THROWS JSONException: TODO : 10 nonZeroFeature=true
1F default=THROWS JSONException: TODO : 12 nonZeroFeature=true
1D default=THROWS JSONException: TODO : 13 nonZeroFeature=true
```

```suggestion
    private boolean isOneNumber() {
        switch (valueType) {
            case JSON_TYPE_INT:
            case JSON_TYPE_INT8:
            case JSON_TYPE_INT16:
            case JSON_TYPE_INT64:
                return isOneInt();
            case JSON_TYPE_DEC:
            case JSON_TYPE_BIG_DEC:
                return isOneDecimal();
            default:
                return false;
        }
    }
```

The integer-family suffixed types share the mag layout, so `isOneInt()` works for them as-is (`1L` → `true`, matching the feature path); `FLOAT`/`DOUBLE` keep the pre-PR no-throw fallback. If you apply this, please add cases like `assertValue(true, "1L")` / `assertValue(false, "1F")` to `JSONReaderBooleanTest.numberNotation()` and confirm they fail with `JSONException` if the old dispatch is restored.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3868,6 +3868,37 @@ public Boolean readBool() {
...
+                && mag1 == 0
+                && mag2 == 0
+                && mag3 == 1;
+    }
+
+    private boolean isOneDecimal() {
+        return getBigDecimal()
+                .abs()
+                .compareTo(BigDecimal.ONE) == 0;
+    }
```
wenshao, line 3900: **[Critical]** `isOneDecimal()` routes the exact ±1 test through `getBigDecimal()`, whose `JSON_TYPE_DEC` branch applies a non-zero exponent by round-tripping through `Double.parseDouble` (`JSONReader.java:4333-4337`). This reintroduces notation-dependent booleans for exponent forms — the defect class the linked issue was filed for — and turns large exponents into a hard parse failure.

Two shapes, both reproduced on the UTF-16 and UTF-8 readers:
- Values within half an ulp of 1.0 snap to exactly 1.0: `{"enabled":10000000000000000001e-19}` (exact value 1.0000000000000000001 ≠ 1) coerces to `true`, while the same value written `{"enabled":1.0000000000000000001}` coerces to `false`. Likewise `99999999999999999999e-20` → `true`, expected `false`. Equivalent notations, opposite booleans.
- Exponents beyond double range (the reader admits up to `MAX_EXP = 2047`) overflow to `Infinity` and escape as a raw `NumberFormatException` (`Infinity` → `writeSpecial` writes `null` → `new BigDecimal("null")`); before this PR, `{"enabled":1e400}` coerced to `false`.

Witness (probe on this PR's compiled code):

```
1.0000000000000000001 => false   10000000000000000001e-19 => true
99999999999999999999e-20 => true (expected false)
1e400 default=THROWS NumberFormatException: Character n is neither a decimal digit number, ...
-1e400, 9.9e400 => same; 1e400 with NonZeroNumberCastToBooleanAsTrue => true (no throw)
```

Root-cause fix: in `getBigDecimal()`'s `JSON_TYPE_DEC` case, replace the three-line `Double.parseDouble` round-trip with `return decimal.scaleByPowerOfTen(exponent);` — exact, no throw, and it repairs the same lossy conversion for existing callers of that branch. If the change should stay scoped to this diff, compare exactly from the parsed state instead of delegating:

```suggestion
    private boolean isOneDecimal() {
        if (valueType == JSON_TYPE_BIG_DEC) {
            return getBigDecimal()
                    .abs()
                    .compareTo(BigDecimal.ONE) == 0;
        }

        int[] mag = mag0 == 0
                ? mag1 == 0
                ? mag2 == 0
                ? new int[]{mag3}
                : new int[]{mag2, mag3}
                : new int[]{mag1, mag2, mag3}
                : new int[]{mag0, mag1, mag2, mag3};
        BigDecimal decimal = new BigDecimal(BIG_INTEGER_CREATOR.apply(negative ? -1 : 1, mag), scale);
        return decimal.scaleByPowerOfTen(exponent)
                .abs()
                .compareTo(BigDecimal.ONE) == 0;
    }
```

If you apply either fix, please add `assertValue(false, "10000000000000000001e-19")` and `assertValue(false, "1e400")` to `JSONReaderBooleanTest.numberNotation()` and confirm the first fails (returns `true`) and the second throws when the double round-trip is restored.

core/src/main/java/com/alibaba/fastjson2/JSONReader.java
```diff
@@ -3868,6 +3868,37 @@ public Boolean readBool() {
...
+        return mag0 == 0
+                && mag1 == 0
+                && mag2 == 0
+                && mag3 == 1;
+    }
+
+    private boolean isOneDecimal() {
+        return getBigDecimal()
+                .abs()
+                .compareTo(BigDecimal.ONE) == 0;
```
wenshao, line 3899: **[Suggestion]** Every non-INT number in a boolean position now pays one extra `BigDecimal` allocation through `getBigDecimal()` here (plus a second one from `abs()` for negatives), where the removed code returned `false` without it. `readNumber()` → `getNumber()` already allocates one discarded `Number` for these inputs (pre-existing), so this adds a second per-value allocation.

Measured with a `ThreadMXBean` allocation probe against this PR's build: `+40 B` per parse for `1.0`/`2.0`/`0.5`, `+80 B` for negatives, `+352 B` for exponent forms (`toPlainString` + string concat + `Double.parseDouble` + `toBigDecimal(double)`). The integer path is unaffected (`1` and `true` both 232 B), consistent with the PR's stated goal of keeping it allocation-free.

The exact ±1 test is computable allocation-free from `mag0..mag3`/`scale`/`exponent` for `JSON_TYPE_DEC` (the mantissa must equal 10^(scale−exponent); powers of ten fit the mag representation up to 38 digits); for `JSON_TYPE_BIG_DEC` the source is `stringValue`, so a cheap character scan suffices. The impact is modest — this path only fires when a number appears where a boolean is expected — but an allocation-free fast path ahead of the `getBigDecimal()` fallback would keep a cold path cold in a performance-first library. If the exact-comparison fix from the Critical above is applied, the two naturally combine.

wenshao (changes requested):

PR #7833, opened 2026-08-26
fix: guard readFieldName lookahead against AIOOBE on trailing backslash

### LINKED ISSUE #7809: [BUG] `ArrayIndexOutOfBoundsException` on a field name ending in a backslash — `JSON.parse("{\"\\")`
**Version:** 2.0.64 (reproduced against the published jar)
**Class:** `com.alibaba.fastjson2.JSONReaderASCII`
**Related:** #7790 — same trigger, different reader and method (see below)

A three-byte input throws `ArrayIndexOutOfBoundsException` instead of `JSONException`:

```java
JSON.parse("{\"\\");     // the three bytes: {  "  \
```

```
java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
    at com.alibaba.fastjson2.JSONReaderASCII.readFieldName(JSONReaderASCII.java:949)
    at com.alibaba.fastjson2.JSONReader.read(JSONReader.java:3243)
    at com.alibaba.fastjson2.JSON.parse(JSON.java:138)
```

It is specific to a backslash ending a **field name**:

| input | result |
|---|---|
| `{"\` | **AIOOBE** |
| `{"a\` | **AIOOBE** |
| `{"a":1,"b\` | **AIOOBE** |
| `{"a":"\` | `JSONException` — escape in a value, different path |
| `[{"\` | `JSONException` — array path uses `readFieldNameHashCode()` |

As with the other cases of this kind, callers following the documented contract do not
catch it:

```java
try { JSON.parse(untrusted); } catch (JSONException e) { /* reject */ }
```

## Cause

`JSONReaderASCII.readFieldName()`, lines 945-952:

```java
for (int i = 0; offset < end; ++i) {        // guard: offset < end
    int c = bytes[offset];
    if (c == '\\') {
        nameEscape = true;
        c = bytes[offset + 1];              // line 949 -- reads offset + 1
        offset += (c == 'u' ? 6 : (c == 'x' ? 4 : 2));
        continue;
    }
```

The loop condition covers `bytes[offset]`, but line 949 looks **one ahead** to decide the
skip distance. With the backslash as the final byte, `offset + 1 == end`.

Note this differs from the other reports of this class: it is not a missing
`offset == end` check before a `bytes[offset++]`, it is a lookahead the existing guard
does not cover.

## Fix

```diff
--- a/core/src/main/java/com/alibaba/fastjson2/JSONReaderASCII.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONReaderASCII.java
@@ -946,6 +946,11 @@
             int c = bytes[offset];
             if (c == '\\') {
                 nameEscape = true;
+                // The loop condition only guarantees offset < end; this reads one past the
+                // cursor, so a backslash as the final byte would run off the array.
+                if (offset + 1 >= end) {
+                    throw new JSONException(info("illegal input"));
+                }
                 c = bytes[offset + 1];
                 offset += (c == 'u' ? 6 : (c == 'x' ? 4 : 2));
                 continue;
```

Throwing is appropriate here — a name ending in a backslash cannot be completed, and
`JSONException` is what the surrounding code already raises for malformed names.

Verified against 2.0.64: `{"\`, `{"a\` and `{"a":1,"b\` now raise `JSONException`, and
escaped field names are unaffected — `{"a\"b":1}`, `{"a\\b":1}`, `{"aAb":1}`,
`{"a\tb":1}`, `{"a":1}`, `{}`, `{"a":"x\"y"}` all parse exactly as before.

## Relationship to #7790

#7790 reports the same trigger — a lone trailing backslash — in
`JSONReaderUTF16.readString()` at line 3287. This is the field-name equivalent in the
ASCII reader:

| | #7790 | this issue |
|---|---|---|
| reader | `JSONReaderUTF16` | `JSONReaderASCII` |
| method | `readString()` | `readFieldName()` |
| position | string value | field name |
| minimal | `"\` (needs a UTF16-backed String) | `{"\` |

Both reproduce on 2.0.64, and neither fix affects the other.

## Not a security issue

The out-of-range access is a read, and the JVM bounds check stops it — no memory is read
or disclosed, and the exception message contains only the input length. The practical
impact is an unchecked exception escaping a caller's `catch (JSONException)`.

---
Found by the CISPA Fandango Team

waterWang:
Fixes #7809

`JSONReaderASCII.readFieldName()` reads `bytes[offset + 1]` after seeing a backslash, but the loop guard only covers `bytes[offset]`. When a field name ends in a backslash, `offset + 1 == end`, so the lookahead runs off the array and throws `ArrayIndexOutOfBoundsException` instead of `JSONException`.

Add an explicit check: if the backslash is the final byte, throw `JSONException`, matching how malformed field names are handled elsewhere.


core/src/main/java/com/alibaba/fastjson2/JSONReaderASCII.java
```diff
@@ -946,6 +946,9 @@ public final String readFieldName() {
             int c = bytes[offset];
             if (c == '\\') {
                 nameEscape = true;
+                if (offset + 1 >= end) {
+                    throw new JSONException(info("illegal input"));
```
wenshao, line 950: **[Suggestion]** The new throw diverges from this file's established wording for the same condition. `JSONReaderASCII` already reports "escape ran off the end" as `error("invalid escape character EOI")` at six sites (lines 147, 213, 284, 546, 1269, 1398), and `JSONReader.error(String)` is exactly `new JSONException(info(message))` — so the added line hand-inlines that factory with a different message, and the commit's stated rationale ("matching how malformed field names are handled elsewhere") does not hold. Measured on this commit: `JSON.parseObject("{\"a\\")` (Map path) now reports `illegal input, offset 2` while `JSON.parseObject("{\"a\\", Foo.class)` (bean path, line 147) reports `invalid escape character EOI, offset 2` for the same one-line document — tests or alert rules that match the established string (16 call sites across core) will not recognize the new branch. Suggest reusing the factory: ` ```suggestion
 throw error("invalid escape character EOI");
``` ` No test asserts `illegal input` today (grep over `core/src/test` returns nothing), so the swap contradicts nothing; and if the UTF16 twin at `JSONReaderUTF16.java:1991` gets the same guard in a follow-up, it is worth choosing the wording once for both readers.

_— qwen3.8-flash via Qwen Code /review (v0.22.2)_

File: core/src/main/java/com/alibaba/fastjson2/JSONReaderASCII.java [hunk 2]
```diff
@@ -946,6 +946,9 @@ public final String readFieldName() {
             int c = bytes[offset];
             if (c == '\\') {
                 nameEscape = true;
+                if (offset + 1 >= end) {
```
wenshao, line 949: **[Suggestion]** Nothing in the test suite pins this guard: reverting all three added lines leaves the entire reactor suite green. Measured during this review — the full 15-module `mvn test` on the PR reports 9606 tests / 0 failures, and the same suite on an identical tree with only these three lines reverted (the file restored to `30dd009d1`) reports the exact same 9606 / 0 failures. The behavioral delta is real, though: with the guard, `JSON.parseObject("{\"a\\")` throws `JSONException: illegal input`; with it reverted, the same call throws `ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4`. `readFieldName()` is a large method that gets hot-path refactors regularly, so a future edit that drops this branch would silently resurrect the original #7809 symptom — an unchecked `ArrayIndexOutOfBoundsException` escaping callers' `catch (JSONException)` — with no CI signal. Suggest adding `core/src/test/java/com/alibaba/fastjson2/issues_7000/Issue7809.java` (the convention used by the sibling fixes Issue7668/7678/7691): `assertThrows(JSONException.class, () -> JSON.parseObject("{\"a\\"))` for the inputs `{"\`, `{"a\` and `{"a":1,"b\` — asserting the specific type matters here, since `ArrayIndexOutOfBoundsException` is also a `RuntimeException` and a loose `assertThrows(Exception.class, ...)` would stay green on the unfixed build — plus one positive control such as `{"a\"b":1}` still parsing to key `a"b`. As the acceptance criterion for that new test, remove these three lines and confirm it goes red with the `ArrayIndexOutOfBoundsException`.

_— qwen3.8-flash via Qwen Code /review (v0.22.2)_

wenshao (commented):

Not explored to full depth (tool budget reached): `"agent 5"`: `未执行  mvn test （共享 worktree 未编译， core/target/classes  不存在，规则也禁止在此强制构建），因此没有经验性确认"加了守卫后整套现有测试仍全绿"。`; `"agent 5"`: `未实测  JSON.parseObject("{\\\"中文\\\\")  走 UTF16 抛 AIOOBE 的复现——witness: not run —结论来自源码阅读（ toCharArray()  长度恰为  end 、循环不变式  offset < end ），非执行结果。`; `"agent 5"`: `未核实 JSONReaderUTF8 的  bytes  是否会被分配得比  end  大（若被 pad，则其 UTF8 表现可能是读到脏字节而非 AIOOBE）。`; `"agent 3b"`: `did not check whether  readFieldNameHashCode()  and its  readFieldNameHashCode0()/…Error()  helpers in the three readers carry the same unguarded lookahead (tex…`; `"agent 3b"`: `the sibling-path AIOOBE was demonstrated against the locally installed 2.0.65-SNAPSHOT jar plus the HEAD source, not a build of HEAD (no build run in the shared…`.

_— qwen3.8-flash via Qwen Code /review (v0.22.2)_

PR #7835, opened 2026-08-27
fix: Fix NPE when serializing Kotlin classes without kotlin-reflect on classpath

obabichevjb:

When `kotlin-reflect` is not on the classpath, serializing a Kotlin class with fastjson2 throws:

```
java.lang.NullPointerException: Cannot read the array length because "paramNames" is null
```

at `ObjectWriterBaseModule.java:322`

The call chain:

1. `ObjectWriterBaseModule` detects a Kotlin class via `@Metadata` annotation
2. Calls `KotlinUtils.getKoltinConstructorParameters()`
3. That method returns `null` when `kotlin-reflect` is absent
4. The caller does `for (int i = 0; i < paramNames.length; i++)` → **NPE**

This affects users who have `kotlin-stdlib` transitively (e.g. via RocketMQ, OkHttp) but not `kotlin-reflect`. The serialization side throws; the deserialization side already handles this correctly.

### Summary of your change

Add `if (paramNames != null)` guard around the parameter name iteration — the same pattern already used on the deserialization side (`com.alibaba.fastjson.util.TypeUtils.java:751`). When `kotlin-reflect` is missing, the code now gracefully skips the Kotlin-specific parameter annotation lookup and falls back to standard Java reflection.

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterBaseModule.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterBaseModule.java
@@ -319,16 +319,18 @@ public class ObjectWriterBaseModule
                 Constructor kotlinConstructor = KotlinUtils.getKotlinConstructor(getConstructor(objectClass));
                 if (kotlinConstructor != null) {
                     String[] paramNames = KotlinUtils.getKoltinConstructorParameters(objectClass);
-                    for (int i = 0; i < paramNames.length; i++) {
-                        if (paramNames[i].equals(field.getName())) {
-                            annotations = kotlinConstructor.getParameterAnnotations()[i];
-                            break;
+                    if (paramNames != null) {
+                        for (int i = 0; i < paramNames.length; i++) {
+                            if (paramNames[i].equals(field.getName())) {
+                                annotations = kotlinConstructor.getParameterAnnotations()[i];
+                                break;
+                            }
                         }
-                    }
-                    if (fieldInfo.ignore) {
-                        for (Annotation annotation : annotations) {
-                            if (annotation.annotationType() == JSONField.class) {
-                                fieldInfo.ignore = !((JSONField) annotation).serialize();
+                        if (fieldInfo.ignore) {
+                            for (Annotation annotation : annotations) {
+                                if (annotation.annotationType() == JSONField.class) {
+                                    fieldInfo.ignore = !((JSONField) annotation).serialize();
+                                }
                             }
                         }
                     }


core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterBaseModule.java
```diff
@@ -319,16 +319,18 @@ public void getFieldInfo(BeanInfo beanInfo, FieldInfo fieldInfo, Class objectCla
                 Constructor kotlinConstructor = KotlinUtils.getKotlinConstructor(getConstructor(objectClass));
                 if (kotlinConstructor != null) {
                     String[] paramNames = KotlinUtils.getKoltinConstructorParameters(objectClass);
-                    for (int i = 0; i < paramNames.length; i++) {
-                        if (paramNames[i].equals(field.getName())) {
-                            annotations = kotlinConstructor.getParameterAnnotations()[i];
-                            break;
+                    if (paramNames != null) {
```
wenshao, line 322: **[Suggestion]** The null guard — the only behavioural change in this PR — ships with no regression test, and no test module can reach the guarded branch: kotlin-reflect is on every test classpath that serializes Kotlin classes (core declares it `provided`, which lands on the test classpath; kotest pulls it transitively into every other module), so `getKoltinConstructorParameters` never returns null under `mvn test`. A future refactor that drops or reorders the guard would therefore stay green in CI while the NullPointerException this PR fixes silently returns for users with kotlin-stdlib but without kotlin-reflect — measured: reverting the PR's single hunk leaves the affected suites fully green (core 7978/7978, kotlin 36/36), so nothing in the repo gates this fix.

Witness:
```
git checkout <merge-base> -- core/.../ObjectWriterBaseModule.java   # hunk reverted
mvn -B -ntp -pl core,kotlin test
BUILD SUCCESS — Tests run: 7978 (core) + 36 (kotlin), Failures: 0, Errors: 0
```

Suggested fix: add a regression test that forces the null path and asserts serialization succeeds — e.g. reflectively set `KotlinUtils.kotlin_error` to true (restoring it in a `finally`), serialize a Kotlin class whose fields carry no field-level annotations, and assert JSON is produced; or a dedicated surefire execution whose test classpath excludes kotlin-reflect.

Note that `core/pom.xml:84-86` declares kotlin-reflect with `<scope>provided</scope>` — present on core's test classpath, so a test inside core cannot reach the null branch via classpath alone; and `kotlin_error` (core/src/main/java/com/alibaba/fastjson2/util/KotlinUtils.java:29) is a JVM-global sticky flag — a test toggling it must restore it or later Kotlin tests in the same Surefire JVM silently take the fallback path.

Fix acceptance: removing `if (paramNames != null)` must make the new test fail with the original NullPointerException at `paramNames.length`.

_— qwen3.8-max via Qwen Code /review (v0.22.3)_

wenshao, line 322: **[Suggestion]** When kotlin-reflect is absent, this fallback silently drops Kotlin constructor-parameter annotations: Kotlin's default use-site target puts property annotations (`@JSONField(serialize = false)`, `@JSONField(name = ...)`) on the constructor parameter and the backing field carries nothing, so those directives silently stop applying — with zero log output. In a deployment with kotlin-stdlib but not kotlin-reflect (the exact environment this PR targets), `data class User(@JSONField(serialize = false) val password: String)` serializes the password anyway: the annotation sits on the constructor parameter, the guard skips the lookup, and the field is included — observed `{"id":42,"password":"secret-pw"}` with no log output, while the same class with kotlin-reflect present correctly yields `{"id":42}`. Pre-PR the same configuration crashed loudly with an NPE, so the degradation is silent and indistinguishable from correct output. The deserialization path already warns once per class for the same missing-kotlin-reflect condition via `KotlinUtils.warnParameterNamesUnresolved`; the writer side emits nothing.

Witness:
```
BASE (hunk reverted, no kotlin-reflect): THREW NullPointerException at ObjectWriterBaseModule.java:322
PR (no kotlin-reflect): SERIALIZED {"id":42,"password":"secret-pw"} PASSWORD_LEAKED=true (no log output)
PR (with kotlin-reflect): SERIALIZED {"id":42} PASSWORD_LEAKED=false
READER side, same env: WARNING ... constructor parameter names ... cannot be resolved ... add kotlin-reflect
FIX (type-match fallback patched in scratch tree): SERIALIZED {"id":42} PASSWORD_LEAKED=false
```

Suggested fix: emit a once-per-class warning when the fallback trips, mirroring `KotlinUtils.warnParameterNamesUnresolved` (it already carries the right message and dedups via `unresolvedParameterNamesWarned`); optionally honour parameter annotations by unique type-match — `getParameterAnnotations()`/`getParameterTypes()` are plain Java reflection and work without kotlin-reflect:

```java
if (paramNames == null && kotlinConstructor != null) {
    // best-effort: match constructor parameters to fields by unique type,
    // then apply kotlinConstructor.getParameterAnnotations()[i]
}
```

Note that the reader-side warning dedupes via `unresolvedParameterNamesWarned` (core/src/main/java/com/alibaba/fastjson2/util/KotlinUtils.java:33); if the writer-side warning shares that set, whichever path warns first for a class suppresses the other path's warning.

Fix acceptance: a writer-side test shaped like `core/src/test/java/com/alibaba/fastjson2/util/KotlinUtilsWarningTest.java` — force the fallback (paramNames null) and assert the once-per-class warning is recorded; removing the logging call must turn it red.

_— qwen3.8-max via Qwen Code /review (v0.22.3)_

wenshao, line 322: **[Suggestion]** This guard also converts a pre-existing loud failure into silent wrong output via the sticky `kotlin_error` latch (core/src/main/java/com/alibaba/fastjson2/util/KotlinUtils.java:228,257-258): one Kotlin class whose `@Metadata` kotlin-reflect cannot parse latches the JVM-global flag permanently (caught `Throwable`, raw `e.printStackTrace()`, flag never reset), so every later Kotlin class takes the null-paramNames path — even with kotlin-reflect present. Pre-PR that state crashed loudly at writer creation; post-PR it silently skips constructor-parameter annotations. With `JSONWriter.Feature.FieldBased` (where this block is the only annotation source) an `@JSONField(serialize = false)` on a constructor property is then ignored: observed `{"name":"bob","password":"secret"}` after the latch, versus `{"name":"bob"}` healthy. The default path stays correct because the getter/method variant (ObjectWriterBaseModule.java:852-868) resolves parameter names independently of the latch.

Witness:
```
POISON_JSON={"token":"poison"} (latch set, raw InvocationTargetException stack on stderr)
USER_JSON_AFTER_E2E_POISON={"name":"bob"} (default path fine)
USER_FIELDBASED_AFTER_E2E_POISON={"name":"bob","password":"secret"} (leak)
de-latch fix -> USER_FIELDBASED_JSON={"name":"bob"} (probe flipped)
BASE, latch + FieldBased: THREW NullPointerException (loud)
```

Suggested fix: handle kotlin-reflect failures per-class instead of latching JVM-global (track failed classes in a set, the way `unresolvedParameterNamesWarned` does), and replace the raw `e.printStackTrace()` with a once-per-class logger warning in the `warnParameterNamesUnresolved` style.

Fix acceptance: a test that first serializes a malformed-`@Metadata` class, then asserts FieldBased serialization of a Kotlin class with `@JSONField(serialize = false)` still excludes the field; removing the per-class handling must turn it red.

_— qwen3.8-max via Qwen Code /review (v0.22.3)_

wenshao (commented):

_— qwen3.8-max via Qwen Code /review (v0.22.3)_

obabichevjb:
@wenshao Thanks for the merge! We appreciate it.

PR #7847, opened 2026-09-03
fix: Jackson JsonFormat Shape.STRING on BigDecimal writes illegal JSON, for issue #7837

### LINKED ISSUE #7837: [BUG] 2.0.64 序列化带 @JsonFormat(shape = STRING) 的 BigDecimal 时生成非法 JSON
### 问题描述

从 `com.alibaba:fastjson:2.0.47` 升级到 `2.0.64` 后，使用 Fastjson 1.x 兼容包序列化带 Jackson `@JsonFormat(shape = JsonFormat.Shape.STRING)` 的 `BigDecimal` 字段时，数值 `0.47` 被输出为未加引号的 `string0`，生成非法 JSON；随后调用 `JSON.parseObject` 报 `illegal input error`。回退到 2.0.47 后恢复正常。

### 环境信息

- OS 信息：Windows
- JDK 信息：JDK 1.8.0_162
- Spring Boot：2.3.12.RELEASE
- 版本信息：Fastjson 2.0.64（`com.alibaba:fastjson` 兼容包）

### 重现步骤

1. 引入依赖：

```xml
<dependency>
    <groupId>com.alibaba</groupId>
    <artifactId>fastjson</artifactId>
    <version>2.0.64</version>
</dependency>
```

2. 运行以下最小复现代码：

```java
import com.fasterxml.jackson.annotation.JsonFormat;

import java.math.BigDecimal;

public class FastjsonJacksonFormatBigDecimalTest {

    public static class Bean {
        @JsonFormat(shape = JsonFormat.Shape.STRING)
        public BigDecimal examScore = new BigDecimal("0.47");
    }

        String json = JSON.toJSONString(new Bean());
        System.out.println(json);
        JSON.parseObject(json, Bean.class);
    }
}
```

### 实际结果

Fastjson 2.0.64 输出：

```text
{"examScore":string0}
```

随后解析失败：

```text
com.alibaba.fastjson.JSONException: illegal input error, offset 14, character s, line 1, column 14, fastjson-version 2.0.64
```

### 期待的正确结果

应生成合法 JSON，并保持 Jackson 注解指定的字符串形式：

```json
{"examScore":"0.47"}
```

### 版本对比与临时规避

- 2.0.47 输出 `{"examScore":"0.47"}`，可以正常反序列化。
- 2.0.64 输出 `{"examScore":string0}`，生成非法 JSON。
- 2.0.64 添加 JVM 参数 `-Dfastjson2.useJacksonAnnotation=false` 后输出 `{"examScore":0.47}` 并可解析，但会全局关闭 Jackson 注解兼容，不能作为通用解决办法。
- 当前临时回退并固定在 2.0.47。

### 初步定位

初步对比源码发现，`BeanUtils.processJacksonJsonFormat(FieldInfo, Annotation)` 对 `JsonFormat.Shape.STRING` 的处理发生了变化：

- 2.0.47 使用 `JSONWriter.Feature.WriteNonStringValueAsString`；
- 2.0.64 设置 `fieldInfo.format = "string"`。

对 `BigDecimal` 字段，`"string"` 似乎被当作数值格式化模式，最终把 `0.47` 格式化为未加引号的 `string0`。请确认这是否为回归问题，并协助修复。

WLLTB:
Fixes #7837

Jackson `@JsonFormat(shape = STRING)` on a `BigDecimal` field serializes to illegal JSON: `{"examScore":string0}`, which then fails to parse (`illegal input error`). Since #3750, `BeanUtils.processJacksonJsonFormat` maps `STRING` to `fieldInfo.format = "string"`, but `FieldWriter` built `new DecimalFormat("string")` for float/double/BigDecimal, and `writeDecimal` prefers the `DecimalFormat` path over `WriteNonStringValueAsString` — `DecimalFormat("string").format(0.47)` yields `string0` written raw.

### Summary of your change

- `FieldWriter`: skip `DecimalFormat` creation when `format` is the `"string"`/`"millis"` keyword (constructor + `BigDecimal`/`BigDecimal[]` writer resolution), so the existing `WriteNonStringValueAsString` path writes the value quoted (`"0.47"`).
- `ObjectWriterProvider.getObjectWriter(Type, format)`: same guard for `Double`/`Float`/`BigDecimal`, return the default `INSTANCE` instead of `DecimalFormat("string"/"millis")`.
- Base `FieldWriter.writeInt32`: respect `WriteNonStringValueAsString` (consistent with `writeInt64`/`writeInt16`), fixing the ASM path for `Integer` fields via `FieldWriterObject`.
- New regression test `core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java` covering `BigDecimal`/`Double`/`Float`/`Integer` with `@JsonFormat(shape = STRING)`.

Verified: `core clean test` 7992 passed in both standard and `-Dfastjson2.creator=reflect` modes, `fastjson1-compatible clean test` 1355 passed, root `mvn validate` clean.

- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed. (No docs impact: pure bugfix, no public API change.)

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
@@ -122,6 +122,8 @@ public abstract class FieldWriter<T>
 
         DecimalFormat decimalFormat = null;
         if (format != null
+                && !"string".equals(format)
+                && !"millis".equals(format)
                 && (fieldClass == float.class
                 || fieldClass == float[].class
                 || fieldClass == Float.class
@@ -644,7 +646,11 @@ public abstract class FieldWriter<T>
 
     public void writeInt32(JSONWriter jsonWriter, int value) {
         writeFieldName(jsonWriter);
-        jsonWriter.writeInt32(value);
+        if ((features & WriteNonStringValueAsString.mask) != 0) {
+            jsonWriter.writeString(Integer.toString(value));
+        } else {
+            jsonWriter.writeInt32(value);
+        }
     }
 
     public void writeInt64(JSONWriter jsonWriter, long value) {
@@ -1048,7 +1054,7 @@ public abstract class FieldWriter<T>
             }
 
             if (BigDecimal.class == valueClass) {
-                if (format == null || format.isEmpty()) {
+                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
                     return ObjectWriterImplBigDecimal.INSTANCE;
                 } else {
                     return new ObjectWriterImplBigDecimal(new DecimalFormat(format), null);
@@ -1056,7 +1062,7 @@ public abstract class FieldWriter<T>
             }
 
             if (BigDecimal[].class == valueClass) {
-                if (format == null || format.isEmpty()) {
+                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
                     return new ObjectWriterArrayFinal(BigDecimal.class, null);
                 } else {
                     return new ObjectWriterArrayFinal(BigDecimal.class, new DecimalFormat(format));
--- a/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
@@ -72,6 +72,13 @@ public abstract class FieldWriter<T>
             "initObjectWriter"
     );
 
+    static boolean isFormatKeyword(String format) {
+        return "string".equals(format)
+                || "millis".equals(format)
+                || "trim".equals(format)
+                || "symbol".equals(format);
+    }
+
     FieldWriter(
             String name,
             int ordinal,
@@ -122,8 +129,7 @@ public abstract class FieldWriter<T>
 
         DecimalFormat decimalFormat = null;
         if (format != null
-                && !"string".equals(format)
-                && !"millis".equals(format)
+                && !isFormatKeyword(format)
                 && (fieldClass == float.class
                 || fieldClass == float[].class
                 || fieldClass == Float.class
@@ -1054,7 +1060,7 @@ public abstract class FieldWriter<T>
             }
 
             if (BigDecimal.class == valueClass) {
-                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
+                if (format == null || format.isEmpty() || isFormatKeyword(format)) {
                     return ObjectWriterImplBigDecimal.INSTANCE;
                 } else {
                     return new ObjectWriterImplBigDecimal(new DecimalFormat(format), null);
@@ -1062,7 +1068,7 @@ public abstract class FieldWriter<T>
             }
 
             if (BigDecimal[].class == valueClass) {
-                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
+                if (format == null || format.isEmpty() || isFormatKeyword(format)) {
                     return new ObjectWriterArrayFinal(BigDecimal.class, null);
                 } else {
                     return new ObjectWriterArrayFinal(BigDecimal.class, new DecimalFormat(format));
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
@@ -460,21 +460,21 @@ public class ObjectWriterProvider
      */
     public ObjectWriter getObjectWriter(Type objectType, String format, Locale locale) {
         if (objectType == Double.class) {
-            if ("string".equals(format) || "millis".equals(format)) {
+            if (format == null || format.isEmpty() || FieldWriter.isFormatKeyword(format)) {
                 return ObjectWriterImplDouble.INSTANCE;
             }
             return new ObjectWriterImplDouble(new DecimalFormat(format));
         }
 
         if (objectType == Float.class) {
-            if ("string".equals(format) || "millis".equals(format)) {
+            if (format == null || format.isEmpty() || FieldWriter.isFormatKeyword(format)) {
                 return ObjectWriterImplFloat.INSTANCE;
             }
             return new ObjectWriterImplFloat(new DecimalFormat(format));
         }
 
         if (objectType == BigDecimal.class) {
-            if ("string".equals(format) || "millis".equals(format)) {
+            if (format == null || format.isEmpty() || FieldWriter.isFormatKeyword(format)) {
                 return ObjectWriterImplBigDecimal.INSTANCE;
             }
             return new ObjectWriterImplBigDecimal(new DecimalFormat(format), null);
--- a/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
+++ b/core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
@@ -460,14 +460,23 @@ public class ObjectWriterProvider
      */
     public ObjectWriter getObjectWriter(Type objectType, String format, Locale locale) {
         if (objectType == Double.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplDouble.INSTANCE;
+            }
             return new ObjectWriterImplDouble(new DecimalFormat(format));
         }
 
         if (objectType == Float.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplFloat.INSTANCE;
+            }
             return new ObjectWriterImplFloat(new DecimalFormat(format));
         }
 
         if (objectType == BigDecimal.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplBigDecimal.INSTANCE;
+            }
             return new ObjectWriterImplBigDecimal(new DecimalFormat(format), null);
         }
 
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java
@@ -1,11 +1,14 @@
 package com.alibaba.fastjson2.issues_7800;
 
 import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.annotation.JSONField;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
@@ -31,6 +34,76 @@ public class Issue7837 {
     public static class BeanInteger {
         @JsonFormat(shape = JsonFormat.Shape.STRING)
         public Integer value = 47;
+    }
+
+    public static class BeanObject {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public Object examScore = new BigDecimal("0.47");
+    }
+
+    public static class BeanObjectMillis {
+        @JSONField(format = "millis")
+        public Object examScore = new BigDecimal("0.47");
+    }
+
+    public static class BeanMillisDouble {
+        @JSONField(format = "millis")
+        public Double value = 0.47;
+    }
+
+    public static class BeanMillisBigDecimal {
+        @JSONField(format = "millis")
+        public BigDecimal value = new BigDecimal("0.47");
+    }
+
+    public static class BeanMillisBigDecimalArray {
+        @JSONField(format = "millis")
+        public BigDecimal[] scores = {new BigDecimal("0.47")};
+    }
+
+    public static class BeanObjectBigDecimalArray {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public Object scores = new BigDecimal[]{new BigDecimal("0.47")};
+    }
+
+    public static class BeanObjectMillisBigDecimalArray {
+        @JSONField(format = "millis")
+        public Object scores = new BigDecimal[]{new BigDecimal("0.47")};
+    }
+
+    public static class BeanStringDoubleList {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public List<Double> values = Arrays.asList(0.47, 2.0);
+    }
+
+    public static class BeanStringBigDecimalList {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public List<BigDecimal> values = Arrays.asList(new BigDecimal("0.47"));
+    }
+
+    public static class BeanStringFloatList {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public List<Float> values = Arrays.asList(0.47f);
+    }
+
+    public static class BeanMillisDoubleList {
+        @JSONField(format = "millis")
+        public List<Double> values = Arrays.asList(0.47);
+    }
+
+    public static class BeanMillisBigDecimalList {
+        @JSONField(format = "millis")
+        public List<BigDecimal> values = Arrays.asList(new BigDecimal("0.47"));
+    }
+
+    public static class BeanMillisFloatList {
+        @JSONField(format = "millis")
+        public List<Float> values = Arrays.asList(0.47f);
+    }
+
+    public static class BeanTrimDouble {
+        @JSONField(format = "trim")
+        public Double value = 0.47;
     }
 
     @Test
@@ -61,5 +134,103 @@ public class Issue7837 {
         BeanInteger bean = new BeanInteger();
         String json = JSON.toJSONString(bean);
         assertEquals("{\"value\":\"47\"}", json);
+    }
+
+    @Test
+    public void testObjectBigDecimal() {
+        BeanObject bean = new BeanObject();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"examScore\":\"0.47\"}", json);
+    }
+
+    @Test
+    public void testObjectBigDecimalMillis() {
+        BeanObjectMillis bean = new BeanObjectMillis();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"examScore\":0.47}", json);
+    }
+
+    @Test
+    public void testMillisDouble() {
+        BeanMillisDouble bean = new BeanMillisDouble();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"value\":0.47}", json);
+    }
+
+    @Test
+    public void testMillisBigDecimal() {
+        BeanMillisBigDecimal bean = new BeanMillisBigDecimal();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"value\":0.47}", json);
+    }
+
+    @Test
+    public void testMillisBigDecimalArray() {
+        BeanMillisBigDecimalArray bean = new BeanMillisBigDecimalArray();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"scores\":[0.47]}", json);
+    }
+
+    @Test
+    public void testObjectBigDecimalArray() {
+        BeanObjectBigDecimalArray bean = new BeanObjectBigDecimalArray();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"scores\":[\"0.47\"]}", json);
+    }
+
+    @Test
+    public void testObjectMillisBigDecimalArray() {
+        BeanObjectMillisBigDecimalArray bean = new BeanObjectMillisBigDecimalArray();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"scores\":[0.47]}", json);
+    }
+
+    @Test
+    public void testStringDoubleList() {
+        BeanStringDoubleList bean = new BeanStringDoubleList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[\"0.47\",\"2.0\"]}", json);
+    }
+
+    @Test
+    public void testStringBigDecimalList() {
+        BeanStringBigDecimalList bean = new BeanStringBigDecimalList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[\"0.47\"]}", json);
+    }
+
+    @Test
+    public void testStringFloatList() {
+        BeanStringFloatList bean = new BeanStringFloatList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[\"0.47\"]}", json);
+    }
+
+    @Test
+    public void testMillisDoubleList() {
+        BeanMillisDoubleList bean = new BeanMillisDoubleList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[0.47]}", json);
+    }
+
+    @Test
+    public void testMillisBigDecimalList() {
+        BeanMillisBigDecimalList bean = new BeanMillisBigDecimalList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[0.47]}", json);
+    }
+
+    @Test
+    public void testMillisFloatList() {
+        BeanMillisFloatList bean = new BeanMillisFloatList();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"values\":[0.47]}", json);
+    }
+
+    @Test
+    public void testTrimDouble() {
+        BeanTrimDouble bean = new BeanTrimDouble();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"value\":0.47}", json);
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java
@@ -1,0 +1,65 @@
+package com.alibaba.fastjson2.issues_7800;
+
+import com.alibaba.fastjson2.JSON;
+import com.fasterxml.jackson.annotation.JsonFormat;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Tag("regression")
+@Tag("annotation")
+@Tag("compat-jackson")
+public class Issue7837 {
+    public static class Bean {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public BigDecimal examScore = new BigDecimal("0.47");
+    }
+
+    public static class BeanDouble {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public Double value = 0.47;
+    }
+
+    public static class BeanFloat {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public Float value = 0.47f;
+    }
+
+    public static class BeanInteger {
+        @JsonFormat(shape = JsonFormat.Shape.STRING)
+        public Integer value = 47;
+    }
+
+    @Test
+    public void testBigDecimal() {
+        Bean bean = new Bean();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"examScore\":\"0.47\"}", json);
+        Bean parsed = JSON.parseObject(json, Bean.class);
+        assertEquals(0, new BigDecimal("0.47").compareTo(parsed.examScore));
+    }
+
+    @Test
+    public void testDouble() {
+        BeanDouble bean = new BeanDouble();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"value\":\"0.47\"}", json);
+    }
+
+    @Test
+    public void testFloat() {
+        BeanFloat bean = new BeanFloat();
+        String json = JSON.toJSONString(bean);
+        assertEquals("{\"value\":\"0.47\"}", json);
+    }
+
+    @Test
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -122,6 +122,8 @@ public abstract class FieldWriter<T>
 
         DecimalFormat decimalFormat = null;
         if (format != null
+                && !"string".equals(format)
+                && !"millis".equals(format)
```
wenshao: **[Suggestion]** The `"millis"` half of this guard has no test anywhere in the suite: every existing `format = "millis"` test is on a date/time field, and `Issue7837` covers only `shape = STRING`, so reverting or narrowing this check leaves all 7992 core tests green while `@JSONField(format = "millis")` on a `Double`/`BigDecimal` field regresses to `{"value":millis0}` — the same illegal-JSON class issue #7837 was filed to close. `BeanUtils.processJacksonJsonFormat` maps Jackson `Shape.NUMBER` to exactly this keyword (BeanUtils.java:2945-2948), and unlike `"string"` it is not promoted to a feature bit in this constructor, so this guard is the only protection for that shape.

Witness:
```
delete the four millis guard clauses ->
  mvn -pl core test: exit 0, 0 failures (7992 tests)
  @JSONField(format="millis") Double     -> {"value":millis0}
  @JSONField(format="millis") BigDecimal -> {"value":millis0}
```

Extend `Issue7837` with `@JSONField(format = "millis")` cases on `Double`/`BigDecimal` (and a `List<Double>` variant) asserting plain numbers like `{"value":0.47}`. The millis assertions must use numeric fields, not `Date` fields — date writers consume `"millis"` for epoch-millis output (DateUtils.java:146), so a date-field test would pin behavior this change does not touch. To prove the new tests gate this guard, remove the `!"millis".equals(format)` check and confirm they go red — the witness above is exactly that mutation.

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -1048,15 +1054,15 @@ static ObjectWriter getObjectWriter(
             }
 
             if (BigDecimal.class == valueClass) {
-                if (format == null || format.isEmpty()) {
+                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
                     return ObjectWriterImplBigDecimal.INSTANCE;
```
wenshao: **[Suggestion]** This branch fixes a reachable path that no test in the PR or the existing suite exercises: it is reached through `FieldWriterObject.initWriter` and `ObjectWriterImplOptional` (an `Object`-typed field holding a `BigDecimal`, or `Optional<BigDecimal>`), and all four `Issue7837` tests use directly-typed fields that never call it. Reverting it alone leaves the full core suite green while `@JsonFormat(shape = STRING) public Object examScore = new BigDecimal("0.47")` serializes to `{"examScore":string0}` — the exact bug of issue #7837, one type-hole over.

Witness:
```
revert this branch only (test kept) ->
  mvn -pl core test: exit 0, 0 failures
  Object-typed BigDecimal, @JsonFormat(shape=STRING) -> {"examScore":string0}
```

Add an `Object`-typed case to `Issue7837` asserting `{"examScore":"0.47"}`. To prove it gates this branch, revert the branch and confirm the test goes red — the witness above is exactly that revert.

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -1048,15 +1054,15 @@ static ObjectWriter getObjectWriter(
...
                 } else {
                     return new ObjectWriterImplBigDecimal(new DecimalFormat(format), null);
                 }
             }
 
             if (BigDecimal[].class == valueClass) {
-                if (format == null || format.isEmpty()) {
+                if (format == null || format.isEmpty() || "string".equals(format) || "millis".equals(format)) {
                     return new ObjectWriterArrayFinal(BigDecimal.class, null);
```
wenshao: **[Suggestion]** The `BigDecimal[]` branch ships with nothing gating it: reverting it alone leaves the entire 7992-test core suite green, and a natural `@JsonFormat(shape = STRING) BigDecimal[]` test would not pin it either, because the field-level `WriteNonStringValueAsString` feature (set at FieldWriter.java:103 for `"string"`) wins inside `JSONWriter.writeDecimal(value, features, format)` — so even the annotation-shaped test stays green with this branch gone.

Witness:
```
revert this branch only ->
  mvn -pl core test: exit 0, 0 failures
```

Pin it with the feature-less keyword instead: `@JSONField(format = "millis") public BigDecimal[] scores = {new BigDecimal("0.47")};` asserting `{"scores":[0.47]}` — `"millis"` is not mapped to the feature at FieldWriter.java:103, so it reaches this branch unmasked. Remove this branch's `"millis"` check and that test must go red with `millis0`-style items, the same dispatch as the measured scalar case `{"value":millis0}`.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
```diff
@@ -460,14 +460,23 @@ public void getBeanInfo(BeanInfo beanInfo, Class objectClass) {
      */
     public ObjectWriter getObjectWriter(Type objectType, String format, Locale locale) {
         if (objectType == Double.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplDouble.INSTANCE;
```
wenshao: **[Suggestion]** This branch (and its `Float`/`BigDecimal` siblings below) fixes a reachable path that no test exercises: the 3-arg `getObjectWriter` is called with the field format only from `FieldWriterList.getItemWriter` (FieldWriterList.java:104), i.e. collection fields — and `Issue7837` tests only directly-typed fields. Reverting the whole hunk leaves the full core suite green while a `List<Double>` field regresses: pre-fix behaviour measured as `{"values":[string0,string2]}` for `@JsonFormat(shape = STRING)` and `{"values":[millis0]}` for `format = "millis"` — the issue #7837 failure one level down.

Witness:
```
revert this hunk ->
  mvn -pl core test: exit 0, 0 failures
  List<Double> @JsonFormat(shape=STRING) -> {"values":[string0,string2]}
  List<Double> format="millis"            -> {"values":[millis0]}
```

Add to `Issue7837` a `List<Double>` (and `List<BigDecimal>`) field with `@JsonFormat(shape = STRING)` asserting `{"values":["0.47"]}`, plus a `@JSONField(format = "millis") List<Double>` asserting `{"values":[0.47]}` — a STRING-shape list test alone would not pin the `"string"` half (feature masking as above), which is why the millis variant matters here. Revert the hunk and both must go red; the witness above is that revert.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
```diff
@@ -460,14 +460,23 @@ public void getBeanInfo(BeanInfo beanInfo, Class objectClass) {
...
+            }
             return new ObjectWriterImplDouble(new DecimalFormat(format));
         }
 
         if (objectType == Float.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplFloat.INSTANCE;
```
wenshao: **[Suggestion]** Same gap as the `Double` branch above, for `Float`: this branch is reached only via `FieldWriterList.getItemWriter` (a `List<Float>` field with a format), no test covers it, and reverting it leaves the full core suite green while `List<Float>` items regress to `DecimalFormat("string")`-driven illegal JSON.

Witness:
```
revert the provider hunk (all three branches) ->
  mvn -pl core test: exit 0, 0 failures
```

Add a `List<Float>` case to `Issue7837` asserting quoted items (`["0.47"]`); revert this branch and it must go red.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
```diff
@@ -460,14 +460,23 @@ public void getBeanInfo(BeanInfo beanInfo, Class objectClass) {
...
+            }
             return new ObjectWriterImplFloat(new DecimalFormat(format));
         }
 
         if (objectType == BigDecimal.class) {
+            if ("string".equals(format) || "millis".equals(format)) {
+                return ObjectWriterImplBigDecimal.INSTANCE;
```
wenshao: **[Suggestion]** Same gap as the `Double` branch above, for `BigDecimal`: reached only via `FieldWriterList.getItemWriter` (a `List<BigDecimal>` field with a format), untested, and reverting it leaves the full core suite green while list items regress to `string0`/`millis0` illegal JSON.

Witness:
```
revert the provider hunk (all three branches) ->
  mvn -pl core test: exit 0, 0 failures
```

Add a `List<BigDecimal>` case to `Issue7837` asserting `{"values":["0.47"]}`; revert this branch and it must go red.

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
wenshao: **[Suggestion]** The keyword-exclusion predicate this PR adds is now hand-pasted at six sites in two files, in two already-divergent shapes, and the hand-listed vocabulary is incomplete at birth: the same `FieldWriter` constructor this guard protects also interprets `"trim"`/`"symbol"` as keywords (FieldWriter.java:143-144), but they are not excluded here — so `@JSONField(format = "trim")` on a `Double` field still builds `new DecimalFormat("trim")` and serializes as `{"v":trim0}`, the same illegal-JSON class issue #7837 closes. The shapes also disagree: the `FieldWriter` copies (1057/1065) accept null/empty format while these three do not, so `getObjectWriter(Double.class, "", locale)` builds `new DecimalFormat("")` and writes raw `.47` — and a sweep found 14 keyword-spelling sites in core where the 8 pre-existing ones all carry a null/empty check, leaving only the three new copies here diverging from the file's own convention.

Witness:
```
PR head, measured:
  @JSONField(format="trim") Double   -> {"v":trim0}
  @JSONField(format="symbol") Double -> {"v":symbol0}
  getObjectWriter(Double.class, "", null) -> new DecimalFormat("") -> .47
sweep: 14 keyword sites in core; 8 pre-existing, all null/empty-guarded
```

Extract one package-private predicate and call it from all six `DecimalFormat`-creation sites (this spans multiple locations, so no single-line suggestion):

```java
static boolean isFormatKeyword(String format) {
    return "string".equals(format) || "millis".equals(format);
}
```

The predicate must gate only `DecimalFormat` creation: at ObjectWriterCreator.java:1352-1364 a `long` field with format `"millis"` must still fall through to `new FieldWriterMillis` (epoch-millis semantics), so `"millis"` cannot be globally normalized to "no format". To prove the fix, add a test asserting `@JSONField(format = "trim")` on a `Double` yields `{"v":0.47}` — red today (`{"v":trim0}` measured), green once the predicate covers the vocabulary.

wenshao (commented):
Reviewed. Suggestions are inline. 1 Suggestion-level finding(s) could not be anchored to a changed line and were dropped; nothing further to act on here.

Not explored to full depth (tool budget reached): `"agent 6b"`: `mutation run of the  FieldWriter.getObjectWriter  BigDecimal/BigDecimal[] hunk (base  FieldWriter  shadow) — load-bearing status asserted from code reading plus…`; `"agent 6b"`: `bytecode-size comparison of the changed  FieldWriter.writeInt32 /constructor against the 35/325-byte inlining thresholds — skipped because the constructor is co…`; `"agent 6b"`: `full  mvn -pl core test  (only the new + adjacent format tests were run).`.

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -72,6 +72,13 @@ public abstract class FieldWriter<T>
             "initObjectWriter"
     );
 
+    static boolean isFormatKeyword(String format) {
+        return "string".equals(format)
+                || "millis".equals(format)
```
wenshao, line 77: **[Suggestion]** R2-1: The new guard is a denylist that leaves its class open: it names four keywords, and every other non-pattern `format` still reaches `DecimalFormat`, so the illegal-JSON class this PR closes stays reachable one keyword over.

Measured on this build, on the lines this PR guards:

- `@JSONField(format = "iso8601")` on a `Double` field → `{"v":iso8601}`
- `@JSONField(format = "unixtime")` on a `BigDecimal` → `{"v":unixtime0}`
- `@JSONField(format = "abc")` → `{"v":abc0}`; `format = "yyyyMMdd"` → `{"v":yyyyMMdd0}`
- the *legal* pattern `@JSONField(format = "0000")` → `{"v":0000}`

Each is an unquoted, non-JSON token emitted with no exception thrown, so the reader silently loses the value — the same failure issue #7837 was filed for. `"iso8601"` is not hypothetical user text: the library emits it itself from `@JSONField(serializeFeatures = UseISO8601DateFormat)` (`ObjectWriterBaseModule.java:730`).

Witness:
```
HEAD, one probe per case (the BASE arm yields the same tokens):
  IsoDouble        {"v":iso8601}
  UnixBigDecimal   {"v":unixtime0}
  D0000            {"v":0000}
  new DecimalFormat("iso8601").format(0.47) -> "iso8601"
```

**Suggested fix**

Decide applicability of a `format` to a numeric field fail-closed where it is owned, instead of listing keywords: accept the format only when the formatted probe is a valid JSON number, or validate the formatted text before `writeRaw`. If the list shape is kept, it must at least cover the vocabulary the library emits itself (`iso8601`, `unixtime`).

The acceptance criterion: a case asserting `{"value":0.47}` for `@JSONField(format = "iso8601")` on a `Double`, plus the `"0000"` case, should be red before the change — then remove the guard and confirm both red again.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -72,6 +72,13 @@ public abstract class FieldWriter<T>
...
+                || "trim".equals(format)
+                || "symbol".equals(format);
```
wenshao: **[Suggestion]** R2-2: The `"symbol"` arm of this new guard has no test gating it, while the other three keywords are each pinned by the new test class.

Renaming `"symbol".equals(format)` to `"symbolZ".equals(format)` leaves the 18 new tests and the whole core suite green, so a later edit that narrows or reorders this list silently restores a malformed payload for that keyword — `new DecimalFormat("symbol").format(0.47)` returns `symbol0`, written unquoted as `{"value":symbol0}`.

Witness:
```
Mutant "symbol" -> "symbolZ", scratch copy of this PR tree:
  -Dtest=Issue7837 : Tests run: 18, Failures: 0
  mvn -pl core test: Tests run: 8006, Failures: 0, Errors: 0   (survivor)
  new DecimalFormat("symbol").format(0.47) -> "symbol0"
```

**Suggested fix**

Add the missing case alongside `testTrimDouble`:

```suggestion
    public static class BeanSymbolDouble {
        @JSONField(format = "symbol")
        public Double value = 0.47;
    }

    @Test
    public void testSymbolDouble() {
        assertEquals("{\"value\":0.47}", JSON.toJSONString(new BeanSymbolDouble()));
    }
```

The acceptance criterion: that new case must go red when the `"symbol"` disjunct is removed or renamed — today nothing reds.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/main/java/com/alibaba/fastjson2/writer/ObjectWriterProvider.java
```diff
@@ -460,14 +460,23 @@ public void getBeanInfo(BeanInfo beanInfo, Class objectClass) {
      */
     public ObjectWriter getObjectWriter(Type objectType, String format, Locale locale) {
         if (objectType == Double.class) {
+            if (format == null || format.isEmpty() || FieldWriter.isFormatKeyword(format)) {
+                return ObjectWriterImplDouble.INSTANCE;
```
wenshao: **[Suggestion]** R2-3: The `format == null || format.isEmpty() ||` prefix this commit adds at `:463`, `:470` and `:477` carries real behaviour and nothing pins it.

Removing the prefix from all three sites leaves the 18 new tests and the whole core suite green. What it carries is not cosmetic: an empty format previously built `new ObjectWriterImplDouble(new DecimalFormat(""))`, and `new DecimalFormat("").format(0.47)` returns `.47`, emitted unquoted as `{"values":[.47]}` — invalid JSON. The `format == null` half additionally removes a latent NPE, since `new DecimalFormat(null)` throws on this JDK.

Witness:
```
Mutant removing the prefix at ObjectWriterProvider.java:463/470/477:
  -Dtest=Issue7837 : Tests run: 18, Failures: 0
  mvn -pl core test: Tests run: 8006, Failures: 0, Errors: 0   (survivor)
  new DecimalFormat("").format(0.47) -> ".47"
```

**Suggested fix**

Pin the empty-format behaviour directly so the disjunct cannot be dropped unnoticed:

```suggestion
    @Test
    public void testEmptyFormat() {
        ObjectWriterProvider provider = JSONFactory.getDefaultObjectWriterProvider();
        assertSame(ObjectWriterImplBigDecimal.INSTANCE, provider.getObjectWriter(BigDecimal.class, "", null));
        assertSame(ObjectWriterImplDouble.INSTANCE, provider.getObjectWriter(Double.class, "", null));
        assertSame(ObjectWriterImplFloat.INSTANCE, provider.getObjectWriter(Float.class, "", null));
    }
```

One premise to respect: `FieldWriterList.java:100` (`if (format != null) {`) is the only in-repo caller of this overload and it already guards null, so a testable trigger has to come through the empty string or the public API — not through a field path with a null format.

The acceptance criterion: the assertion above must go red when the prefix is removed at any of the three sites.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -1048,15 +1060,15 @@ static ObjectWriter getObjectWriter(
             }
 
             if (BigDecimal.class == valueClass) {
-                if (format == null || format.isEmpty()) {
+                if (format == null || format.isEmpty() || isFormatKeyword(format)) {
                     return ObjectWriterImplBigDecimal.INSTANCE;
```
wenshao: **[Suggestion]** R2-4: The predicate was extracted, but the guard around it is still hand-inlined at five sites — and the helper's name invites a consolidation that would change behaviour.

`format == null || format.isEmpty() || FieldWriter.isFormatKeyword(format)` is written out at `FieldWriter.java:132`, `:1063`, `:1071` and `ObjectWriterProvider.java:463`, `:470`, `:477`. Meanwhile five pre-existing sites ask the same-looking question with a different vocabulary — `if (format == null || format.isEmpty() || "string".equals(format))` at `ObjectWriterCreator.java:1072`, `:1353`, `:2066`, `ObjectWriterCreatorASM.java:4412` and `:1931`. They are not interchangeable: the creator sites route a `long` field to `FieldWriterMillis` when the predicate is false, so unifying the two vocabularies on this helper's name silently changes the encoding for `format = "millis"` on `long` fields.

Witness:
```
@JSONField(format = "millis") long v = 1700000000000L, both writers:
  LongMillis (FieldWriterMillis)  json={"v":1700000000000}  jsonb=a64a76ac6553f100a5
  LongPlain  (FieldWriterInt64)   json={"v":1700000000000}  jsonb=a64a76be0000018bcfe56800a5
  LongIso    (FieldWriterMillis)  json={"v":"2023-11-15T06:13:20+08:00"}
```

**Suggested fix**

Let the helper absorb the "no format" test so each new site is one call, and scope its name to what it actually means:

```suggestion
    static boolean isFormatKeyword(String format) {
        return format == null || format.isEmpty()
                || "string".equals(format)
                || "millis".equals(format)
                || "trim".equals(format)
                || "symbol".equals(format);
    }
```

Rename it (or add one line of javadoc) so it reads as "formats that must not be interpreted as a `DecimalFormat` pattern" — `"millis"` is a real format for `long`/date fields, and the four `ObjectWriterCreator` sites must keep their narrower test.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

wenshao (commented):

Not explored to full depth (tool budget reached): `"agent 1a"`: `whether a declared  double[] / Double[] / Float[]  field with  @JsonFormat(shape = STRING)  still emits unquoted numbers after the guard leaves  decimalFormat  …`; `"agent reverse-audit (round 2)"`: `the fastjson1-compatible route for the reported repro (R2-8's scope) — not executed.`; `"agent reverse-audit (round 2)"`: `JSONB output of the changed writeInt32 branch — not executed.`; `"agent reverse-audit (round 2)"`: `the revert probe ran Issue7837 (18 tests) only, not the full core suite.`; `"agent reverse-audit (round 2)"`: `the five pre-existing format sites (ObjectWriterCreator 1072/1353/2066, ObjectWriterCreatorASM 4412, DateUtils 146) — not read; R2-4's scope.`, and 14 more.



Deferred under the convergence posture (round 2, not a blocker) — recorded, not requested in this round:

- `core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java:655 — [probe] the added writeInt32 branch restates a decision that already lives in FieldWriterInt32, and the family now expresses the rule three divergent ways`

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -644,7 +652,11 @@ public void writeInt16(JSONWriter jsonWriter, short[] value) {
 
     public void writeInt32(JSONWriter jsonWriter, int value) {
         writeFieldName(jsonWriter);
-        jsonWriter.writeInt32(value);
+        if ((features & WriteNonStringValueAsString.mask) != 0) {
+            jsonWriter.writeString(Integer.toString(value));
```
wenshao, line 666: **[Suggestion]** R2-5: Use the typed `JSONWriter.writeString(int)` overload here instead of boxing the value into a `String` first.

Every write of an int field whose features carry this mask takes this branch, so a bean with `@JsonFormat(shape = STRING)` on an `Integer` field performs one avoidable young-gen `String` allocation plus a redundant string encode per value, on the path these writers are built to keep allocation-free. The same branch grows `FieldWriter.writeInt32` past the inlining target this project holds hot-path methods to (`AGENTS.md`: 35 bytes for C1).

Witness:
```
javap -c -p, project compile flags (last instruction offset + width):
  BASE writeInt32(JSONWriter,int)         10: return  -> 11 bytes
  HEAD                                    37: return  -> 38 bytes
  SWAP jsonWriter.writeString(value)      34: return  -> 35 bytes

Output identity, both arms (UTF8/UTF16 writeString(int) drops its own quotes then calls
writeInt32, which applies the reciprocal mask test, so the two polarities cancel):
  HEAD  A plain={"v":"47"}  B ctx-feature={"v":"47"}  C explicit-ctx={"v":"47"}  E jsonb=a64a764b3437a5
  SWAP  A plain={"v":"47"}  B ctx-feature={"v":"47"}  C explicit-ctx={"v":"47"}  E jsonb=a64a764b3437a5
```

**Suggested fix**

```suggestion
                jsonWriter.writeString(value);
```

`JSONWriter.writeString(int)` (`JSONWriter.java:2646`; implemented at `JSONWriterUTF8.java:401` and `JSONWriterUTF16.java:2902`) writes the quotes and digits directly, and `JSONWriterJSONB.writeString(int)` delegates to `writeString(Integer.toString(value))`, so JSONB bytes are unchanged.

One premise to respect: the quoting of `writeString(int)` and `writeInt32(int)` is complementary (`JSONWriterUTF8.java:401` computes `writeAsString = (context.features & mask) == 0`; `:1714` computes the inverse), so keep calling `writeString` and do not "simplify" this to `writeInt32(value)` — that would drop the quotes in exactly the case this branch exists for.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7837.java
```diff
@@ -0,0 +1,236 @@
+package com.alibaba.fastjson2.issues_7800;
+
+import com.alibaba.fastjson2.JSON;
+import com.alibaba.fastjson2.annotation.JSONField;
```
wenshao, line 4: **[Suggestion]** R2-6: This test class replays the issue's bean shape but not the entry point it was filed against: every case goes through `com.alibaba.fastjson2.JSON`, and none through the compat `com.alibaba.fastjson.JSON` the reporter used.

Issue #7837 names `com.alibaba:fastjson 2.0.64` and its repro imports `com.alibaba.fastjson.JSON`, whose write plumbing is a separate configuration (`fastjson1-compatible/src/main/java/com/alibaba/fastjson/JSON.java:86-95`, `:1634`). A later change there re-breaks the reported path — `{"examScore":string0}` followed by the reporter's `illegal input error` — while this class and the core suite stay green: a grep over `fastjson1-compatible/src/test` finds no `shape =` case at all (the only Jackson-annotation test there, `Issue632.java`, uses `pattern`).

The behaviour is fixed on that route today — that is exactly why the missing test is the finding:
```
BASE: compat json={"examScore":string0}
      compat parseObject -> com.alibaba.fastjson.JSONException: illegal input error, offset 14,
                            character s, line 1, column 14      (the issue text verbatim)
HEAD: compat json={"examScore":"0.47"}
      compat parseObject -> OK
```

**Suggested fix**

Add `fastjson1-compatible/src/test/java/com/alibaba/fastjson/issues_compatible/Issue7837.java` running the issue's repro verbatim through `com.alibaba.fastjson.JSON` — `toJSONString` asserting `{"examScore":"0.47"}`, and `parseObject` round-tripping to `0.47`. That module already hosts compat-artifact issue tests using Jackson annotations (`Issue632.java:5`).

One premise to respect: `fastjson1-compatible/pom.xml:111-115` — `com.alibaba.fastjson.JSON.<clinit>` needs `fastjson2-extension` on the classpath (without it you get `NoClassDefFoundError: com/alibaba/fastjson2/support/AwtRederModule`), which is why this test belongs in `fastjson1-compatible` rather than in core.

The acceptance criterion: that compat test must be red at the merge base (the output and parse failure quoted above) and green at head.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

core/src/main/java/com/alibaba/fastjson2/writer/FieldWriter.java
```diff
@@ -122,6 +129,7 @@ public abstract class FieldWriter<T>
 
         DecimalFormat decimalFormat = null;
         if (format != null
+                && !isFormatKeyword(format)
```
wenshao: **[Suggestion]** R2-7: With the pattern suppressed for this keyword, the quoting of a `BigDecimal` field rides on the `WriteNonStringValueAsString` feature alone — and the JSONB side drops that feature, so the two encodings of the same bean disagree on the value's *type*.

For `@JsonFormat(shape = JsonFormat.Shape.STRING) public BigDecimal examScore = new BigDecimal("0.47")`, `JSON.toJSONString` gives `{"examScore":"0.47"}` while `JSONB.toJSONString(JSONB.toBytes(bean))` gives `{"examScore":0.47}` — a number. The identical annotation on a `Double`/`Float` field quotes in both encodings, and `BigDecimal[]`/`List<BigDecimal>` inherit the asymmetry. A service writing JSON and a consumer reading the JSONB stream of the same bean therefore see different types, and anything validating "examScore is a string" passes on one path and fails on the other.

Witness:
```
HEAD  BB_JSON={"examScore":"0.47"}   BB_JSONB={"examScore":0.47}   BB_BYTES=a6...b9022fa5 (b9=BC_DECIMAL)
BASE  BB_JSON={"examScore":string0}  BB_JSONB={"examScore":0.47}   BB_BYTES=a6...b9022fa5  <- byte-identical to HEAD
HEAD  DD_JSON={"value":"0.47"}  DD_JSONB={"value":"0.47"}   FF_JSON={"value":"0.47"}  FF_JSONB={"value":"0.47"}
HEAD  BBLIST_JSON={"values":["0.47"]}  BBLIST_JSONB=[0.47]
HEAD+fix  BB_JSONB="0.47" (4d=BC_STR_ASCII_FIX_4); every feature-off row keeps its exact bytes
```
The JSONB bytes for this bean are identical between base and head, so the JSONB side is untouched and the text-side fix is what makes the pair disagree. Every assertion in this PR is text-path — `grep` of the new test class returns 18 × `JSON.toJSONString` and 0 × `JSONB`.

**Suggested fix**

Honour the feature on the JSONB side for decimals, mirroring the `Double` path — in `JSONWriterJSONB.writeDecimal(BigDecimal value, long features, DecimalFormat format)` (`core/src/main/java/com/alibaba/fastjson2/JSONWriterJSONB.java:1431`), before the compact-int path:

```suggestion
        if ((features & WriteNonStringValueAsString.mask) != 0) {
            writeString(value.toString());
            return;
        }
```

Keep the patch to the feature only: the same method also ignores `format`, but that is separate, pre-existing behaviour this PR does not change.

One premise to respect: `ObjectWriterImplDouble.writeJSONB` already honours this feature on the JSONB side (`ObjectWriterImplDouble.java:26`) while `JSONWriterJSONB.writeDecimal` ignores `features` for every `BigDecimal` write — so feature-off output must stay byte-identical (same `BC_DECIMAL` compact encoding), or every default-encoded `BigDecimal` stream changes shape. Measured with the patch applied: all six feature-off rows keep their exact bytes.

The acceptance criterion: a JSONB assertion added to `testBigDecimal` (`JSONB.toJSONString(JSONB.toBytes(new Bean()))` equals `{"examScore":"0.47"}`) is red today and green with the fix.

_— deepseek-v4.1-flash@e6bf8ffe via Qwen Code /review (v0.23.4)_

PR #7848, opened 2026-09-03
fix: StringToAny to Boolean rejects TRUE/True/T/Y spellings, for issue #7842

### LINKED ISSUE #7842: [BUG]TypeUtils.castToJavaBean对Boolean支持的问题
### 问题描述
com.alibaba.fastjson.util.TypeUtils#castToJavaBean(java.lang.Object, java.lang.Class<T>)方法在转化为Boolean类型时候与fastjson严重不一致，所有的String都会走到StringToAny方法，非"true"会全部返回false
而com.alibaba.fastjson2.function.impl.ToBoolean中实现了为String类型的多种值判断，其中的String类型则永远走不到

### 环境信息
*请填写以下信息：*

 - 版本信息fastjson 2.0.31
 

### 重现步骤
*如何操作可以重现该问题：*

1.使用代码
```
TypeUtils.castToJavaBean("TRUE", Boolean.class)
```

### 期待的正确结果
*对您期望发生的结果进行清晰简洁的描述。*
希望可以支持相关字符串，如TRUE, 字符串1，True等，或者能与fastjson1保持一致的处理：

WLLTB:
Fixes #7842

`TypeUtils.castToJavaBean("TRUE", Boolean.class)` returns `false`. All String-to-Boolean conversions go through `StringToAny`, whose boolean branch only accepted lowercase `"true"`, while `ToBoolean` (used for every other source type) accepts `TRUE`/`True`/`T`/`Y`/`YES` and friends — so the String source type could never reach the richer logic.

### Summary of your change

- `StringToAny`: boolean branch now accepts `"true"`/`"T"`/`"Y"` case-insensitively (matching `ToBoolean`'s single-letter set and fastjson1's `castToBoolean` spellings); unrecognized strings keep the existing lenient `FALSE` fallback instead of throwing, since this converter is shared by `JSONObject`/`JSONArray`/`JSONPath`/list conversion paths.
- Added regression test `core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java`.

Note: `"1"` is deliberately left as `false` — `JSONObjectTest.test_invoke` pins that behavior, and `ToBoolean` doesn't accept `"1"` either. Happy to flip it if maintainers prefer fastjson1's `castToBoolean` semantics there.

Verified: core 7992 passed (standard + `-Dfastjson2.creator=reflect`), fastjson1-compatible 1355 passed, root `validate` clean.

- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed. (No docs impact.)

### DIFF
--- a/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
@@ -812,10 +812,7 @@ public class JSONObject
                 return null;
             }
 
-            return "true".equalsIgnoreCase(str)
-                    || "1".equals(str)
-                    || "T".equalsIgnoreCase(str)
-                    || "Y".equalsIgnoreCase(str);
+            return TypeUtils.isTrueSpelling(str) || "1".equals(str);
         }
 
         throw new JSONException("Can not cast '" + value.getClass() + "' to boolean");
--- a/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONObject.java
@@ -812,7 +812,10 @@ public class JSONObject
                 return null;
             }
 
-            return "true".equalsIgnoreCase(str) || "1".equals(str);
+            return "true".equalsIgnoreCase(str)
+                    || "1".equals(str)
+                    || "T".equalsIgnoreCase(str)
+                    || "Y".equalsIgnoreCase(str);
         }
 
         throw new JSONException("Can not cast '" + value.getClass() + "' to boolean");
--- a/core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
+++ b/core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
@@ -64,7 +64,12 @@ public class StringToAny
         }
 
         if (targetClass == boolean.class || targetClass == Boolean.class) {
-            return "true".equals(str);
+            if ("true".equalsIgnoreCase(str)
+                    || "T".equalsIgnoreCase(str)
+                    || "Y".equalsIgnoreCase(str)) {
+                return Boolean.TRUE;
+            }
+            return Boolean.FALSE;
         }
 
         if (targetClass == BigDecimal.class) {
--- a/core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
+++ b/core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
@@ -5,6 +5,7 @@ import com.alibaba.fastjson2.JSONArray;
 import com.alibaba.fastjson2.JSONException;
 import com.alibaba.fastjson2.util.DateUtils;
 import com.alibaba.fastjson2.util.IOUtils;
+import com.alibaba.fastjson2.util.TypeUtils;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -64,9 +65,7 @@ public class StringToAny
         }
 
         if (targetClass == boolean.class || targetClass == Boolean.class) {
-            if ("true".equalsIgnoreCase(str)
-                    || "T".equalsIgnoreCase(str)
-                    || "Y".equalsIgnoreCase(str)) {
+            if (TypeUtils.isTrueSpelling(str)) {
                 return Boolean.TRUE;
             }
             return Boolean.FALSE;
--- a/core/src/main/java/com/alibaba/fastjson2/util/TypeUtils.java
+++ b/core/src/main/java/com/alibaba/fastjson2/util/TypeUtils.java
@@ -2517,6 +2517,12 @@ public class TypeUtils {
         throw new JSONException("can not cast to int");
     }
 
+    public static boolean isTrueSpelling(String str) {
+        return "true".equalsIgnoreCase(str)
+                || "T".equalsIgnoreCase(str)
+                || "Y".equalsIgnoreCase(str);
+    }
+
     public static boolean toBooleanValue(Object value) {
         if (value == null) {
             return false;
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
@@ -1,6 +1,8 @@
 package com.alibaba.fastjson2.issues_7800;
 
+import com.alibaba.fastjson2.JSON;
 import com.alibaba.fastjson2.JSONArray;
+import com.alibaba.fastjson2.JSONException;
 import com.alibaba.fastjson2.JSONObject;
 import com.alibaba.fastjson2.util.TypeUtils;
 import org.junit.jupiter.api.Tag;
@@ -9,6 +11,7 @@ import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @Tag("regression")
@@ -65,16 +68,46 @@ public class Issue7842 {
         jo.put("f", "1");
         assertTrue(jo.getBoolean("f"));
         assertFalse(jo.getObject("f", Boolean.class));
+        // ... and except mixed-case null spellings, also a deliberate per-path contract:
+        // getBoolean uses an equalsIgnoreCase null guard (null), while getObject routes
+        // through StringToAny's case-sensitive "null" guard and falls back to FALSE
+        for (String s : new String[]{"NULL", "Null"}) {
+            JSONObject joNull = new JSONObject();
+            joNull.put("f", s);
+            assertNull(joNull.getBoolean("f"), s);
+            assertFalse(joNull.getObject("f", Boolean.class), s);
+        }
     }
 
     @Test
     public void testJSONArrayAgreement() {
         for (String s : new String[]{
-                "true", "TRUE", "T", "t", "Y", "y",
-                "false", "FALSE", "0", "F", "f", "N", "n", "abc"}) {
+                "true", "TRUE", "True", "T", "t", "Y", "y",
+                "false", "FALSE", "False", "0", "F", "f", "N", "n",
+                "YES", "abc", "2", "", "null"}) {
             JSONArray ja = new JSONArray();
             ja.add(s);
             assertEquals(ja.getObject(0, Boolean.class), ja.getBoolean(0), s);
         }
+        // "1" keeps the same deliberate per-path contract as the JSONObject twin
+        JSONArray ja = new JSONArray();
+        ja.add("1");
+        assertTrue(ja.getBoolean(0));
+        assertFalse(ja.getObject(0, Boolean.class));
+    }
+
+    public static class FlagBean {
+        public Boolean f;
+    }
+
+    @Test
+    public void testBeanParseBoundary() {
+        // deliberate boundary contract: tree/cast paths stay lenient ...
+        JSONObject jo = JSON.parseObject("{\"f\":\"t\"}");
+        assertTrue(jo.getBoolean("f"));
+        assertTrue(jo.getObject("f", Boolean.class));
+        assertTrue(TypeUtils.cast("t", Boolean.class));
+        // ... while bean-binding stays strict and rejects "t" as malformed
+        assertThrows(JSONException.class, () -> JSON.parseObject("{\"f\":\"t\"}", FlagBean.class));
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
@@ -1,5 +1,7 @@
 package com.alibaba.fastjson2.issues_7800;
 
+import com.alibaba.fastjson2.JSONArray;
+import com.alibaba.fastjson2.JSONObject;
 import com.alibaba.fastjson2.util.TypeUtils;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -44,5 +46,35 @@ public class Issue7842 {
         assertEquals(Boolean.FALSE, TypeUtils.cast("FALSE", boolean.class));
         assertEquals(Boolean.FALSE, TypeUtils.cast("abc", boolean.class));
         assertEquals(Boolean.FALSE, TypeUtils.cast("", boolean.class));
+    }
+
+    @Test
+    public void testAccessorAgreement() {
+        // getBoolean and getObject(Boolean.class) must agree on every spelling ...
+        for (String s : new String[]{
+                "true", "TRUE", "True", "T", "t", "Y", "y",
+                "false", "FALSE", "False", "0", "F", "f", "N", "n",
+                "YES", "abc", "2", "", "null"}) {
+            JSONObject jo = new JSONObject();
+            jo.put("f", s);
+            assertEquals(jo.getObject("f", Boolean.class), jo.getBoolean("f"), s);
+        }
+        // ... except "1", which is a deliberate per-path contract:
+        // getBoolean("1") is true, cast("1") stays false (pinned by JSONObjectTest.test_invoke)
+        JSONObject jo = new JSONObject();
+        jo.put("f", "1");
+        assertTrue(jo.getBoolean("f"));
+        assertFalse(jo.getObject("f", Boolean.class));
+    }
+
+    @Test
+    public void testJSONArrayAgreement() {
+        for (String s : new String[]{
+                "true", "TRUE", "T", "t", "Y", "y",
+                "false", "FALSE", "0", "F", "f", "N", "n", "abc"}) {
+            JSONArray ja = new JSONArray();
+            ja.add(s);
+            assertEquals(ja.getObject(0, Boolean.class), ja.getBoolean(0), s);
+        }
     }
 }
--- a/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
+++ b/core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
@@ -1,0 +1,48 @@
+package com.alibaba.fastjson2.issues_7800;
+
+import com.alibaba.fastjson2.util.TypeUtils;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Tag("regression")
+public class Issue7842 {
+    @Test
+    public void testTrueSpellings() {
+        for (String s : new String[]{"true", "TRUE", "True", "T", "t", "Y", "y"}) {
+            assertTrue(TypeUtils.cast(s, Boolean.class), s);
+        }
+    }
+
+    @Test
+    public void testFalseSpellings() {
+        for (String s : new String[]{"false", "FALSE", "False", "0", "F", "f", "N", "n"}) {
+            assertFalse(TypeUtils.cast(s, Boolean.class), s);
+        }
+    }
+
+    @Test
+    public void testLenientFallback() {
+        assertFalse(TypeUtils.cast("abc", Boolean.class));
+        assertFalse(TypeUtils.cast("2", Boolean.class));
+        assertFalse(TypeUtils.cast("YES", Boolean.class));
+        // "1" stays false: pinned by JSONObjectTest.test_invoke, do not change it here
+        assertFalse(TypeUtils.cast("1", Boolean.class));
+        assertNull(TypeUtils.cast("", Boolean.class));
+        assertNull(TypeUtils.cast("null", Boolean.class));
+        assertNull(TypeUtils.cast(null, Boolean.class));
+    }
+
+    @Test
+    public void testPrimitive() {
+        assertEquals(Boolean.TRUE, TypeUtils.cast("TRUE", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("1", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("FALSE", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("abc", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("", boolean.class));
+    }
+}
--- a/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
+++ b/core/src/main/java/com/alibaba/fastjson2/JSONArray.java
@@ -656,7 +656,10 @@ public class JSONArray
                 return null;
             }
 
-            return "true".equalsIgnoreCase(str) || "1".equals(str);
+            return "true".equalsIgnoreCase(str)
+                    || "1".equals(str)
+                    || "T".equalsIgnoreCase(str)
+                    || "Y".equalsIgnoreCase(str);
         }
 
         throw new JSONException("Can not cast '" + value.getClass() + "' to boolean");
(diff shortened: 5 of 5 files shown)


core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
```diff
@@ -64,7 +64,12 @@ public Object apply(Object from) {
         }
 
         if (targetClass == boolean.class || targetClass == Boolean.class) {
-            return "true".equals(str);
+            if ("true".equalsIgnoreCase(str)
+                    || "T".equalsIgnoreCase(str)
+                    || "Y".equalsIgnoreCase(str)) {
```
wenshao: **[Suggestion]** R1-1: The new truthy set `{true, T, Y}` (case-insensitive) matches none of the library's other String→Boolean converters, so the same JSON value now yields different booleans — or an exception — depending on which fastjson2 API reads it. For `{"flag":"Y"}`, `getBoolean("flag")` returns `false` (`JSONObject.java:815` accepts only `"true"` any case or `"1"`) while `getObject("flag", Boolean.class)` returns `TRUE` (routes through this converter) — two accessors of the same `JSONObject` disagree, and before this change both returned `false`. For `"T"`, `TypeUtils.cast("T", Boolean.class)` now returns `TRUE`, but `JSON.parseObject("{\"flag\":\"T\"}", Bean.class)` throws `JSONException` because the reader's single-char boolean arm accepts only `1`/`Y`/`0`/`N`. fastjson1-compatible splits the same way: `castToBoolean("T")` stays `false` while `castToJavaBean("T", Boolean.class)` returns `true`.

Witness:
```
head: value=Y getBoolean=false getObject=true AGREE=false
      value=T getBoolean=false getObject=true AGREE=false
base: value=Y AGREE=true / value=T AGREE=true  (hunk reverted — the split is created by this change)
```

Aligning in one direction fixes the whole family: extend `JSONObject.getBoolean`/`JSONArray.getBoolean`'s String branch (and optionally `TypeUtils.toBoolean`) to accept `T`/`Y` case-insensitively, or extract one shared helper used by all sites; alternatively narrow this set to an existing convention. Any unification must keep `"1"` per-path — `JSONObjectTest.java:1036-1037` pins `cast("1")` → `false` through this same converter, while `getBoolean("1")` → `true` is that method's own contract. To pin the fix, add `assertEquals(jo.getObject("f", Boolean.class), jo.getBoolean("f"))` over the spelling set — red today, green once the accessors agree — and confirm it reds again with the alignment reverted.

wenshao: **[Suggestion]** R1-2: Issue #7842's expected result explicitly names `"1"` — "希望可以支持相关字符串，如TRUE, 字符串1，True等，或者能与fastjson1保持一致的处理" — and offers fastjson1 consistency as the alternative. fastjson 1.2.83's `castToBoolean("1")` returns `TRUE`, but this change keeps `cast("1")` → `FALSE`, and the choice is deferred ("Happy to flip it if maintainers prefer") with no ruling on record — so merging as-is would close #7842 with its named spelling unanswered. One nuance found while verifying: the rationale "ToBoolean doesn't accept `1`" doesn't fully hold as a String-path reference, because `ToBoolean`'s String arm is registered only for Number sources (`ObjectReaderBaseModule.java:73-81`) — String inputs can never reach it.

Witness:
```
fastjson-1.2.83 jar (live run): castToJavaBean("1", Boolean.class) = true
this PR (live run):             castToJavaBean("1", Boolean.class) = false
```

Get a maintainer ruling and record it either way: accept `"1"` → `TRUE` here, or state on the PR/issue that `"1"` → `false` is the deliberate resolution. `JSONObjectTest.java:1036-1037` pins `"1"` → `false` through this same converter, so a flip must move that pin in the same change. Whichever way the ruling goes, `Issue7842.testLenientFallback`'s `cast("1")` assertion is the pin that must move — run it after the change and confirm it fails when the ruling is reverted.

wenshao (commented):

Not explored to full depth (tool budget reached): `"agent 6a"`: `verify JSONReaderUTF16.readBoolValue0 rejects the string "T" the same way JSONReaderUTF8 does (cited in Finding 1; only the UTF8 twin was read)`.

core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
wenshao: **[Suggestion]** R1-1: The widened truthy set `{true, T, Y}` now answers true on the accessor and cast layers, but the untouched String→boolean surfaces still disagree: `readBoolValue` in JSONReaderUTF8/JSONReaderUTF16 throws on `"T"`/`"t"`/`"y"` (the quoted single-char branch accepts only `'0'`/`'N'`→false, `'1'`/`'Y'`→true, case-sensitive; multi-char only true/false — and UTF8's message prints the raw byte, e.g. `: 116`), `JSONReaderJSONB` has no `"T"`/`"t"`/`"y"`/`"True"`, `TypeUtils.toBoolean`/`toBooleanValue`, `internal/Cast.toBooleanValue` and the CSV `parseBoolean` return false, and the fastjson1-compatible facades (`com.alibaba.fastjson` `JSONObject`/`JSONArray` `getBoolean`, compat `castToBoolean`) return false. This is the still-standing remainder of round 1: the round-2 commit closed the getBoolean-vs-getObject facet (both accessors now accept T/Y), and the reader/JSONB, toBoolean/Cast/CSV and compat facets remain. For payload `{"flag":"t"}` an application that validates inbound documents by bean-binding rejects them as malformed while a tree-based consumer of the same document reads true.

Witness (probe on this build):
```
JSONObject.getBoolean(t) = true / getObject(t, Boolean) = true / cast(t, Boolean) = true
UTF16 bean-binding throw: can not convert to boolean : t
UTF8  bean-binding throw: can not convert to boolean : 116
TypeUtils.toBoolean(t) = false / toBooleanValue(t) = false / parseBoolean('y') = false
compat: JSONObject.getBoolean(t) = false, castToBoolean(t) = false
getBoolean("1") = true, cast("1", Boolean) = false   <- "1" contract intact
```

Suggested fix: either align the untouched surfaces — add `'T'`/`'t'`/`'y'` to the `readBoolValue` single-char branches (UTF8 and UTF16 in lockstep, keeping the allocation-free byte compare and `'0'`/`'N'`→false, `'1'`/`'Y'`→true) and decide the toBoolean/Cast/CSV/compat surfaces — or, if the per-surface split is deliberate, document it in the PR and pin it in `Issue7842` the way the `"1"` contract is pinned. The `"1"` per-path contract (getBoolean true, cast false, pinned by `JSONObjectTest.test_invoke`) must stay intact either way. Acceptance criterion: a boundary assertion in `Issue7842` pinning the chosen bean-parse contract for `{"f":"t"}` — it must go red if either side of the chosen contract is reverted; no assertion covers that boundary today.

_— glm-5.3-flash via Qwen Code /review (v0.23.2)_

core/src/main/java/com/alibaba/fastjson2/JSONObject.java
```diff
@@ -812,7 +812,10 @@ public Boolean getBoolean(String key) {
                 return null;
             }
 
-            return "true".equalsIgnoreCase(str) || "1".equals(str);
+            return "true".equalsIgnoreCase(str)
+                    || "1".equals(str)
+                    || "T".equalsIgnoreCase(str)
+                    || "Y".equalsIgnoreCase(str);
```
wenshao: **[Suggestion]** R2-1: The truthy-spelling test is now hand-rolled in three places — byte-identical in `JSONObject.getBoolean` and `JSONArray.getBoolean`, minus the `"1"` term in `StringToAny` — so the next spelling change needs a three-file lockstep edit. That tax was already paid inside this PR: `b3ffa7e25` changed only `StringToAny` and `bca5118f1` had to retro-align both accessors. A future "add `YES`" request (which `function/impl/ToBoolean` already accepts) applied to two of the three sites would silently diverge the accessor paths from the cast path.

Witness: the proposed helper measured at 33 bytes of bytecode at `--release 8` (`javap`, `ireturn` at offset 32) — under the 35-byte `MaxInlineSize` gate.

Suggested fix:
```java
// TypeUtils
public static boolean isTrueSpelling(String str) {
    return "true".equalsIgnoreCase(str)
            || "T".equalsIgnoreCase(str)
            || "Y".equalsIgnoreCase(str);
}
// accessors:  return TypeUtils.isTrueSpelling(str) || "1".equals(str);
// StringToAny: if (TypeUtils.isTrueSpelling(str)) { return Boolean.TRUE; } return Boolean.FALSE;
```

The extraction must keep the deliberate per-path `"1"` divergence — accessors keep `|| "1".equals(str)` while `cast("1")` stays false (pinned by `JSONObjectTest.test_invoke`) — and the helper must stay ≤35 bytes of bytecode per the AGENTS.md codeSize rule. Acceptance criterion: `Issue7842`'s `testTrueSpellings`/`testAccessorAgreement`/`testPrimitive` stay green after the extraction, and dropping a spelling from the helper turns them red.

_— glm-5.3-flash via Qwen Code /review (v0.23.2)_

core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
```diff
@@ -0,0 +1,80 @@
...
+        assertEquals(Boolean.FALSE, TypeUtils.cast("FALSE", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("abc", boolean.class));
+        assertEquals(Boolean.FALSE, TypeUtils.cast("", boolean.class));
+    }
+
+    @Test
+    public void testAccessorAgreement() {
+        // getBoolean and getObject(Boolean.class) must agree on every spelling ...
+        for (String s : new String[]{
+                "true", "TRUE", "True", "T", "t", "Y", "y",
```
wenshao, line 58: **[Suggestion]** R2-3: The agreement loop's comment says the accessors "must agree on every spelling", but the loop omits the mixed-case null spellings where they actually disagree: `getBoolean`'s `"null".equalsIgnoreCase` guard returns null for `"NULL"`/`"Null"`, while `getObject("f", Boolean.class)` routes through `StringToAny`'s case-sensitive `"null".equals` guard, falls to the widened boolean branch, and returns `Boolean.FALSE`. The loop covers only lowercase `"null"`, where both sides agree — so the test certifies an agreement that does not hold. The underlying guard divergence predates this PR; what is new is the test claiming "every spelling".

Witness (probe on this build):
```
NULL: getObject=false getBoolean=null agree=false
Null: getObject=false getBoolean=null agree=false
null: getObject=null  getBoolean=null agree=true    <- control
with StringToAny's guard flipped to equalsIgnoreCase: all spellings agree
```

Suggested fix: add `"NULL"`/`"Null"` to the loop and align `StringToAny`'s null guard to `equalsIgnoreCase` — only the `Boolean.class` path changes (FALSE → null; the primitive path is unaffected because `boolean.class` registers `defaultValue=false`) — or, if the null-spelling split is deliberate, exclude the case variants with a comment pinning the contract, the way the `"1"` block does. Acceptance criterion: the added `"NULL"`/`"Null"` arms go red if `StringToAny.java:30` reverts to `"null".equals(str)`.

_— glm-5.3-flash via Qwen Code /review (v0.23.2)_

core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
```diff
@@ -0,0 +1,80 @@
...
+        jo.put("f", "1");
+        assertTrue(jo.getBoolean("f"));
+        assertFalse(jo.getObject("f", Boolean.class));
+    }
+
+    @Test
+    public void testJSONArrayAgreement() {
+        for (String s : new String[]{
+                "true", "TRUE", "T", "t", "Y", "y",
+                "false", "FALSE", "0", "F", "f", "N", "n", "abc"}) {
```
wenshao: **[Suggestion]** R2-2: `testJSONArrayAgreement` silently drops arms its JSONObject twin covers — `"True"`, `"False"`, `"YES"`, `"2"`, `""`, `"null"` — and omits the explicit `"1"` divergence pin, so `JSONArray.getBoolean`'s empty/null→null arms and the deliberate per-path `"1"` contract are unpinned on the array side.

Witness (flip pair on this build): breaking `JSONArray.getBoolean`'s null guard to return `Boolean.FALSE` leaves this test class 6/6 green; the identical break in `JSONObject.getBoolean` turns `testAccessorAgreement` red (`expected: <null> but was: <false>`).

Suggested fix: reuse `testAccessorAgreement`'s 20-item list (minus `"1"`) in the array loop and add the two explicit `"1"` assertions for JSONArray, exactly as the JSONObject twin does:
```java
ja.add("1");
assertTrue(ja.getBoolean(0));
assertFalse(ja.getObject(0, Boolean.class));
```

Both accessors must keep returning null (not false) for `""`/`"null"` — `StringToAny(Boolean.class, null)`'s defaultValue early return plus the `isEmpty`/`"null"` guards.

_— glm-5.3-flash via Qwen Code /review (v0.23.2)_

wenshao (commented):

Not explored to full depth (tool budget reached): `"agent 6a"`: `none — all planned checks completed within budget (base-side bytecode compile skipped as unnecessary:  getBoolean  sits 160 bytes under the 325 cap and the pre-…`; `"agent 1d"`: `none — the checklist walk completed within budget (9 tool calls of ~34).`; `"agent 3a"`: `none — all planned checks completed (core + all-module sweeps, v1-compat check, dead-code traces, citation verification, head-side and suggested-fix-side measur…`; `"agent 6c"`: `none — all checks cut short by the ceiling; the two probes I skipped (a runtime  JSON.parseObject  reader test and a javap measurement of  StringToAny.apply ) w…`.

_— glm-5.3-flash via Qwen Code /review (v0.23.2)_

core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
```diff
@@ -64,7 +65,10 @@ public Object apply(Object from) {
         }
 
         if (targetClass == boolean.class || targetClass == Boolean.class) {
-            return "true".equals(str);
+            if (TypeUtils.isTrueSpelling(str)) {
+                return Boolean.TRUE;
+            }
+            return Boolean.FALSE;
```
wenshao, line 71: **[Critical]** R1-1: [certifies-falsely] [regression] Still standing from rounds 1–2, now probe-confirmed at Critical strength: the widened truthy set `{true, T, Y}` reaches the cast / `getBoolean` / `getObject` layers, but the untouched String→boolean surfaces still disagree — `TypeUtils.toBoolean`/`toBooleanValue` (still `Boolean.parseBoolean`), the streaming/bean reader `readBoolValue` (UTF8/UTF16/JSONB throw on `"T"`/`"t"`/`"y"`), `internal/Cast.toBooleanValue`, the CSV `parseBoolean`, and the fastjson1-compatible facades, where the widening now even contradicts the facade itself (`getObject` picks up the widened `StringToAny` through the shared provider while `getBoolean` keeps the old set). An application holding `"t"`/`"Y"` data (issue #7842's motivation) reads opposite booleans through different fastjson2 paths with no exception: before this diff every path returned `false` consistently, so the split is introduced by this change.

Witness (probe, input `"t"`, PR build vs merge base):
```
cast("t") = true        getBoolean("t") = true     getObject("t", Boolean) = true
JSONObject.to(FlagBean).f = true   (Map assembly -> StringToAny)
JSONArray.to(FlagBean).f  = false  (Collection assembly -> FieldReaderBool.accept -> toBoolean)
parseObject(str, FlagBean) = JSONException: can not convert to boolean (streaming)
TypeUtils.toBoolean("t") = false   toBooleanValue("t") = false
base (merge-base): all paths false — the split is new in this diff
```

Suggested fix: align the untouched String→boolean surfaces on `isTrueSpelling` — at minimum `toBoolean`/`toBooleanValue`'s String branches (keeping their empty/`"null"` guards) — or pin and document the per-surface split deliberately in `Issue7842` the way the `"1"` contract is pinned. Note the fix must not violate the `"1"` per-path contract: `getBoolean("1")` stays `true` (JSONObject.java:815) and `cast("1")` stays `false`, pinned by `Issue7842.testLenientFallback:39` and `JSONObjectTest.test_invoke:1037` (verified live: folding `"1"` into `isTrueSpelling` reddens `test_invoke`). Acceptance criterion: a boundary assertion in `Issue7842.testBeanParseBoundary` pinning the chosen contract for the assembly entries — it must go red if either side of the chosen contract is reverted; no assertion covers that boundary today.

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

wenshao, line 71: **[Suggestion]** R1-2: Still standing — issue #7842's expected result explicitly names `"字符串1"` ("希望可以支持相关字符串，如TRUE, 字符串1，True等，或者能与fastjson1保持一致的处理") and offers fastjson1 consistency as the alternative, yet this PR keeps `cast("1")` → `false` with the choice deferred and no maintainer ruling on record; merging as-is would close #7842 with its named spelling unanswered.

Witness:
```
fastjson-1.2.83 jar (live run): castToJavaBean("1", Boolean.class) = true
                                castToBoolean("1") = true
this PR (live run):             castToJavaBean("1", Boolean.class) = false
```

Suggested fix: get a maintainer ruling and record it either way — accept `"1"` → `TRUE` here, or state on the issue that `"1"` → `false` is the deliberate resolution. Note the fix must not silently break the existing pins: `JSONObjectTest.java:1037` (test_invoke) pins `"1"` → `false` through this same converter, so a flip must move that pin in the same change, and `Issue7842.testLenientFallback`'s `cast("1")` assertion is the pin that must move — run it after the change and confirm it fails when the ruling is reverted.

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    public void testBeanParseBoundary() {
+        // deliberate boundary contract: tree/cast paths stay lenient ...
+        JSONObject jo = JSON.parseObject("{\"f\":\"t\"}");
+        assertTrue(jo.getBoolean("f"));
+        assertTrue(jo.getObject("f", Boolean.class));
+        assertTrue(TypeUtils.cast("t", Boolean.class));
+        // ... while bean-binding stays strict and rejects "t" as malformed
+        assertThrows(JSONException.class, () -> JSON.parseObject("{\"f\":\"t\"}", FlagBean.class));
```
wenshao, line 111: **[Suggestion]** R3-1: The comment here — "while bean-binding stays strict and rejects `\"t\"` as malformed" — is true only of the streaming entry. The assembly entries bind silently and in opposite directions, and no assertion pins either one, so the boundary this comment documents is not the boundary the tests enforce.

Witness (probe, input `"t"`, PR build):
```
parseObject(str, FlagBean)  = THROW JSONException   (streaming — the only pinned entry)
parseObject(json).to(FlagBean).f      = true   (Map assembly -> StringToAny)
parseArray("[\"t\"]").to(FlagBean).f  = false  (Collection assembly -> FieldReaderBool.accept -> toBoolean)
```

Suggested fix: add one assertion per assembly entry pinning its actual ruling (or the aligned behavior once R1-1's fix lands), and narrow the comment to the entry it describes. Note the streaming `assertThrows` stays valid regardless — any new assertion must not contradict it. Acceptance criterion: the added `to(FlagBean.class)` assertions themselves — reverting the boundary they pin must redden them; none exists today.

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

core/src/main/java/com/alibaba/fastjson2/util/TypeUtils.java
```diff
@@ -2517,6 +2517,12 @@ public static int toIntValue(Object value) {
         throw new JSONException("can not cast to int");
     }
 
+    public static boolean isTrueSpelling(String str) {
+        return "true".equalsIgnoreCase(str)
+                || "T".equalsIgnoreCase(str)
+                || "Y".equalsIgnoreCase(str);
+    }
```
wenshao, line 2524: **[Suggestion]** R3-2: `isTrueSpelling` and `function/impl/ToBoolean` are now two public owners of the truthy-spelling policy, answering opposite directions with nothing marking the difference as deliberate: `cast("YES")` → `false` (pinned by `testLenientFallback`) while `ToBoolean.apply("yes")` → `true`; `cast("t")` → `true` while `ToBoolean.apply("t")` throws. `ToBoolean`'s String arm is registry-unreachable (it is registered for Number source types only), so nothing breaks today — but the next spelling-level request must land in two places, deepening the split this PR's own commit message ("share isTrueSpelling helper") works to reduce.

Witness:
```
"YES": cast=false ToBoolean=true      "yes": cast=false ToBoolean=true
"t":   cast=true  ToBoolean=THROW     "y":   cast=true  ToBoolean=THROW
"No":  cast=false ToBoolean=false
```

Suggested fix: a javadoc line on `isTrueSpelling` naming the deliberate set `{true, T, Y}` (case-insensitive) and cross-referencing `ToBoolean` (and the readers' `{1, Y}` set) as intentionally different — or unify `ToBoolean`'s String arm on the helper if maintainers prefer one policy. Note `Issue7842.testLenientFallback` pins `cast("YES")=false` and `testTrueSpellings` pins T/t/Y/y as `true` — any unification must keep both green or explicitly re-pin them.

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java
wenshao, line 71: **[Suggestion]** R3-3: This widened branch reaches more registry consumers than the tests cover: typed-Map tree binding, record (no-default-constructor) binding, and `JSONPath.set` on Boolean fields all route through `getTypeConvert(String→Boolean)` and flipped from `false` to `true` with this diff — none is pinned, so a future regression in any of these faces would land silently.

Witness (probe, PR vs merge base):
```
record-tree("t")        false -> true
jsonpath-set(bean,"T")  false -> true
typedMap-tree("T")      false -> true
```
(The streaming `TypeReference<Map<String,Boolean>>` parse throws on both sides — the reader's `{1,Y}` branch answers first; the flips are the tree-binding routes above.)

Suggested fix: pin at least one real consumer face in `Issue7842`, e.g. a typed-map tree binding or `JSONPath.set` assertion on `"T"`. Note the new assertions must agree with the file's existing pins: `"1"`→`false` on cast and `"t"` throwing on streaming bean parse. Acceptance criterion: the added assertion — reverting any of these routes to the old spelling set must redden it.

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

wenshao (changes requested):
Not explored to full depth (tool budget reached): `"agent 2"`: `无（未触及工具预算上限）。`; `"agent 6a"`: `未运行测试套件验证  testBeanParseBoundary  当前通过（共享 worktree 禁止构建）；Finding 1 的  to(FlagBean.class)  lenient 行为是静态调用链推断（JSONObject.java:1297 → ObjectReaderAdapter.java:669…`; `"agent 6a"`: `ObjectReaderNoneDefaultConstructor.createInstance(Map) （:497）的字段装配分支未读完，无默认构造器 bean 的  to()  路径行为基于  ObjectReaderAdapter  主干推断。`; `"agent 6a"`: `fastjson1 兼容层  com.alibaba.fastjson.util.TypeUtils.castToBoolean （:423）方法体未逐行读，仅确认其为独立实现、不引用本 diff 的转换链。`; `"agent 1b"`: `无 —— 所有检查在工具预算内完成（约 17/35 次调用），无因预算中止的检查项。`, and 2 more.

Convergence: round 3 posted 5 inline comment(s), 4 of them reported for the first time; the previous round posted 4 (3 new). Findings keep coming back to the same files: `core/src/main/java/com/alibaba/fastjson2/function/impl/StringToAny.java` (findings in round 1; 2 more now); `core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7842.java` (findings in round 2; 1 more now). The rate of new findings is not falling. A cluster that keeps producing siblings usually means the fixes are treating instances of a shared root cause — triaging that cause before the next round, or splitting an independent cluster into its own pull request, tends to end the loop faster than fixing them one at a time. Batching the remaining fixes and verifying them before the next push, or dropping this PR's reviews to `--severity-floor critical`, keeps the loop from re-deriving the same set. (Observation only — nothing was withheld from this review because of this observation.)

_— glm-5.3-flash via Qwen Code /review (v0.24.3)_

PR #7859, opened 2026-09-06
fix: resolve forward schema references in array items

### LINKED ISSUE #7855: [BUG] JSONSchema silently fails to resolve forward references in $defs (loses validation rules depending on key order)
### Description
`JSONSchema` silently fails to correctly resolve forward references inside the `$defs` (or `definitions`) block when parsing a schema using `JSONSchema.of()`. It does not throw any exceptions, but quietly loses validation information.

According to the JSON Schema specification, the order of keys in a JSON object is not regulated, and validators must support forward references. However, `fastjson2` processes `$defs` sequentially in a single pass. If a schema definition references another schema that appears later in the `$defs` object, the parser cannot find it in the internal catalog yet. Instead of throwing an error or resolving it lazily, it silently falls back to an empty schema (`{}`), causing incorrect validation behavior.

This issue seems directly related to an incomplete fix in **#1766** (where recursive references were partially introduced). While #1766 fixed certain cyclical cases, the strict key ordering in `$defs` still breaks the parser if the referenced schema hasn't been processed yet during the single-pass iteration.

### Steps to Reproduce
Given a JSON Schema where a definition (`attachments`) references a type (`attachment`) defined later in the `$defs` map:

```json
{
  "\$schema": "https://json-schema.org",
  "type": "object",
  "properties": {
    "data": { "\(ref": "#/\)defs/attachments" }
  },
  "\$defs": {
    "attachments": {
      "type": "array",
      "items": { "\(ref": "#/\)defs/attachment" }
    },
    "attachment": {
      "type": "object",
      "properties": {
        "id": { "type": "string" }
      },
      "required": ["id"]
    }
  }
}
```

### Actual Behavior
No exceptions are thrown, but the reference isn't resolved correctly. When `JSONSchema.of` compiles `attachments`, it looks for `#/$defs/attachment` inside `this.defs`. Since the iteration hasn't reached the `attachment` key yet, it is missing from the catalog, resulting in `items` being compiled as an empty schema `{}`.

As a result, the parsed schema for `attachments` incorrectly becomes:
```json
{"type":"array","items":{}}
```
This breaks any strict validation for array items.

### Expected Behavior
The parser should support forward references within `$defs` regardless of the key order. This can be achieved by using a two-pass compilation approach (registering all definitions first, then resolving internal `$ref` links) or utilizing proper lazy evaluation for references.

### Environment
* **Fastjson2 Version:** 2.0.65
* **JDK Version:** 17

hansu650:

Fixes #7855.

An array definition can refer to another definition that has not been compiled yet. Previously, `items` and `prefixItems` retained an unresolved schema (or `null` for `definitions`), so changing definition order could silently accept invalid array elements or produce a `NullPointerException`. Only object properties participated in the existing deferred-reference resolution, and array roots did not collect those tasks.

### Summary of your change

- Register deferred resolution for array `items` and individual `prefixItems` slots, including schemas whose root is an array.
- Keep the originating definition map with each unresolved reference so `$defs` and `definitions` resolve independently, even when they contain the same names.
- Pre-register the names in `definitions` so deferred references are created only for declared definitions, preserving the rejection of missing legacy property references.
- Resolve pending tasks after the root schema is constructed, then release the task list. Validation continues to call the resolved schemas directly; it does not perform a new lookup for each data item.
- Add 21 regression cases covering forward/backward order, object/array roots, tuple items, recursive children, required properties, type validation, and separate definition namespaces.

This is a focused fix for local definition references, not a new remote-reference loader or a change to the public API. No documentation changes are needed.

### Verification

- Before the fix, the new regression suite runs 21 cases: 11 failures, 3 errors, 7 passing controls.
- After the fix, all 21 cases pass, along with both existing `Issue1766` test classes (23 total).
- `mvnw.cmd -pl core test`: 8,009 tests, no failures or errors.
- `mvnw.cmd -pl core -Dfastjson2.creator=reflect test`: 8,009 tests reported, no failures or errors; 3 existing ASM-only tests are explicitly disabled in reflect mode (8,006 executed successfully).
- `mvnw.cmd validate`: all 17 active modules passed, with no Checkstyle violations.
- Tested on Windows 11, Temurin JDK 21.0.12.1, Maven 3.9.9. Other JDK/OS combinations are left to CI.

<details>
<summary>中文说明</summary>

修复 #7855：当数组定义引用了后面才出现的定义时，`items` / `prefixItems` 会保留未解析占位对象（旧式 `definitions` 则可能得到 `null`），造成非法数组元素被静默放行或空指针异常。原有延迟解析只处理对象属性，数组根节点也没有收集解析任务。

本次复用延迟解析机制，补齐数组元素和元组位置的引用；保留引用所属的定义表，避免 `$defs` 与 `definitions` 同名条目混淆；预登记旧式定义名称，只为已声明的条目创建延迟引用，避免把不存在的旧式属性引用变成默认放行；在根 Schema 构建完成后统一解析并释放任务列表。数据校验阶段仍直接调用目标 Schema，不增加逐元素查表开销。

新增 21 个回归用例，覆盖定义先后顺序、对象/数组根节点、`prefixItems`、递归子节点、必填字段、类型约束及两个定义命名空间。修复前为 11 个断言失败、3 个异常、7 个通过；修复后全部通过，现有两组 `Issue1766` 回归也通过。

验证环境为 Windows 11、Temurin JDK 21.0.12.1、Maven 3.9.9。核心模块标准模式 8,009 个用例通过；反射模式共报告 8,009 个用例，8,006 个通过、3 个按项目配置跳过的 ASM 专用用例，均无失败或错误。仓库全部 17 个启用模块的 `validate` 通过，无 Checkstyle 违规。其他 JDK/操作系统组合交由 CI 验证。本 PR 不引入远程引用加载或新的公开 API。

</details>


core/src/main/java/com/alibaba/fastjson2/schema/JSONSchema.java
```diff
@@ -452,7 +452,11 @@ public static JSONSchema of(JSONObject input, JSONSchema parent) {
                         if (ref.startsWith("#/definitions/")) {
                             final int PREFIX_LEN = 14; // "#/definitions/".length();
                             String refName = ref.substring(PREFIX_LEN);
-                            return definitions.get(refName);
+                            JSONSchema refSchema = definitions.get(refName);
+                            if (refSchema == null && definitions.containsKey(refName)) {
+                                refSchema = new UnresolvedReference(refName, definitions);
+                            }
```
wenshao, line 458: **[Critical]** R1-1: [certifies-falsely] [regression] Forward `#/definitions/` refs now materialize an `UnresolvedReference` in every schema slot, but only `properties`/`items`/`prefixItems` register a resolve task. In the unwired slots the placeholder survives forever and `UnresolvedReference.validateInternal` returns SUCCESS, so data the base rejected (or crashed on) is silently accepted. Probe-confirmed entrances, head vs merge base `ec14674bf` (definitions must list the referrer before the target): `additionalProperties` inside a definition entry — `isValid({"a":{"anything":1}})` is true on HEAD where base rejected via the `add additionalProperties` branch; `patternProperties` — HEAD true where base NPE'd; `allOf`/`anyOf`/`oneOf` entries — base NPE'd (`allOf` at construction via `AllOf.java:63` `getType()`; the others at validate), HEAD silently accepts, and `oneOf` corrupts both directions: the always-SUCCESS placeholder counts as a matching branch, so invalid input is accepted (count=1) and valid input is rejected when another branch also matches (count>=2).

Witness:
```
HEAD 1ce9463f vs BASE ec14674bf, same input both arms:
addProps-FWD  isValid=true  / isValid=false
patternProps  isValid=true  / THROWN NullPointerException
anyOf-fwd     validate=true / THROWN NullPointerException
allOf-fwd     construct=OK  / construct=THREW NullPointerException
oneOf-valid   validate=false (WRONG) / THREW
(items-FWD control: head=false / base=true — probe distinguishes the PR's own fix)
```

Suggested fix, ideally both layers: (1) wire the remaining slots the way `items`/`prefixItems` are wired — an `UnresolvedReference.ItemsResolveTask` (holds `JSONSchema[] items` + index) registered via `parent.addResolveTask(...)` in the `AllOf`/`AnyOf`/`OneOf` constructors when `of(item, parent)` returns an `UnresolvedReference`, plus equivalent tasks writing `additionalPropertySchema` and the matching `PatternProperty` schema; (2) defense in depth — make a placeholder that survives construction fail loudly instead of validating as SUCCESS.

Note: `AllOf.java:63` runs `type = itemSchema.getType()` immediately after `of(item, parent)`, so a fix that returns null for forward combinator refs re-introduces the construction-time NPE — the task-based fix never returns null and stays safe. Also, `UnresolvedReference.validateInternal` returning SUCCESS is the pre-existing behavior of the `$defs` branch for missing names, so making placeholders fail at validate time requires guarding `$defs` first (see the `$defs` suggestion in this review).

Please pin the fix with tests: one new `Issue7855` case per entrance (additionalProperties forward-ref asserting `isValid(...)` false, anyOf ditto, oneOf asserting invalid rejected AND valid accepted) — each goes red on current HEAD.

前向 `#/definitions/` 引用现在会在所有 schema 槽位生成 `UnresolvedReference`，但只有 `properties`/`items`/`prefixItems` 注册了解析任务。未接线的槽位会永远保留占位对象，`validateInternal` 返回 SUCCESS，导致基线版本会拒绝（或抛 NPE）的数据被静默放行。探针验证的入口（head 对比合并基线，definitions 中引用方需先于目标声明）：定义条目内的 `additionalProperties`（HEAD `isValid=true`，基线拒绝）、`patternProperties`（HEAD true，基线 NPE）、`allOf`/`anyOf`/`oneOf` 条目（基线 NPE，HEAD 静默接受；`oneOf` 双向出错：恒 SUCCESS 的占位对象被计为匹配分支，非法输入被接受、合法输入被拒绝）。建议两层修复：(1) 仿照 `items`/`prefixItems` 为其余槽位注册解析任务（组合器用持有 `JSONSchema[] items` + 下标的 `ItemsResolveTask`，经 `parent.addResolveTask(...)` 注册）；(2) 兜底：构造结束后仍存活的占位对象应响亮报错而非校验为 SUCCESS。注意 `AllOf.java:63` 在 `of(...)` 之后立即调用 `getType()`，返回 null 会重新引入构造期 NPE；另外 `validateInternal` 返回 SUCCESS 也是 `$defs` 分支对缺失名称的既有行为，改占位对象校验行为需先处理 `$defs` 分支。请为每个入口补充测试（每条在当前 HEAD 上都会变红）。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/UnresolvedReference.java
```diff
@@ -21,37 +24,59 @@ protected ValidateResult validateInternal(Object value) {
...
-        abstract void resolve(JSONSchema root);
+        final UnresolvedReference reference;
+
+        ResolveTask(UnresolvedReference reference) {
+            this.reference = reference;
+        }
+
+        JSONSchema getResolvedSchema() {
+            return reference.schemas.get(reference.refName);
+        }
```
wenshao, line 35: **[Critical]** R1-2: [certifies-falsely] [regression] Reference chains that never reach a concrete schema in one hop silently accept all input. `getResolvedSchema()` returns the raw map value with no chain-chasing and no concreteness check, so: a dangling alias (`a→missing`) resolves to null and the task no-ops, leaving the placeholder in place; a circular (`a→b→a`) or self alias keeps a placeholder; and a >=2-hop chain (`a→b→c`, `c` concrete) installs `U_b` — itself a placeholder — as the live property/item schema. In every shape `UnresolvedReference.validateInternal` returns SUCCESS forever, where the merge base threw NPE loudly. 1-hop aliases resolve correctly.

Witness:
```
HEAD 1ce9463f vs BASE ec14674bf (definitions namespace, properties slot):
dangling a→missing  isValid=true / THROWN NullPointerException
circular a→b→a     isValid=true / THROWN NullPointerException
2-hop a→b→c(string) isValid({"p":123})=true / THROWN NullPointerException
dangling $defs      isValid=true / isValid=true (pre-existing on base)
```
A candidate fix was applied in a scratch tree (chase alias chains bounded by map size, then throw `JSONException("schema error, unresolved reference : " + refName)` when the chase ends null or still unresolved): dangling/circular throw, the 2-hop chain resolves to the concrete tail and correctly rejects 123, and all 21 `Issue7855` tests stay green (recursive + forward variants keep resolving; `separateDefinitionNamespaces`' loud failure untouched).

Suggested fix: in `ResolveTask.resolve()` (both subclasses), chase `schemas.get(refName)` while the value is an `UnresolvedReference` (bounded by the map size as the cycle guard), then throw if the chase ends null or still unresolved. The drain runs at constructor end after both fill loops, so only permanently-unresolvable refs remain at that point.

Legitimate forward and recursive references stay unresolved mid-construction and must keep resolving — pinned by `Issue7855.recursiveDefinitions` (`attachment.children → attachments` resolves only at root-construction end), and the direct missing-name throw pinned by `separateDefinitionNamespaces` must remain.

Please pin with tests: `assertThrows(RuntimeException.class, ...)` for the dangling and circular variants plus a 2-hop chain asserting `isValid({"p":123})` false — all go red on current HEAD.

无法在一跳内到达具体 schema 的引用链会静默接受所有输入。`getResolvedSchema()` 直接返回 map 原值，既不追链也不检查具体性：悬空别名（`a→missing`）解析为 null，任务空操作、占位对象留存；循环（`a→b→a`）与自引用同理；两跳及以上的链（`a→b→c`，`c` 具体）会把 `U_b`（本身是占位对象）装成实际的属性/元素 schema。所有情形下 `validateInternal` 永远返回 SUCCESS，而合并基线会响亮地抛 NPE。建议修复：在 `ResolveTask.resolve()`（两个子类）中先按 map 大小为界追逐别名链，追逐结果仍为 null 或仍是占位对象则抛 `JSONException`。构造器末尾的 drain 在两个填充循环之后执行，此时只剩真正无法解析的引用。已用候选修复在临时目录验证：悬空/循环抛异常、两跳链正确解析并拒绝非法值，21 个 `Issue7855` 测试保持全绿。合法的前向与递归引用在构造中途保持未解析状态，必须继续可解析（`recursiveDefinitions` 已锚定）。请补充悬空/循环 `assertThrows` 与两跳链断言测试（当前 HEAD 上均变红）。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/ArraySchema.java
```diff
@@ -87,6 +92,11 @@ public ArraySchema(JSONObject input, JSONSchema root) {
...
             additionalItemsSupport = true;
             this.itemSchema = JSONSchema.of((JSONObject) items, root != null ? root : this);
+            if (itemSchema instanceof UnresolvedReference) {
+                JSONSchema resolveRoot = root == null ? this : root;
+                resolveRoot.addResolveTask(
+                        new UnresolvedReference.ArrayResolveTask(this, -1, (UnresolvedReference) itemSchema));
+            }
         }
 
         if (additionalItems instanceof JSONObject) {
```
wenshao, line 102: **[Suggestion]** R1-3: `additionalItems` is not covered by the new deferred resolution — no `instanceof UnresolvedReference` check and no `ArrayResolveTask` at this parse site, so a forward `$ref` there stays a permanent placeholder and items beyond `prefixItems` are accepted regardless of the referenced schema (e.g. `prefixItems:[{"type":"integer"}], additionalItems:{"$ref":"#/$defs/late"}}` with `late` a string definition accepts `[1,{"x":true},42]`; declaring `late` first rejects it — the exact order-dependence of issue #7855, one keyword over). Probe A/B shows this is pre-existing on the merge base for both namespaces (base built the UR for `$defs`, returned null for `definitions` — same accepted outcome), hence Suggestion rather than Critical; but this PR builds exactly the machinery that closes it.

Witness:
```
HEAD vs BASE, same input both arms:
addItems-FWD-$defs  isValid=true / isValid=true
addItems-FWD-defs   isValid=true / isValid=true
items-FWD-control   isValid=false / isValid=true (the PR's actual fix)
```

Suggested fix: mirror the items guard after `additionalItem = JSONSchema.of(...)` — register a task variant (or a `-2` sentinel) that assigns `arraySchema.additionalItem` in `resolve()`. `final JSONSchema additionalItem;` (ArraySchema.java:26) must become non-final exactly as `itemSchema` already is, and the sentinel must not collide with the `-1`/`>=0` discriminator in `ArrayResolveTask.resolve()`.

Please pin with a test in the `prefixItems` style: additionalItems as a forward `$ref`, asserting an extra item violating the referenced definition is rejected — it goes red on current HEAD.

`additionalItems` 未被新的延迟解析机制覆盖：该解析点没有 `instanceof UnresolvedReference` 检查、没有注册 `ArrayResolveTask`，前向 `$ref` 会永久保留占位对象，`prefixItems` 之后的元素不论引用的 schema 如何都被接受（与 issue #7855 完全同类的顺序依赖问题，仅隔一个关键字）。探针 A/B 显示这在合并基线上两个命名空间均已如此（基线对 `$defs` 生成占位对象、对 `definitions` 返回 null，结果同样是接受），故定为 Suggestion；但本 PR 恰好建立了能闭合它的机制。建议修复：仿照 items 的守卫，在 `additionalItem = JSONSchema.of(...)` 之后注册任务变体（或 `-2` 哨兵值）在 `resolve()` 中回写 `arraySchema.additionalItem`；`ArraySchema.java:26` 的 `final` 需去掉（与 `itemSchema` 一致），哨兵值不能与 `-1`/`>=0` 冲突。请补充对应测试（当前 HEAD 上变红）。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/JSONSchema.java
```diff
@@ -463,7 +467,7 @@ public static JSONSchema of(JSONObject input, JSONSchema parent) {
                             refName = URLDecoder.decode(refName);
                             JSONSchema refSchema = defs.get(refName);
                             if (refSchema == null) {
-                                refSchema = new UnresolvedReference(refName);
+                                refSchema = new UnresolvedReference(refName, defs);
```
wenshao, line 470: **[Suggestion]** R1-4: The `$defs` branch still creates an `UnresolvedReference` unconditionally for absent names, so a `$ref` to a name missing from `$defs` validates as SUCCESS forever — `{"$defs":{},"properties":{"p":{"$ref":"#/$defs/missing"}}}` silently accepts anything — while the `definitions` twin this PR guarded now fails loudly (pinned by your own `separateDefinitionNamespaces` test). Pre-existing on base (probe A/B identical on both arms), but this PR touched the line and pinned the asymmetric twin, which makes the asymmetry actionable now.

Witness:
```
HEAD vs BASE:
$defs-MISSING  isValid=true / isValid=true (byte-for-byte pre-existing)
defs-MISSING    THROWN NPE / THROWN NPE
```

Suggested fix: mirror the `definitions` treatment — pre-seed the `$defs` names with null in the `ObjectSchema`/`ArraySchema` constructors (next to the existing `definitions` pre-population), then guard this branch with `refSchema == null && defs.containsKey(refName)` so missing names return null like the `definitions` branch. A bare `containsKey` guard without the pre-seed would break your own `$defs,true` forward tests (a forward target is simply absent mid-fill). The R1-2 drain-time throw also closes this shape as a side effect.

Please pin with the `$defs` twin of `separateDefinitionNamespaces`' missing case: `assertThrows(RuntimeException.class, ...)` — currently no exception is thrown.

`$defs` 分支对缺失名称仍然无条件生成 `UnresolvedReference`，指向 `$defs` 中不存在名称的 `$ref` 会永远校验为 SUCCESS（`{"$ref":"#/$defs/missing"}` 静默放行一切），而本 PR 已加固的 `definitions` 孪生分支会响亮报错（你自己的 `separateDefinitionNamespaces` 测试已锚定）。基线上已如此（探针 A/B 两臂一致），但本 PR 改动了该行并锚定了不对称的孪生行为。建议修复：镜像 `definitions` 的处理——在两个构造器中预置 `$defs` 名称为 null，再用 `containsKey` 守卫该分支；仅加守卫而不预置会破坏你自己的 `$defs,true` 前向测试。R1-2 的 drain 期抛错方案也会顺带闭合此问题。请补充 `assertThrows` 测试（当前不抛异常）。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/ArraySchema.java
```diff
@@ -135,6 +150,21 @@ public ArraySchema(JSONObject input, JSONSchema root) {
         allOf = allOf(input, null);
         anyOf = anyOf(input, null);
         oneOf = oneOf(input, null);
+
+        if (resolveTasks != null) {
+            for (UnresolvedReference.ResolveTask resolveTask : resolveTasks) {
+                resolveTask.resolve();
+            }
+            resolveTasks = null;
+        }
```
wenshao, line 159: **[Suggestion]** R1-5: The resolve-task machinery — the `transient` field, the lazy-add `addResolveTask` override, and this drain loop — plus the `definitions` null pre-seed loop now exist as byte-identical copies in `ArraySchema` and `ObjectSchema`, while the base `JSONSchema.addResolveTask` hook is an empty no-op. The concrete cost: this very PR already had to relocate the drain once and add the `ArraySchema` copies — the duplication is the mechanism by which the historical object/array asymmetry arose. A future drain-semantics correction applied to `ObjectSchema` only would silently leave array-root schemas on the old behavior, reintroducing the asymmetry class this PR exists to eliminate.

Witness: not run — a structural duplication claim settled by direct code quotes (field at ArraySchema.java:37 == ObjectSchema.java:44; lazy-add and drain blocks behaviorally identical in both classes; base hook empty at JSONSchema.java:50); no runtime behavior distinguishes the copies.

Suggested fix: hoist the field, the lazy-add implementation, and a `protected final runResolveTasks()` drain into `JSONSchema` (replacing the no-op `addResolveTask`), call the drain at the end of both constructors, and extract a shared helper for the pre-seed loops. Keep the `transient` modifier. A short comment documenting the null-marker protocol (null = pending during construction; `containsKey` distinguishes a forward reference from a missing name; tasks must not drain before the fill loops complete) would also help — nothing in the code tells a future editor the loops and the drain are coupled.

`Issue7855`'s forward variants already pin both copies — if the hoist drops the drain call in either class, those tests go red.

解析任务机制（`transient` 字段、惰性 `addResolveTask` 重写、此 drain 循环）以及 `definitions` 置 null 预置循环如今在 `ArraySchema` 与 `ObjectSchema` 中逐字节重复，而基类 `JSONSchema.addResolveTask` 是空实现。具体代价：本 PR 已不得不迁移一次 drain 位置并补上 `ArraySchema` 的副本——这种重复正是历史上对象/数组不对称产生的机制。未来只改 `ObjectSchema` 一侧的 drain 语义修正会让数组根 schema 静默停留在旧行为。建议修复：将字段、惰性注册实现与 `protected final runResolveTasks()` drain 上提到 `JSONSchema`，两个构造器末尾调用，预置循环提取共享辅助方法；保留 `transient`；并加注释说明 null 标记协议（null = 构造期待解析；`containsKey` 区分前向引用与缺失名称；填充循环完成前不得 drain）。`Issue7855` 的前向用例已锚定两份副本。

</details>

core/src/test/java/com/alibaba/fastjson2/issues_7800/Issue7855.java
```diff
@@ -0,0 +1,123 @@
...
+        assertFalse(schema.isValid(JSONArray.of(JSONArray.of(JSONObject.of("id", 1)))));
+    }
+
+    @ParameterizedTest
+    @CsvSource({"$defs,true", "$defs,false", "definitions,true", "definitions,false"})
+    public void prefixItems(String keyword, boolean forward) {
+        JSONObject defs = definitions(keyword, forward);
+        JSONObject attachments = defs.getJSONObject("attachments");
+        attachments.remove("items");
+        attachments.put("prefixItems", JSONArray.of(ref(keyword, "attachment")));
```
wenshao, line 55: **[Suggestion]** R1-6: The only deferred `prefixItems` `$ref` in the whole corpus sits at index 0, where a wrong index equals the right one — no test would catch a regression that writes resolved prefix refs to the wrong slot (always-index-0, constant, off-by-one): slot 1 would stay unresolved (SUCCESS) and/or slot 0 would be overwritten, silently accepting `["s",{"id":"wrong"}]` and/or rejecting valid slot-0 strings. Corpus sweep: 0 deferred `#/$defs/`/`#/definitions/` refs at `prefixItems` index > 0 across `core/src/test` (the index>0 `$ref`s in `schema/draft2020-12/items.json` all resolve eagerly — targets precede referrers; `prefixItems.json` contains no `$ref` at all; `ref.json`'s index-1 ref is a `#/prefixItems/0` slot-ref through a different path and its test is commented out).

Witness: sweep over `core/src/test` — 51 `prefixItems` matches; the only deferred one is this single-element `JSONArray.of(ref(keyword, "attachment"))`.

Suggested fix: extend the def to two elements, `JSONArray.of(JSONObject.of("type","string"), ref(keyword, "attachment"))`, asserting `assertTrue(isValid([["s",{"id":"a"}]]))`, `assertFalse(isValid([["s",{"id":1}]]))` (slot 1 enforced), `assertFalse(isValid([[1,{"id":"a"}]]))` (slot 0 not clobbered) — garbled index propagation turns the `assertFalse`s red.

整个测试语料中唯一的延迟 `prefixItems` `$ref` 位于下标 0——错误下标恰好等于正确下标的位置。若回归把解析结果写到错误槽位（恒 0、常量、差一），不会有任何测试变红：槽位 1 保持未解析（SUCCESS）且/或槽位 0 被覆盖。语料扫描确认 `core/src/test` 中不存在下标大于 0 的延迟引用。建议：把定义扩展为两个元素并断言槽位 1 被约束、槽位 0 未被覆盖——下标传递出错时两条 `assertFalse` 变红。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/ArraySchema.java
```diff
@@ -135,6 +150,21 @@ public ArraySchema(JSONObject input, JSONSchema root) {
         allOf = allOf(input, null);
         anyOf = anyOf(input, null);
         oneOf = oneOf(input, null);
+
+        if (resolveTasks != null) {
+            for (UnresolvedReference.ResolveTask resolveTask : resolveTasks) {
+                resolveTask.resolve();
+            }
+            resolveTasks = null;
```
wenshao, line 158: **[Suggestion]** R1-7: The `resolveTasks = null;` clear is ungated — measured mutant probe: deleting this statement (alone, in an otherwise-PR tree) leaves all 21 `Issue7855` tests green, so nothing pins it. If it is deliberate hygiene (prevent re-resolution, keep the transient field null for serialization), a one-line comment saying so would prevent a future cleanup from removing it as dead code; otherwise drop it. Same for the twin clear in `ObjectSchema`.

Witness:
```
mutant probe (Agent 7, /tmp scratch trees):
delete clear in ArraySchema alone -> Issue7855 21/21 green
delete clear in ObjectSchema alone -> Issue7855 21/21 green
(behavioural hunks around it are gated: 14 of 21 cases go red on the merge base)
```

`resolveTasks = null;` 清空语句无测试锚定——实测删除该语句（单独删，其余保持 PR 状态）后 21 个 `Issue7855` 测试依然全绿。若属刻意清理（防重复解析、保持 transient 字段为 null），建议加一行注释说明，以免未来被当作死代码清除；否则可直接删除。`ObjectSchema` 中的孪生清空语句同理。

</details>

core/src/main/java/com/alibaba/fastjson2/schema/ObjectSchema.java
```diff
@@ -211,6 +208,13 @@ public ObjectSchema(JSONObject input, JSONSchema root) {
         allOf = allOf(input, null);
         anyOf = anyOf(input, null);
         oneOf = oneOf(input, null);
+
+        if (resolveTasks != null) {
+            for (UnresolvedReference.ResolveTask resolveTask : resolveTasks) {
+                resolveTask.resolve();
+            }
+            resolveTasks = null;
```
wenshao, line 216: **[Suggestion]** R1-7: The `resolveTasks = null;` clear is ungated — measured mutant probe: deleting this statement (alone, in an otherwise-PR tree) leaves all 21 `Issue7855` tests green, so nothing pins it. If it is deliberate hygiene, a one-line comment saying so would prevent a future cleanup from removing it as dead code; otherwise drop it. Same for the twin clear in `ArraySchema`.

Witness:
```
mutant probe (Agent 7, /tmp scratch trees):
delete clear in ObjectSchema alone -> Issue7855 21/21 green
delete clear in ArraySchema alone -> Issue7855 21/21 green
(behavioural hunks around it are gated: 14 of 21 cases go red on the merge base)
```

`resolveTasks = null;` 清空语句无测试锚定——实测删除该语句（单独删，其余保持 PR 状态）后 21 个 `Issue7855` 测试依然全绿。若属刻意清理，建议加一行注释说明；否则可直接删除。`ArraySchema` 中的孪生清空语句同理。

</details>

wenshao (changes requested):

